Issue #12, opened 2021-05-10
C -> python interface for regions


MatthiasKohl:
I'm not sure whether this is related to issue #11 but it would be good to provide an easier interface to handle existing C-level data in python, in particular for regions, but possibly other objects, too.

In particular, if we have a handle to a region in C, it's difficult to create a corresponding `legate.core.legion.Region` object:
1. I believe there is no way right now to get the fields of a field space (this is simply lacking from Legion's C interface AFAICT)
2. The `legate.core.legion.FieldSpace` class would have to support being created with a handle, like the `Region` and `IndexSpace` classes atm

Without this, I'm not sure if/how you can use the `__legate_data_interface__`, unless you create the underlying C objects on the python side.

manopapad:
I have added support for initializing a `FieldSpace` object from an existing handle, it should be ready to use once https://gitlab.com/StanfordLegion/legion/-/merge_requests/387 and https://github.com/nv-legate/legate.core/pull/13 are merged. Please try it out once possible and let me know if these changes are sufficient.

manopapad:
One comment that came up during the review of the above changes is that the existing FieldSpace class assumes it's the only code that will be allocating fields in the managed field space, so passing it an externally-created field space handle is potentially unsafe. We might want to create a different class for this use-case.

How exactly are you getting the field space handle, and what do you want to do with it?

Why do you need to list the fields in the field space?

MatthiasKohl:
Sorry for the late response, I had an issue with the mail associated with this account.

> How exactly are you getting the field space handle, and what do you want to do with it?

It is created by a separate module in C++ which the python code doesn't have control over. But the python code should take ownership of that field space.

> Why do you need to list the fields in the field space?

I don't need this at all actually, but ultimately, I want to be able to obtain a `Region` object in python (which should work with legate.numpy) from a `LogicalRegion` object that was created by another module in C++ and that the python code is taking ownership of.

So the only functionality I need in python is that the region object should be compatible with the current `Region` object and the resources should be released when deleting the object.

lightsighter:
Sorry for the slow response on this. One dimension we should at least consider here is whether this could be improved by using Cython which is something we've been considering to improve our interfaces for calling from Python to C. It might make transitioning back and forth between C/C++ and Python in both directions easier as well.

MatthiasKohl:
I think Cython would be useful here if it allows to directly call the Legion C++ API.
[...]

PR #13, opened 2021-05-10
Support initializing FieldSpace from existing handle

manopapad:
Requires the changes at https://gitlab.com/StanfordLegion/legion/-/merge_requests/387 to first be merged into legate_stable.

### DIFF
--- a/legate/core/legion.py
+++ b/legate/core/legion.py
@@ -1005,7 +1005,7 @@ class IndexPartition(object):
 
 
 class FieldSpace(object):
-    def __init__(self, context, runtime, owned=True):
+    def __init__(self, context, runtime, handle=None, owned=True):
         """
         A FieldSpace wraps a `legion_field_space_t` in the Legion C API.
         It is used to represent the columns in a LogicalRegion. Users can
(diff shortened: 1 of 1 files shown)


legate/core/legion.py
```diff
@@ -1017,14 +1017,42 @@ def __init__(self, context, runtime, owned=True):
             The Legion context from get_legion_context()
         runtime : legion_runtime_t
             Handle for the Legion runtime from get_legion_runtime()
+        handle : legion_logical_region_t
+            Created handle for a logical region from a Legion C API call
```
lightsighter, line 1021: Created handle for a field space...

legate/core/legion.py
```diff
@@ -1017,14 +1017,42 @@ def __init__(self, context, runtime, owned=True):
             The Legion context from get_legion_context()
         runtime : legion_runtime_t
             Handle for the Legion runtime from get_legion_runtime()
+        handle : legion_logical_region_t
```
lightsighter, line 1020: legion_field_space_t

lightsighter (changes requested):

legate/core/legion.py
manopapad: fixed

legate/core/legion.py
manopapad: fixed

legate/core/legion.py
```diff
@@ -1017,14 +1017,42 @@ def __init__(self, context, runtime, owned=True):
...
+                fids,
+                num_fields,
+            )
+            for fid in fids:
+                self.fields[fid] = legion.legion_field_id_get_size(
+                    runtime,
+                    context,
+                    handle,
+                    fid,
+                )
```
lightsighter, line 1051: I feel like this code is actually unnecessary and potentially unsafe. It's not clear that we can actually add/use these fields to our data structure since we're not sure who or what made the fields. For certain we can't delete these fields at any time since we didn't make them. I would be in favor of just ignoring any fields already in the field space. We could also provide a way for the client to pass in an optional list of fields that are already in in the field space that we are going to take ownership for.

manopapad: The user whose request motivated this change (see https://github.com/nv-legate/legate.core/issues/12) only needs to initialize a `FieldSpace` object from a `legion_field_space_t` so he can use that on a `Region` constructor, and won't be querying the fields on this `FieldSpace`, or creating new fields. Therefore it was sufficient to simply disallow such operations on `FieldSpace`s created from a handle.

@lightsighter Take a look if you can and let me know if the changes make sense to you.

PR #31, opened 2021-08-04
When generating CI images, name them after the branch

marcinz:
We want images for branches as we need to match the core image to the library version we are using in CI.


marcinz:
Everything appears to be working in a pull request. I will merge now to check if everything is working correctly in push events as well.

PR #63, opened 2021-10-04
Improve support for incremental builds

manopapad:
Now `install.py --noclean` will work even if all outstanding changes are in header files, and it will automatically know to recompile downstream libraries when relevant core header files have changed.


magnatelee:
This looks good to me.

PR #68, opened 2021-10-12
Support attaching onto existing Stores + related cleanups

manopapad:


legate/core/runtime.py
```diff
@@ -1012,24 +983,8 @@ def create_output_region(self, fspace, fields):
             fields=fields,
         )
 
-    def attach_external_allocation(self, context, alloc, shape, dtype, share):
-        if not isinstance(alloc, ExternalAllocation):
```
magnatelee, line 1016: Do we still have this check somewhere in the code?

magnatelee (approved):

PR #69, opened 2021-10-14
Rearrange user flags

magnatelee:
This PR rearranges user flags for the launcher so that they come last in the command. This allows them to override the values chosen by the launcher.


manopapad:
Note that the `--no-interpreter` option won't work anymore with this change (this case depended on the user to provide an executable), so we should remove it.

magnatelee:
@manopapad good catch. I'm gonna remove that flag as well.

PR #105, opened 2021-11-13
bug fix in install.py

sunwayihep:
In `find_active_python_version_and_path` function, the `p.replace(".a", ".dylib" if os_name == "Darwin" else ".so")` will lead to the bug if the directory path including the `.a` substring, e.g. `/home/software.ihep.ac.cn/python/libxxx.a` will be replaced with 
`/home/software.ihep.soc.cn/python/libxxx.so`, therefore a simple fix for this bug.

### DIFF
--- a/install.py
+++ b/install.py
@@ -104,7 +104,7 @@ def find_active_python_version_and_path():
     paths = [os.path.join(cv[p], cv["LDLIBRARY"]) for p in ("LIBDIR", "LIBPL")]
     # ensure that static libraries are replaced with the dynamic version
     paths = [
-        p.replace(".a", ".dylib" if os_name == "Darwin" else ".so")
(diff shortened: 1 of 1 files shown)


magnatelee:
@sunwayihep Thanks for this PR and sorry for the late response. The team is currently setting up contribution guidelines. Once we have them settled, we will merge this PR.

magnatelee:
@sunwayihep Sorry for the late response. We're ready to merge this PR, but before doing so, we'd like you to sign off your commit: https://github.com/nv-legate/legate.core/blob/branch-22.01/CONTRIBUTING.md#how-do-i-sign-the-dco. Note that you should amend the commit, instead of reverting and recommitting it, otherwise the change without a sign-off still appears in the history. Again, thank you for your contribution!

PR #113, opened 2021-12-06
Shiv1/op fusion3

shivsundram:
## **Task Fusion PR guide**
**This PR implements:**
1. Task Fusion in the Legate Core 
    i. Fusibility determination and creation of the fused ops, done in the python code (see `/core/runtime.py`)
    ii. Issuing of a Fused task's sub-operations, currently done in cunumeric c++ code, but this file/task will be moved to the core
    iii. An interface for accessing each registered Legate Tasks' function pointer, used for efficiently launching a fused task's sup-tasks
2. Basic constant optimization
   i. eliminate  issuing of expensive `convert` tasks for static/constant scalars (ie there's no need to defer execution of this type conversion, when the underlying Future's value is a constant embedded in the code)
3. Some changes to the benchmarks
    i. re-organization of the synchronizations in black_scholes. Instead checking the calls and puts!=NaN at the end of every benchmarking run, we check this only after all benchmarking runs have concluded.
    ii. A new 27-pt stencil benchmark. For demonstrating how fusion boosts perf even though it theoretically decreases task-level parallelism
4. TO_REMOVE: A C++ utility API for serializing a Legate Task, for issuing a legate task from another legate task - this will be removed from this PR, so feel free to ignore this code for now

**Required for merge (_ie TODO)_:**
-Move Fused_Task from cunumeric to the core
-Add OpenMP Support (ie create a `fused_op_template`file shared by CPU,OpenMP, and GPU codes, similar to `binary_op`)
-General cleanup (mostly  the `schedule` function in `core/runtime`)

## **Overview of Optimizations** 

**Task Fusion**
Given a window of buffered Legate tasks, fusion finds sub-windows of operations that can each be aggregated into a single launch/fused operation. In other words, given a 100% fusable window of N tasks, the runtime would normally issue O(N) meta/utility tasks (mapping calls, scheduling etc) for this window,  whereas the fused operation  would only require O(1) such meta-tasks, thus reducing overhead. There are currently four constraints dictating op fusability:

**Fusability Constraints**
**1 Fusable Operations:** Only a certain subset of Legate tasks are fusable (eg unary, binary, ternary ops)
**2 Producer-Consumer:** In the same fused op, one cannot write to one store representing an underlying buffer/regionfield, and then read from a different store representing the same underlying buffer/regionfield. This prevents the existence of invalid combinations of aliased partitions in a fused operation (eg the reads, adds, and mults in an in-place stencil are fusable, but not the write/update of the buffer at the end of the stencil iteration)
**3 Launch Shape equivalence:** All sub-operations in a fused task must have the same launch shape.
**4 Projection Functor equivalence:** If a store is used by mutliple sub-ops in a fused task, every subop's functor for indexing into that store must be equivalent. 

All constraints live in `core/runtime`. Each constraint's `apply` function is used to apply the constraint to a window of ops. This, by Dec 8, will be replaced with a new set of functions which apply constraints more efficiently (most of these are already in the PR and currently called apply2, but will just be called `apply` eventually)

_Currently, Constraint 4 is the most expensive to run, and it doesn't seem to ever trigger._  It is in the PR currently, but may warrant elimination if I can formally prove it is unnecessary. 

**Constant Optimization**
Currently implemented in cunumeric/array.py, when issuing binary_ops. Instead of creating a `convert` operation to convert a known scalar constant (eg from fp64 to tp32), we simply create the scalar in place and eliminate the need for said operation. Not only does this enable larger levels of fusion (as `convert` is not a fusable op), but the operation in itself provides large performance boosts. Note this only triggers when the constant is a `DeferredArray` (not Eager, as this  optimization actually slows down ops using EagerArrays )

## **Performance:**
Depending on input sizes and GPU counts (1-16 currently tested) and GPU type(Pascal, Volta currently tested), we see 
~1.6-2.x speedup for black scholes 
~1.6-2x speedup for 27pt stencil
~1.15-2x speedup for Logistic Regression
~1.15 speedup for Jacobi Iteration
No statistically significant speedup for conjugate gradient (between 1.0x-1.05x)

The maximum `window_size` in both the fused and unfused baseline (ie nv-legate reference) is set to 50 in these tests. Speedup is even higher when the baseline uses a window_size of 1 (current setting in reference), but the fused code still uses n=50 (current setting in PR)

## **Testing**
Currently running (have been testing as I go). Have not seen a single instance where fused and unfused generate different outputs. 

## **Known Issues:**
1. In 1 benchmark ( conjugate gradient, only noticed on Lassen, not Sapling), we see <0.5% performance degradation with fusion turned on. This is likely not statistically significant, but if concerning I can design an additional constraint to prevent fusion here. Very little gets fused in this example, so the overhead of making the fused op becomes more tangible (specifically the few fused ops are all O(n), and the unfusable ops are all O(n^2))
2. I have seen one instance where a fused task hangs, specifically  on a >=4 node count on Lassen for logistic regression (for smaller node counts it works fine). On my branch this happens with fusion turned off as well, and is likely due to one of the smaller changes I made outside the fusion code. I'm investigating. Worth noting that with the given problem size, the reference implementation actually generates "inf" floating point values.


install.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+#/home/shiv1/pypy/pypy3.8-v7.3.7-linux64/bin/python3.8
```
shivsundram, line 2: Todo(shiv): Remove 
(I tried running the legate core with pypy, this didn't end up working)

install.py
```diff
@@ -125,6 +126,7 @@ def git_clone(repo_dir, url, branch=None, tag=None, commit=None):
         verbose_check_call(
             ["git", "submodule", "update", "--init"], cwd=repo_dir
         )
+        git_reset(repo_dir, commit)
```
shivsundram, line 129: Todo: all changes to this file will be reverted to what's in nv-legate

legate/core/launcher.py
```diff
@@ -600,6 +605,12 @@ def add_output(self, store, proj, tag=0, flags=0):
             self._outputs, store, proj, Permission.WRITE, tag, flags
         )
 
+    # currently this is adding to outputs but we can have a seperate "temps" array in the core
+    def add_temp(self, store, proj, tag=0, flags=0):
+        self.add_store(
+            self._outputs, store, proj, Permission.WRITE, tag, flags
+        )
+
```
shivsundram, line 613: todo: remove
I was thinking of adding a notion of a 'temporary' that could be re-used. Didn't end up getting to this

legate/core/legion.py
```diff
@@ -5043,7 +5050,7 @@ def pack_dtype(self, dtype):
     def get_string(self):
         if self.string is None or self.arglen != len(self.args):
             fmtstr = "".join(self.fmt)
-            assert len(fmtstr) == len(self.args) + 1
+            #assert len(fmtstr) == len(self.args) + 1
```
shivsundram, line 5053: This assertion is being removed for performance reasons. The new method `pack_32bit_int_arr` above is a much faster way of packing a set of 32 bit ints (rather than appending them individually), but it breaks the invariant that `len(fmtstr) == len(self.args) + 1`

legate/core/operation.py
```diff
@@ -150,6 +161,7 @@ def execute(self):
 
     def get_tag(self, strategy, part):
         if strategy.is_key_part(part):
+            return 0
```
shivsundram, line 164: This one is interesting. When creating a fused operation, if one of the stores has  a partition with tag=0, but another reference to the the same store w/ the same partition has tag=1 in the same fused op, the runtime will  complain that these 2 partitions are aliased, even though theoretically nothing is wrong. I will likely change this to to ignore tags when coalescing within a fused o[

legate/core/operation.py
```diff
@@ -195,14 +209,29 @@ def add_dtype_arg(self, dtype):
...
+            launcher.add_fusion_metadata(self._is_fused, self._fusion_metadata)
+        if  self._is_fused: #fused ops re-use encapsulated unfused partitions
+            input_parts = self._unfused_input_parts
+            output_parts = self._unfused_output_parts
+            reduction_parts = self._unfused_reduction_parts
+        else:
+            input_parts = self._input_parts
+            output_parts = self._output_parts
+            reduction_parts = self._reduction_parts
+
```
shivsundram, line 229: There's two way of generating the set of partitions for a fused op:
-1 is to create the fused op, then repartition all the stores of the fused op (even though they were already partitioned in there unfused form)
-The other way is to build a `superstrategy` of all the individual 'unfused' partitions
The second is much cheaper (partitioning_stores twice has significant/tangible overhead) and is what is currently used in the code

legate/core/solver.py
```diff
@@ -290,15 +296,18 @@ def cost(unknown):
             if isinstance(prev_part, NoPartition):
                 partition = prev_part
             else:
-                partition = store.compute_key_partition(restrictions)
+                if store._key_partition is not None:
+                    partition=store._key_partition
+                else:
+                    partition = store.compute_key_partition(restrictions)
```
shivsundram, line 302: remove me (I guess)

legate/core/store.py
```diff
@@ -533,8 +534,28 @@ def storage(self):
...
+            if True:
+                import pdb; pdb.set_trace()
+                start = self
+                while start._storage is None and start._parent:
+                    start=start._parent
+                if start._storage:
+                    self._storage = start._storage
+                else:
+                    self._runtime._launch_outstanding()
+            """
```
shivsundram, line 558: todo: remove me

legate/core/store.py
```diff
@@ -908,3 +930,55 @@ def find_or_create_partition(self, functor):
         part = converted.construct(self.storage.region, complete=complete)
         self._partitions[functor] = (part, proj)
         return part, proj
+
+
```
shivsundram, line 934: This class contains all the metadata required for creating the fused op, and extracting the subops from the fused op within the Legate FusedTask. 
All input/output/reduction stores are organized into an array, and prefix sums/offsets are generated, which can be used to index into a particular sub-ops subset of stores

src/core/data/store.h
```diff
@@ -329,8 +340,29 @@ class Store {
   bool readable_{false};
   bool writable_{false};
   bool reducible_{false};
+
+ friend class MakeshiftSerializer;
 };
 
+//containts prefix sums for a sub-op
+//to index into its own data
```
shivsundram, line 348: any vector with the `starts` suffix refers to a prefix sum array. 
ie for sub-task "i" of the fused task, its input stores are located in: `context.inputs[i]:context.inputs[i+1]`

src/core/runtime/runtime.cc
```diff
@@ -32,6 +32,13 @@ Logger log_legate("legate");
 
 // This is the unique string name for our library which can be used
 // from both C++ and Python to generate IDs
+
+using LegateVariantImpl = void (*)(TaskContext&);
+/*static */ std::vector<std::pair<int64_t, LegateVariantImpl> > Core::opIDs = *(new std::vector<std::pair<int64_t, LegateVariantImpl> >());
+/*static */ std::vector<std::pair<int64_t, LegateVariantImpl> > Core::gpuOpIDs = *(new std::vector<std::pair<int64_t, LegateVariantImpl> >());
+/*static */ std::unordered_map<long, LegateVariantImpl> Core::cpuDescriptors = *(new std::unordered_map<long, LegateVariantImpl>());
+/*static */ std::unordered_map<long, LegateVariantImpl> Core::gpuDescriptors = *(new std::unordered_map<long, LegateVariantImpl>());
+
```
shivsundram, line 41: these arrays store function pointers to the registered Legate tasks. 
An openMP analogue will be added before merge

src/core/task/task.cc
```diff
@@ -57,6 +58,18 @@ void LegateTaskRegistrar::record_variant(TaskID tid,
...
+    auto newID = context.get_task_id(taskIdx.first);
+    Core::cpuDescriptors.insert(std::pair<int64_t, LegateVariantImpl>((int64_t) newID, taskIdx.second));
+  }
+
+  for (auto& taskIdx : Core::gpuOpIDs){
+    auto newID = context.get_task_id(taskIdx.first);
+    Core::gpuDescriptors.insert(std::pair<int64_t, LegateVariantImpl>((int64_t) newID, taskIdx.second));
+  }
+
+
```
shivsundram, line 71: register each legate tasks function pointer, so they can be accessed from a fused task

src/core/utilities/deserializer.cc
```diff
@@ -136,6 +189,62 @@ void MapperDeserializer::_unpack(Store& value)
...
+    {
+        metadata.futureStarts[i] = unpack<int32_t>();
+    }   
+    for (int i=0; i<nOps; i++)
+    {
+        metadata.opIDs[i] = unpack<int32_t>();
+    }   
+}
+
+
```
shivsundram, line 246: Todo: put this in a common method shared with the the other deserializer

src/core/utilities/makeshift_serializer.cc
```diff
@@ -0,0 +1,158 @@
+#include "core/utilities/makeshift_serializer.h"
```
shivsundram, line 1: this whole file will be torched

src/core/utilities/makeshift_serializer.h
```diff
@@ -0,0 +1,168 @@
+
```
shivsundram, line 1: again, this whole file will be torched, as it's no longer needed/used

legate/core/runtime.py
```diff
@@ -41,8 +45,23 @@
...
+    return
+if debugPrint:
+    drint = print
+else:
+    drint = zprint
+
+if futureBugPrint:
+    frint = print
+else:
+    frint = zprint
```
shivsundram, line 64: I used this for debugging/tracing. Will remove

legate/core/runtime.py
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
...
+            self.partitioners.append( partitioner )
+            strategy = partitioner.partition_stores()
+            for output, part, in zip(op._outputs, op._output_parts):
+                partition = strategy.get_partition(part)
+                output.set_key_partition(partition)
+                key_part = partition
+                for input in op._inputs:
+                    if input.shape==output.shape:
+                        input.set_key_partition(key_part)
+            #if len(op.inputs)>1:
```
shivsundram, line 776: This part is important. It sets key partitions so the partition of a store in fused op can propagate  to other references of the same store.
There may be a better way to do this than 
```               for input in op._inputs:
                    if input.shape==output.shape:
                        input.set_key_partition(key_part                        ```

File: legate/core/runtime.py [hunk 18]
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
...
+        if not streamApply:
+            results = [constraint.apply(self.contexts, self.runtime, self.ops, self.partitioners, self.strategies) for constraint in self.constraints]
+            all_fusable = [result[0] for result in results]
+            interval_sets = [result[1] for result in results]
+        else:
+            alpha = self.constraints[0].apply(self.contexts, self.runtime, self.ops, self.partitioners, self.strategies)
+            beta = self.constraints[1].apply2(self.contexts, self.runtime, self.ops, alpha[1],self.partitioners, self.strategies)
+            beta = self.constraints[2].apply2(self.contexts, self.runtime, self.ops, beta[1], self.partitioners, self.strategies)
+        #print("beta", beta)
+        """
```
shivsundram, line 797: yes this is messy, will clean up in the next day or so. The imporant thing is that the `apply2` class of methods are much more efficient at generating the windows of fusable ops than the `apply` methods

legate/core/runtime.py
```diff
@@ -753,10 +1146,14 @@ def __init__(self, core_library):
...
-            ty.uint32,
-        )
+        self._window_size=1
+        self._fusion_threshold =2
+        self._opLens = []
+        self._fusedOpLens = []
+        self._clearing_pipe = False
+        #self._window_size = self._core_context.get_tunable(
+        #    legion.LEGATE_CORE_TUNABLE_WINDOW_SIZE,
+        #    ty.uint32,)
```
shivsundram, line 1156: todo: put back in

legate/core/runtime.py
```diff
@@ -850,7 +1247,16 @@ def destroy(self):
...
+        #print("opLens")
+        #print(self._opLens)
+        #print("fusedOpLens")
+        #print(self._fusedOpLens)
+        #import pickle
+        #with open('fusedOpLens.pkl', 'wb') as f:
+        #    pickle.dump(self._fusedOpLens, f)
+        #with open('flushOpLens.pkl', 'wb') as f:
+        #    pickle.dump(self._fusedOpLens, f)
+        #pr.print_stats(sort='time')
```
shivsundram, line 1259: todo: remove me

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
         else:
             return op.launch(self.legion_runtime, self.legion_context)
 
-    def _schedule(self, ops):
+
```
shivsundram, line 1289: The prefix sums (for doing bookkeeping of the fused ops) are generated here. 
Given op 'i' in a fusable window, its particular inputs will be located between offsets
inputs_starts[i] and input_starts[i+1] within the inputs list

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+                      future_starts, op_ids)
+        fusion_metadata = FusionMetadata(*meta_arrs)
+        meta_maps=None
+        return meta_maps, fusion_metadata
+   
+
+    def build_fused_op(self,ops):
+
+        for i in range(len(ops)):
+            self.propogateFuture(ops[i])
```
shivsundram, line 1342: todo: remove me. this was a hack that is no longer needed

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+            for j in range(start,end):
+                super_strat = {**(super_strat.copy()), **partitions[j]._strategy}
+                super_fspace = {**(super_fspace.copy()), **partitions[j]._fspaces}
+                super_keystore = super_keystore.union(partitions[j]._key_parts)
+            super_strats.append(super_strat)
+            super_fspaces.append(super_fspace)
+            super_keystores.append(super_keystore)
+            super_strategies.append(Strategy(partitions[start]._launch_shape, super_strat, super_fspace, super_keystore))
+        #drint("lens", len(super_strats), len(super_fspaces), len(super_strategies), len(super_keystore))
+
```
shivsundram, line 1376: creates the "super-partition" by aggregating every unfused operations partitions/stratgies

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+        delta=stoptime-starttime
+        total = delta.total_seconds() * 1000.0
+        print("buildtime", total)
+        """
+   
+        #pr.enable()
+        redoPar=False
+        #starttime = datetime.datetime.now()
+        for i,fused_task in enumerate(new_op_list):
+            if redoPar:
```
shivsundram, line 1444: we currently have the option of redo-ing the partition for the fused op. But I will remove this

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+
+    def _launch_outstanding(self, force_eval=True):        
+        if len(self._outstanding_ops):
+            #self._opLens.append(len(self._outstanding_ops))
+            ops = self._outstanding_ops
+            self._outstanding_ops = []
+            self._schedule(ops, force_eval)
+
+
+    def propogateFuture(self,op):
```
shivsundram, line 1468: todo: remove

legate/core/runtime.py
```diff
@@ -898,11 +1570,12 @@ def _schedule(self, ops):
     def flush_scheduling_window(self):
         if len(self._outstanding_ops) == 0:
             return
+        #self._opLens.append(len(self._outstanding_ops))
         ops = self._outstanding_ops
         self._outstanding_ops = []
         self._schedule(ops)
 
-    def submit(self, op):
+    def submitNew(self, op):
```
shivsundram, line 1578: this was wonchan's new submit function.my scheduling will be 'moved' here once it's cleaned up

legate/core/store.py
```diff
@@ -489,7 +489,8 @@ def shape(self):
             # If someone wants to access the shape of an unbound
             # store before it is set, that means the producer task is
             # sitting in the queue, so we should flush the queue.
-            self._runtime.flush_scheduling_window()
+            self._runtime._launch_outstanding(False)
+            #self._runtime.flush_scheduling_window()
```
shivsundram, line 493: todo use flush_scheduling window instead

legate/core/store.py
```diff
@@ -533,8 +534,28 @@ def storage(self):
         # If someone is trying to retreive the storage of a store,
         # we need to execute outstanding operations so that we know
         # it has been initialized correctly.
-        self._runtime.flush_scheduling_window()
+        self._runtime._launch_outstanding(False)
```
shivsundram, line 537: todo: use flush_scheduling window instead

legate/core/store.py
```diff
@@ -544,6 +565,7 @@ def storage(self):
             #       if necessary
             if self._parent is None:
                 if self._kind is Future:
+                    print("supressing in store.py")
```
shivsundram, line 568: todo: remove

legate/core/operation.py
```diff
@@ -30,6 +34,9 @@ def __init__(self, context, mapper_id=0, op_id=0):
         self._inputs = []
         self._outputs = []
         self._reductions = []
+        self._is_fused = False
+        self._temps = []
+
```
shivsundram, line 39: todo: remove _temps

legate/core/operation.py
```diff
@@ -116,6 +123,10 @@ def add_output(self, store, partition=None):
         self._outputs.append(store)
         self._output_parts.append(partition)
 
+    def add_temp(self, store):
+        self._check_store(store)
+        self._temps.append(store) #this may not be necessary
+
```
shivsundram, line 129: todo: remove

legate/core/operation.py
```diff
@@ -195,14 +209,29 @@ def add_dtype_arg(self, dtype):
...
+        if self._is_fused:
+            launcher.add_fusion_metadata(self._is_fused, self._fusion_metadata)
+        if  self._is_fused: #fused ops re-use encapsulated unfused partitions
+            input_parts = self._unfused_input_parts
+            output_parts = self._unfused_output_parts
+            reduction_parts = self._unfused_reduction_parts
+        else:
+            input_parts = self._input_parts
+            output_parts = self._output_parts
+            reduction_parts = self._reduction_parts
```
shivsundram, line 228: Currently, a hash method is used to map a store (or rather it's constraints) to it's partition. all hasehes partition constraints (for inputs, outputs, reductions) are ingest their task's absolute id, which is problematic for fusion as the fused task will have a different absolute id than the unfused task. 
To take care of  this we set `input_parts` to `self._unfused_input_parts`, so we can hash into the original unfused partitions. Otherwise, the method for hashing into a store's partition will complain.

legate/core/runtime.py
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
         self._index_partitions[key] = index_partition
 
 
+class FusionChecker(object):
+    def __init__(self, ops, contexts, runtime):
```
shivsundram, line 733: This class registers a set of fusion constraints. It also currently calls the partitioner, but I will changes this so the partitioner is called outside this class (and the strategies are just passed in as args)

legate/core/runtime.py
shivsundram: `apply2` takes a set of fusable windows (generated by another constraint), and generates an updated set of fusable windows using the current constraint. the old `apply` simply generates a set of fusable windows for this constraint only (the sets of fusable windows for each constraint then have to be coalesced, which is not very efficient)

legate/core/runtime.py
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
...
+        raise NotImplementedError("Implement in derived classes")
+
+
+class NumpyContextExists(FusionConstraint):
+    def apply(self, contexts, runtime, ops, partitioners, strategies):
+        if "cunumeric" in contexts:
+            return True, [(0, len(ops))]
+        else:
+           return False, [(0,0)]
+
```
shivsundram, line 855: This constraint will be removed

legate/core/runtime.py
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
...
+        self.terminals = set()
+        #these ops are almost always fusable
+        self.validIDs.add(2) #Binary op
+        self.validIDs.add(10) #Fill op
+        self.validIDs.add(20) #Where op
+        #self.validIDs.add(6) #Where op
+        #self.validIDs.add(5) #Convert op
+        #self.validIDs.add(7) #dot op
+        #self.terminals.add(7) #dot op is only fusable as a terminal op in a window
+
```
shivsundram, line 873: I initially introduced the notion of a 'terminal' op that is fusable only at the end of a window. I don't use this so will remove it.

legate/core/runtime.py
```diff
@@ -711,14 +729,389 @@ def record_partition(self, index_space, functor, index_partition):
...
+                        start=i
+                        i=start+1
+                        store_to_ops={}
+                        continue
+            i+=1
+        if start<end:
+            intervals.append((start,end))
+        #print("proj", intervals)
+        return True, intervals
+
```
shivsundram, line 965: As stated, the above constraint isn't particularly cheap....and it doesn't seem to ever be "used" I will look into the perf, but I suspect the `get_projection` method is the culprit here

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+        for i in range(len(ops)):
+            self.propogateFuture(ops[i])
+        fusion_checker = FusionChecker(ops, self._contexts, self)
+        fusion_checker.register_constraint(AllValidOps())
+        fusion_checker.register_constraint(IdenticalLaunchShapes())
+        fusion_checker.register_constraint(IdenticalProjection())
+        fusion_checker.register_constraint(ValidProducerConsumer())
+        can_fuse,fusable_sets, partitions = fusion_checker.can_fuse()
+
```
shivsundram, line 1349: fusion constraints are registered above

legate/core/runtime.py
```diff
@@ -880,7 +1286,273 @@ def dispatch(self, op, redop=None):
...
+            else: #strategies already calculated during failed attempt to fuse 
+                for i,op in enumerate(ops):
+                    op.strategy = strats[i]
+            for i,op in enumerate(ops):
+                for input in op._inputs:
+                    if input._storage is None:
+                        frint("launch ufused input", op._task_id, input)
+                self.propogateFuture(op)
+                op.launch(op.strategy)
+
```
shivsundram, line 1538: the above func will be cleaned up and simplified in the next day or so

legate/core/solver.py
```diff
@@ -290,15 +295,17 @@ def cost(unknown):
...
```
shivsundram, line 301: If the store already has a key partition (set by the fuser), no need to compute it 
There may be a more elegant way of doing this

legate/core/runtime.py
shivsundram: TODO: use the alignment data structure to check if the input should be aligned with the output

src/core/data/store.cc
```diff
@@ -121,8 +127,8 @@ Domain FutureWrapper::domain() const { return domain_; }
 ReturnValue FutureWrapper::pack() const
 {
   if (nullptr == rawptr_) {
-    fprintf(stderr, "Found an uninitialized Legate store\n");
-    assert(false);
+    //fprintf(stderr, "Found an uninitialized Legate store\n");
+    //assert(false);
```
shivsundram, line 131: I will come back to this

legate/core/runtime.py
shivsundram: edit: changed to use alignment/constraint info, will be pushed

install.py
```diff
@@ -794,8 +794,8 @@ def driver():
     )
```
shivsundram, line 794: all changes to this file will be reverted before merge

setup.py
```diff
@@ -69,8 +69,8 @@ def run(self):
     # Remove the recurse argument from the list
     sys.argv.remove("--recurse")
     setup(
-        name="legate.core",
-        version="0.1",
+        name="legate-core",
+        version="21.10.00",
```
shivsundram, line 73: Is the current version indeed 0.1? If so I'll undo the above change

src/core/utilities/makeshift_serializer.cc
shivsundram: I'm leaving it in the PR for a few days so you can see it (in case any part of it looks useful), but this will be removed from the PR before merge

legate/core/legion.py
magnatelee: I'll probably generalize this code, as I'm sure this is a utility not just for 32-bit integers.

legate/core/operation.py
magnatelee: This sounds like an auto-partitioner issue: the auto-parallelizer marks a partition as the "key" partition used to distributed tasks across processors. And the auto-parallelizer conveys that information in the tag. With task fusion, the auto-parallelizer shouldn't pick a tag for a particular task, but for the whole window, so that the partition has the same tag throughout the window.

legate/core/launcher.py
```diff
@@ -643,21 +644,35 @@ def set_sharding_space(self, space):
...
+        if is_fused:
+            fusion_metadata.pack(argbuf)
+
+
     def build_task(self, launch_domain, argbuf):
         self._req_analyzer.analyze_requirements()
         self._out_analyzer.analyze_requirements()
 
+        #pack fusion metadata
+        self.pack_fusion_metadata(argbuf, self._is_fused, self._fusion_metadata)
```
magnatelee, line 670: I'm not a fan of this design. The fusion task shouldn't be any different from a normal Legate task. I believe the fusion metadata can be passed as a bunch of scalar arguments to the fusion task. If we do that, then we don't need to make every place in the core handle fusion. Making the default path aware of task fusion doesn't sound like a good design.

legate/core/solver.py
magnatelee: You should check if the key partition satisfies the restrictions before you reuse it. And I think this logic should be moved to `compute_key_partition`. Just so you know, I'm currently refactoring the code that manages legate stores, so don't make any micro optimization like this, as that will become obsolete in the near future.

legate/core/legion.py
shivsundram: makes sense; I found it faster to batch the packing  of ints this way.

src/core/runtime/context.h
```diff
@@ -107,11 +109,18 @@ class LibraryContext {
...
 class TaskContext {
  public:
+  TaskContext() = default;
+
   TaskContext(const Legion::Task* task,
               const std::vector<Legion::PhysicalRegion>& regions,
               Legion::Context context,
               Legion::Runtime* runtime);
 
+  TaskContext(const Legion::Task* task, const std::vector<Legion::PhysicalRegion> regions)
```
magnatelee, line 119: Why is this constructor necessary?

src/core/runtime/context.h
```diff
@@ -121,13 +130,14 @@ class TaskContext {
  public:
   ReturnValues pack_return_values() const;
 
- private:
+ public:
```
magnatelee, line 133: Why did you want to expose the private members here?

shivsundram: ahh this should be reverted. This is an artifact from when we were using the inline launcher.

legate/core/runtime.py
```diff
@@ -880,7 +1122,197 @@ def dispatch(self, op, redop=None):
...
+        """creates a 'header' for a fused op that denotes metadata
+        on each ops inputs, outputs, reductions and scalars
+        """
+        #generate offset maps for all inputs to serialize metadata
+        input_starts, output_starts, offset_starts, offsets= [],[],[],[]
+        reduction_starts, scalar_starts, future_starts, op_ids = [], [], [], []
+        input_start, output_start, offset_start = 0,0,0
+        reduction_start, scalar_start, future_start = 0,0,0
+ 
+        for op in ops:
```
magnatelee, line 1136: I guess we'd eventually want coalescing of arguments here as well. Right now, the fusion code passes the same store used by multiple tasks multiple times, hoping that the launcher at least coalesces region requirements for those duplicates. But, we can identify that the same store is being used in multiple tasks when we scan them for dataflow analysis. So, we could potentially pass a deduplicated list of stores to the fusion task with a mapping between the original indices and those in the store list that is passed to the fusion task.

magnatelee:
Thanks a lot for all your hard work on this PR. I've left several comments and questions, but you don't necessarily need to address all of them. I'll pick this up probably after the break and we can work together to make this ready for the merge.

src/core/runtime/context.h
shivsundram: Each sub-op is called like this  (via function-pointer):
```
for (int i=0; i<fused_op_len; i++){
      auto descriptor = Core::gpuDescriptors.find(opIDs[i])->second;
      descriptor(context);
}
```
However, the `context` above is not the same as the parent tasks's `context`; Whereas the parent task's context contains the "union" of all `input` `outputs` `reductions` and `scalars` over all subops, the above context should only include the `inputs`, `outputs` etc for sub-operation/task `i`. I used this constructor to build the operation-specific contexts, for holding the specific set of `inputs` `outputs` etc. 

This brings up a potential issue of that the new context's `task` and `regions` fields are accordingly not specific to the sub-op. I'm not sure how much of an issue that is, as I only need the context structure to hold stores and scalars for the sub-ops. Ideally I wanted to avoid any performance costs of creating of new Task objects in the Fused Task, so I elected to keep the new `context` object as such.

legate/core/runtime.py
shivsundram: Yeah we could deduplicate these, and I don't imagine this being too difficult. We would just have to store the offset to the dedpulicated store (eg in line 1145) in the `offsets` array, and it should just work

src/core/runtime/context.h
magnatelee: You do need a new context in this case, but my question is more on why you couldn't use the existing constructor. The constructor you added leaves the runtime and context member uninitialized, leading to a crash if the child task tries to access the runtime (though no task in cunumeric is actually doing that). I think the context you are creating here should still have a valid runtime and context pointer.

That said, I think you actually need a slightly different change here, which allows TaskContext to override the legion task's id. The `task` object here is of the fusion task, so if you pass it as-is to a child task, the child task may think that it's not what it thinks it is. I believe no cuNumeric task is doing actually doing it, but leaving this unaddressed can be a foot gun later on.

legate/core/launcher.py
shivsundram: This is def worth discussing (I pondered this myself):
I did think about making them scalars, but was not sure if there were any other implications/downsides of storing all the metadata that way (eg are there a maximum number of scalars we can give a task), since we'd be sending in quite a few scalars to the task (or rather multiple lists, each of which can have >50 scalars; the number of scalars in 4/5 of these lists is equivalent to the fused_op length, even with  potential deduplication of stores).
If we did this, in the task's `scalars` array, we'd thus be mixing "metadata" scalars (which are really a bunch of lists) with the actual scalars used by the  sub-tasks, which seemed to not really adhere to the abstraction of having a dedicated "scalars" array in `context`  data structure.  I thus chose to serialize it as Legate Task  argument data. As you stated, the downside is that the core/default aware has to be "fusion aware".

shivsundram: But if you don't see any potential downsides of making them all scalars, then yeah it could totally be implemented to be scalar based.

src/core/runtime/context.h
shivsundram: Yeah it's a valid point. I initially tried using the normal constructor and got runtime errors with from the task deserializer. Recreating the issue I get: ( the error is from span.h : `Assertion pos < size_ failed.`)
```
legion_python: /home/shiv1/fmultiGPU/legate.core/src/core/utilities/span.h:38: decltype(auto) legate::Span<T>::operator[](size_t) [with T = const Legion::PhysicalRegion; size_t = long unsigned int]: Assertion `pos < size_' failed.
Signal 6 received by node 0, process 3131805 (thread 7f8d47bf3000) - obtaining backtrace
Signal 6 received by process 3131805 (thread 7f8d47bf3000) at: stack trace: 19 frames
  [0] = /lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0) [0x7f8d70ce33c0]
  [1] = /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb) [0x7f8d707f718b]
  [2] = /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b) [0x7f8d707d6859]
  [3] = /lib/x86_64-linux-gnu/libc.so.6(+0x25729) [0x7f8d707d6729]
  [4] = /lib/x86_64-linux-gnu/libc.so.6(+0x36f36) [0x7f8d707e7f36]
  [5] = /home/shiv1/fmultiGPU/legate.core/install/lib/liblgcore.so(+0x5a95d) [0x7f8d4428595d]
  [6] = /home/shiv1/fmultiGPU/legate.core/install/lib/liblgcore.so(legate::TaskDeserializer::_unpack(legate::Store&)+0x10e) [0x7f8d44286d9e]
  [7] = /home/shiv1/fmultiGPU/legate.core/install/lib/liblgcore.so(legate::TaskContext::TaskContext(Legion::Task const*, std::vector<Legion::PhysicalRegion, std::allocator<Legion::PhysicalRegion> > const&, Legion::Internal::TaskContext*, Legion::Runtime*)+0x7f6) [0x7f8d44278626]
  [8] = /home/shiv1/fmultiGPU/legate.core/install/lib/libcunumeric.so(cunumeric::FusedOpTask::gpu_variant(legate::TaskContext&)+0x6e2) [0x7f8d1ce0a852]
```

shivsundram: at the time I didn't spend much time debugging this and just wanted something that worked for the time being, but can dive a little deeper into what's causing this

magnatelee: I suspect the crash is due to the fusion metadata you added that the current serializer doesn't understand, so this issue is tied with the fusion awareness introduced in the code base, which I suggest removing.

marcinz:
@shivsundram @magnatelee What is the status of this PR?

magnatelee:
@marcinz I'm planning to port/rebase this once the planned refactoring on the core is done.

PR #130, opened 2022-01-18
Updates install.py to install to folders not in PYTHONPATH with easy install

m3vaz:


manopapad:
Closing this in favor of https://github.com/nv-legate/legate.core/pull/131, which solves the same problem by skipping `.pth` files and eggs altogether.

It turns out that egg-based installation is problematic for us. Simply adding our custom directory to `PYTHONPATH` (as we do in the legate driver script) is not sufficient. Our custom modules, like `legion_top` and `legate.core`, are stored inside egg directories, that are indexed by `.pth` files:

```
<install-dir>/lib/python3.10/site-packages$ ls 
easy-install.pth  legate.core-0.1-py3.10.egg  legion-0.1-py3.10.egg
<install-dir>/lib/python3.10/site-packages$ cat easy-install.pth
./legion-0.1-py3.10.egg
./legate.core-0.1-py3.10.egg
```

The python interpreter (specifically the `site` module) only processes `.pth` files under `sys.prefix` and `sys.exec_prefix`, so the final `sys.path` doesn't contain the egg directories, and the interpreter can't find our modules. We would have to additionally feed our installation directory to `site.addsitedir()` and run `site.main()` before importing `legion_top`.

The proper way to fix this is to follow standard conventions for installing python packages to custom ("user") directories.

PR #144, opened 2022-02-18
Follow cunumeric convention and use NCCL_PATH vs NCCL_DIR

manopapad:


manopapad:
Closing this in favor of instead using `*_DIR` everywhere, which will come as a later PR.

PR #151, opened 2022-03-01
Support for reduction tree construction

magnatelee:
This PR is to add support for reduction task trees.


lightsighter:
This is just until you have collective instances or do you anticipate switching to collective instances when you have support for them?

magnatelee:
Like I said in the meeting, this is to support the reductions that are not element-wise.

PR #158, opened 2022-03-11
Cleanup / Initial pass for type annotations

bryevdv:
This PR makes some small cleanups: 

* Use new style-classes everywhere
* Remove some obsolete idioms and imports

Additionally, this PR makes an initial pass at adding type annotations, starting with the following modules:
```
legate.core.legion
legate.core.projection
legate.core.types
legate.core.utils
```

For now, type checking is disabled globally, and files are added to an explicit allow-list. Eventually we can just enable type checking on the whole `legate` directory. No modifications to try to add mypy to CI or `pre-commit` were made here. 

Some notes about the approach:

* `from __future__ import annotations` boilerplate was added everywhere. In my experience `__future__` imports should be used everywhere, or nowhere, to prevent the possibility of different behaviours in different modules. 
* For now, had to resort to `Any` for ~5-10% of types. I intend to refine the more complicated types in future PRs. Alot of these are related to CFFI, which is a mess.
* Very basic type stubs for `pyarrow` and `legion_cffi` were added under `typings` at the top level. The legion typings are were generated semi-automated, and are completely generic, but could be refined with a little tedious effort. (Obviously in the ideal case n the future, legion provides its own types)
* Imported `TypeAlias` from `typing_extensions`, which seems to already be in my `legate` conda env, but should make the dependency explicit to be completely safe (or remove use of `TypeAlias`)
* I generally only use `Optional` for parameters, not e.g. for returns or attributes.

Hopefully getting the biggest `legion.py` module passing `mypy` will clear the way for other modules to be updated more smoothly. But I wanted to pause here for review and discussion of any concerns with the approach, etc. cc @magnatelee 

All `cunumeric` tests with `--use-cuda` pass for me. Mypy picked up a few small issues here and there, I will note them in inline comments below (apologies for the one giant file to review, in hindsight I wish I had worked to split up `legion.py` *first*, however commits are split up in case you want to review them individually)


legate/core/legion.py
```diff
@@ -3149,15 +3338,17 @@ def __init__(
...
-        if not isinstance(fields, list):
-            fields = [fields]
-        for field in fields:
-            self.add_field(field)
+        if fields is not None:
+            if isinstance(fields, list):
+                for field in fields:
+                    self.add_field(field)
+            else:
+                self.add_field(fields)
```
bryevdv, line 3346: mypy just got confused here without explicit types and/or new assignment vars

legate/core/legion.py
```diff
@@ -5008,27 +5237,29 @@ def pack_buffer(self, buf):
     # Static member of this class for encoding dtypes
     _dtype_codes = {
         bool: legion.LEGION_TYPE_BOOL,
-        np.bool: legion.LEGION_TYPE_BOOL,
+        # np.bool: legion.LEGION_TYPE_BOOL,
```
bryevdv, line 5240: `np.bool` is just and alias for `bool` so I think including it in the dict is no-op (it is also deprecated as of np 1.20) Similar comments for `int`, `float` below.

bryevdv:
Some comments left. If there are any types that look wrong, or `Any` types that could be quickly refined, please let me know.

legate/core/legion.py
```diff
@@ -1504,17 +1654,27 @@ def get_child(self, index_partition):
...
-    def get_root(self):
+    def get_root(self) -> Region:
         """
         Return the root Region in this tree.
         """
-        if not self.parent:
-            return self
-        return self.parent.parent.get_root()
+        if self.parent is not None:
+            return self.parent.parent.get_root()
```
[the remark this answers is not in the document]
manopapad: A Region has Partition-type objects as children (one for each way that the Region has been partitioned), and each Partition has Region-type objects as children (one for each sub-region in the partition). Therefore it's expected that you can can go up or down 2 steps at a time (if starting from a Region object).

legate/core/legion.py
```diff
@@ -2696,7 +2862,8 @@ def __init__(
         legion.legion_memory_query_destroy(query)
         legion.legion_machine_destroy(machine)
         for (sub_region, buf) in shard_local_data.items():
-            assert sub_region.parent.parent is parent
+            if sub_region.parent is not None:
+                assert sub_region.parent.parent is parent
```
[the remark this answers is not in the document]
manopapad: I assume assertions were used when it's an internal error. Anything in the legion module is only supposed to be used by legate developers, not users.

legate/core/legion.py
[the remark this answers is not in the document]
bryevdv: @manopapad The initializer for `Region` only sets `self.parent = parent`, but the default parameter value for `parent` is None, so it's definitely *possible* to construct a `Region` with `r.parent = None`. This is why mypy flagged the error. Given your comment, I would suggest that possibly the `parent` value to `Region` should be made non-optional. However, the `__del__` has a check that seems to indicate that None is an expected/allowed value:
```
    def __del__(self) -> None:
        if self.owned and self.parent is None:
            self.destroy(unordered=True)
```
And similar in `Region.destroy`:
```
        if self.parent is not None:
            raise RuntimeError("Only root Region objects can be destroyed")
```
So is it the case that `Region` should usually have a non-Null parent, except in specific "root" cases?

legate/core/legion.py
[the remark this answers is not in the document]
bryevdv: Projects will often define their own `InternalError` or similar for this purpose (e.g. [here](https://www.tensorflow.org/api_docs/python/tf/errors/InternalError), [here](https://dev.mysql.com/doc/connector-python/en/connector-python-api-errors-internalerror.html), or [here](https://www.adamsmith.haus/python/docs/keras.losses.K.ctc.gen_ctc_ops._errors.InternalError)) This allows for the possibility of attaching more detailed diagnostic information or state information to the error (which can be used by legate developers, or reported by users making issues, improving the quality of bug reports)

legate/core/utils.py
```diff
@@ -8,15 +8,19 @@
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# WITHOUT WARRANTIES OR CONDITIONS OF Hashable KIND, either express or implied.
```
manopapad, line 11: Search & replace was a little too eager here :-)

legate/core/context.py
```diff
@@ -29,16 +30,16 @@ class ResourceConfig(object):
...
         self.max_mappers = 1
         self.max_reduction_ops = 0
         self.max_projections = 0
         self.max_shardings = 0
 
 
-class ResourceScope(object):
-    def __init__(self, context, base, category):
+class ResourceScope:
+    def __init__(self, context, base, category) -> None:
```
manopapad, line 42: Just curious, is `-> None` implicit if not specified? It seems to be added onto some `__init__` methods but not on others.

legate/core/utils.py
bryevdv: Hah! Thanks for the catch 👍

manopapad:
Just to be aware, our minimum supported python version at the moment is 3.7, in case that restricts the types you'd want to use.

legate/core/context.py
bryevdv: It's not implicit, but this file itself is not yet checked by mypy. I did add a few types here provisionally to make checking  the other files more straightforward.

legate/core/legion.py
```diff
@@ -178,49 +180,51 @@ def __init__(self, p=None, dim=None):
...
+    def __hash__(self) -> int:
         value = hash(self.dim)
         for idx in range(self.dim):
             value = value ^ hash(self[idx])
         return value
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Point):
+            return False
```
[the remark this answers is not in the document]
manopapad: Is this the common practice or `type(self) == type(other)`? I have used explicit type equality in the past, because (I think) that avoids the weird situation where `superclass_instance == subclass_instance` but  `subclass_instance != superclass_instance`.

legate/core/utils.py
bryevdv: https://github.com/nv-legate/legate.core/pull/158/commits/2da2c8560bc27babaa1ec312143cc0ce5b4863ab

legate/core/legion.py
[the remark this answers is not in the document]
manopapad: That's correct, the root of the tree will be a `Region` with null `parent`. Every other `Region` on the same tree should have a `Partition` parent, which in turn has a non-optional `Region` parent.

bryevdv:
> Just to be aware, our minimum supported python version at the moment is 3.7, in case that restricts the types you'd want to use. 

Sigh. Right, you just made me realize I am using built-ins `tuple`, `dict`, etc out of habit, rather than `Tuple` and `Dict` from `typings`.  I had actually briefly discussed just this with @magnatelee in regard to using `Foo | Bar` instead of `Union[Foo, Bar]`. I can definitely update these, but it's worth also providing some context.

Currently I have added `from __future__ import annotations` to all files. What this means is that type annotations are now evaluated lazily, and are *not* evaluate unconditionally at runtime. It also means that things like using `tuple` instead of `Tuple` work with mypy.  Long story short, because of the lazy evaluation, there is absolutely no problem using the library with Python 3.7 for "normal usage". 

Is there* some* usage that might be an issue? Yes, from the [docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563) 

> Some libraries may have use cases for dynamic evaluation of annotations, for instance, through use of typing.get_type_hints or eval. If your annotation would raise an error when evaluated (say by using [PEP 604](https://www.python.org/dev/peps/pep-0604) syntax with Python 3.9), you may need to be careful when using such libraries.)

So, the only folks who might run in to this are developers using tools like `mkdocstrings` and `pytkdocs` with older pythons, on this codebase. My own position is that non-typical-users trying to use non-standard dev/docs tools on a codebase is out of scope, or at least worth the expense of making more work work for OSS dev. But that opinion is also informed by experience with volunteer OSS. :) 

However, I understand the desire to be as conservative as possible, though in that case we might even want to avoid `from __future__ import annotations` altogether. It's not a change I would make but please let me know thoughts @manopapad @magnatelee

manopapad:
> using `Foo | Bar` instead of `Union[Foo, Bar]`

That was actually the primary usecase I had in mind. If `from __future__ import annotations` allows us to use `|` I would be happy to use that instead of `Union`.

> My own position is that non-typical-users trying to use non-standard dev/docs tools on a codebase is out of scope

I don't have the experience to know how common this is, but I agree with your logic.

legate/core/legion.py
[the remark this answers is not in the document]
bryevdv: OK so there is a special case where it's *not* possible to go two steps up. I'm lacking some context to know: is the guard I added above appropriate? Or is it a problem altogether if this codepath is reached with a root region?

bryevdv:
I should explicitly mention the one case that might reasonably come up (which [recently came up for Bokeh](https://github.com/bokeh/bokeh/issues/11637)). If another library dev were to build directly on top of our libraries, and subclass our classes, then using `__future__` annotations might make life more annoying for them. So the question is: is there any likelihood of other projects building directly on top of these, and if so, does that scenario need to be catered to by us?

bryevdv:
@magnatelee do you have insight on the CI failure? It seems to be related to an expired SSH key?

manopapad:
@marcinz fixed the CI failure.

legate/core/legion.py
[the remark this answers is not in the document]
bryevdv: I've definitely seen both, and sometimes even in the same project (both are present inside the Numba codeabse, e.g.) The main motivation to switch is because `mypy` is smart enough to reason about the `isinstance` check, but not smart enough to reason about the other check. We would need a casts for that case, which is not great but also not the end of the world.

However, you encouraged me to dig a bit and I think the actual solution here might simply be to return [`NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented) instead of False.
> When a binary (or in-place) method returns NotImplemented the interpreter will try the reflected operation on the other type

That should make `sub == super` so long as `super == sub`

<details> 

```
In [1]: class sup:
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...:     def __eq__(self, other):
   ...:         if not isinstance(other, sup):
   ...:             return NotImplemented
   ...:         return self.a == other.a
   ...: 

In [2]: class sub(sup):
   ...:     def __init__(self, a, b):
   ...:         super().__init__(a)
   ...:         self.b = b
   ...:     def __eq__(self, other):
   ...:         if not isinstance(other, sub):
   ...:             return NotImplemented
   ...:         return self.a == other.a and self.b == other.b
   ...: 

In [3]: P = sup(10)

In [4]: C = sub(10, 20)

In [5]: P == C
Out[5]: True

In [6]: C == P
Out[6]: True
```
</details>

I will make push this change if that seems good to you. All this said, except for the partition functors, all the classes in `legion.py` are concrete.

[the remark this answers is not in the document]
bryevdv: Ah, and style is exactly what mypy itself recommends if you set a bad type for `other`:
```
egate/core/legion.py:573:5: note: This violates the Liskov substitution principle
legate/core/legion.py:573:5: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
legate/core/legion.py:573:5: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
legate/core/legion.py:573:5: note:     def __eq__(self, other: object) -> bool:
legate/core/legion.py:573:5: note:         if not isinstance(other, AffineTransform):
legate/core/legion.py:573:5: note:             return NotImplemented
legate/core/legion.py:573:5: note:         return <logic to compare two AffineTransform instances>

```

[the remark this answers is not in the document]
bryevdv: https://github.com/nv-legate/legate.core/pull/158/commits/ccfd31f1aeb43bead06d26f152624f0e08ab2495

legate/core/legion.py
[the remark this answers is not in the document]
manopapad: The way you have it is fine. If it's easier on mypy, It would also be fine to do the following, given that `Partition` also defines `get_root()`:

```
    def get_root(self) -> Region:
        """
        Return the root Region in this tree.
        """
        if self.parent is not None:
            return self.parent.get_root()
        return self
```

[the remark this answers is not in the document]
bryevdv: :+1: https://github.com/nv-legate/legate.core/pull/158/commits/168bedb92a1091b0469fe600cfd544b2dc5660e4

bryevdv:
Apart from `conda/core_dev.yml` environement file, it's unclear to me where runtime dependencies are specified for package builds. I've removed `TypeAlias` to avoid the `typing_extensions` module for simplicity for now.

legate/core/projection.py
```diff
@@ -12,34 +12,38 @@
...
-    def __init__(self, dim=-1, weight=1, offset=0):
+class ProjExpr:
+    def __init__(
+        self, dim: int = -1, weight: int = 1, offset: int = 0
+    ) -> None:
         self._dim = dim
         self._weight = weight
         self._offset = offset
-        self._repr = None
+        self._repr: Union[str, None] = None
```
magnatelee, line 29: What's a typical way to express a nullable type in the Python typing parlance? Coming from a PL background, I'd find `Optional[str]` more comfortable and `Union` reserved for a sum type, though I guess `Union[str, None]` is equivalent. I'm just curious to know if there's any reason to prefer `Union` over `Optional` for cases like this.

bryevdv: `Optional[str]` would certainly work here. I personally prefer to reserve `Optional` for parameters with default None values, i.e. where it is optional for a caller to pass a value or not. To me, a value like this is not optional in that sense, it simply a union type that can hold either `str` or `None`. This is not a super strong opinion, though, so I am happy to change all `Union[..., None]` to `Optional[...]` if that's what others prefer.

legate/core/legion.py
```diff
@@ -241,12 +243,18 @@ def set_point(self, p):
...
+    def raw(self) -> Any:
         return self.point
 
 
-class Rect(object):
-    def __init__(self, hi=None, lo=None, exclusive=True, dim=None):
+class Rect:
+    def __init__(
+        self,
+        hi: Optional[list[float]] = None,
```
magnatelee, line 253: Does `list` mean literally a `list` object or does it encompass other iterables? I remember somewhere I passed a `Shape` object to construct `Rect`.

bryevdv: `list` means `list`. It is the the newer replacement for `typing.List` If other iterable types are expected then `Iterable[float]` is probably better

legate/core/legion.py
```diff
@@ -904,7 +1019,7 @@ def partition(self, runtime, context, parent, color_space, kind, part_id):
 
 
 class PartitionByWeights(PartitionFunctor):
-    def __init__(self, weights):
+    def __init__(self, weights: Union[FutureMap, Any]) -> None:
```
magnatelee, line 1022: I believe we want `weights` to be always `FutureMap`. Was there a place where an object of type `Any` is actually being passed? If so, I think we should fix it.

bryevdv: Line 1050 has a branch:
```
 elif isinstance(self.weights, list):
```
which mypy will flag as unreachable code with just `FutureMap`:
```
legate/core/legion.py: note: In member "partition" of class "PartitionByWeights":
legate/core/legion.py:1050:9: error: Statement is unreachable  [unreachable]
            elif isinstance(self.weights, list):
            ^
```
Should the branch be removed entirely?

bryevdv: Also just FYI any instance of `Union[..., Any]` is a placeholder that is meant to be refined.

legate/core/legion.py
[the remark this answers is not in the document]
magnatelee: > sub_region.parent can technically None. Perhaps there is better solution, though.

It would be nice if we could actually enforce this at the typing level, say making a subtype `SubRegion` which does actually the same thing as `Region` but always has a non-null `parent`. Then the caller needs to do a casting to make sure the argument has a right type.

> I assume assertions were used when it's an internal error. Anything in the legion module is only supposed to be used by legate developers, not users.

@manopapad unless this class is consumed completely within Legate core, I suggest we instead raise an exception with a more meaningful message so that the library developer can learn the right usage of this API.

legate/core/legion.py
magnatelee: Ok. Can you actually make the change instead of commenting out those lines?

legate/core/projection.py
magnatelee: I guess I'm fine with `Union`, as long as the same convention is used throughout the code, so no need for changing the code for this.

legate/core/legion.py
magnatelee: Let's make it `Sequence` then. I don't see a reason to limit the type to be just `list` here.

legate/core/legion.py
bryevdv: :+1:  https://github.com/nv-legate/legate.core/pull/158/commits/00ff8701c9005dc9947b6875b5c6b25764c95e59

legate/core/legion.py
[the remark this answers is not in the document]
bryevdv: :+1: to both those suggestions but I given the scope of changes already in this PR I would suggest and issue for now and visiting after `legion.py` is split up

legate/core/legion.py
magnatelee: Ah, didn't realize we had that branch as we never really used it. then I guess `Union[FutureMap, list[int]]` would be more appropriate.

legate/core/legion.py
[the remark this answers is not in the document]
magnatelee: sounds good to me.

legate/core/legion.py
bryevdv: https://github.com/nv-legate/legate.core/pull/158/commits/0daeb173635d99e271aa7254549737d776af414b

legate/core/legion.py
bryevdv: https://github.com/nv-legate/legate.core/pull/158/commits/3318038b119b2369a00999f79ae6f93dd8a48c28

legate/core/projection.py
bryevdv: :+1:  In the future I would actually advocate for `Foo | None` everywhere (less to type, less to import)

magnatelee:
> > Just to be aware, our minimum supported python version at the moment is 3.7, in case that restricts the types you'd want to use.
> 
> Sigh. Right, you just made me realize I am using built-ins `tuple`, `dict`, etc out of habit, rather than `Tuple` and `Dict` from `typings`. I had actually briefly discussed just this with @magnatelee in regard to using `Foo | Bar` instead of `Union[Foo, Bar]`. I can definitely update these, but it's worth also providing some context.
> 
> Currently I have added `from __future__ import annotations` to all files. What this means is that type annotations are now evaluated lazily, and are _not_ evaluate unconditionally at runtime. It also means that things like using `tuple` instead of `Tuple` work with mypy. Long story short, because of the lazy evaluation, there is absolutely no problem using the library with Python 3.7 for "normal usage".
> 
> Is there _any_ usage that might be an issue? Yes, from the [docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563)
> 
> > Some libraries may have use cases for dynamic evaluation of annotations, for instance, through use of typing.get_type_hints or eval. If your annotation would raise an error when evaluated (say by using [PEP 604](https://www.python.org/dev/peps/pep-0604) syntax with Python 3.9), you may need to be careful when using such libraries.)
> 
> So, the only folks who might run in to this are developers using tools like `mkdocstrings` and `pytkdocs` with older pythons, on this codebase. My own position is that non-typical-users trying to use non-standard dev/docs tools on a codebase is out of scope, or at least worth the expense of making more work work for OSS dev. But that opinion is also informed by experience with volunteer OSS. :)
> 
> However, I understand the desire to be as conservative as possible, though in that case we might even want to avoid `from __future__ import annotations` altogether. It's not a change I would make but please let me know thoughts @manopapad @magnatelee

As long as the features available in the later Python version are used only for typing purposes and do not affect the users with older Python versions, I don't want to apply the same level of conservatism to ourselves, as we can control which Python version to use for our development.

Issue #156, opened 2022-03-08
Start Python Before Legion


lightsighter:
Today, the Legate driver script will always run `legion_python` to execute Legate programs using Legion. Alternatively, we could start the Python interpreter first in each process and then start Legion when the `legate.core` module is loaded. This would allow Legate programs to be run like normal Python programs on a single node, or in the case of multi-node execution, to be run like `mpirun -n N python script.py`. That might be preferable for some users, but also will require some work to make it happen and comes with some caveats.

* Machine configuration parameters: today users pass machine configuration parameters on the command line. The Legate driver script converts those into Realm machine command line options. When using `legion_python`, we ensure that those command line parameters are stripped out before the script starts and becomes visible through `sys.argv` in Python. That prevents users from having to know how to ignore our arguments. In the case of starting Python first, machine configuration flags would still be visible on `sys.argv` at least until we start Legion.
* Some Python operations may need to turn around and launch sub-tasks which also are going to need to run on Python processors, such as `numpy.vectorize` which will need to execute a user-defined function. To handle such cases, we need to have support in Realm for drafting external resources as custom processors. Specifically in this case, we need to be able to tell Realm that it can treat the main thread with the Python interpreter as an instance of a Python processor on which Python sub-tasks can be executed. There are some tricky open questions here, such as how to know when it's safe to execute sub-tasks on the drafted processor, e.g. by noticing that the implicit task has paused. Legion can provide some of that context, but it's not a general Realm solution. The concept of an implicit-task currently only exists in Legion. More details: https://github.com/StanfordLegion/legion/issues/716
* Allowing Python to start before Legion will pollute the global address space with all of the Python interpreter's icky global variables. Once that occurs, it's impossible for the work on subprocess in Legion and Realm to provide support for multiple Python interpreters per node as the pollution of the global address space from the Python interpreter will just flow down into any subprocesses. More details: https://github.com/StanfordLegion/legion/issues/627

PR #164, opened 2022-03-16
Quickfix: Fix non-annotations on 3.7

bryevdv:
I did not realize Python 3.7 is not currently run in CI. Many thanks to @manopapad for noticing that there were issues in 3.7 due to `__future__` features unintentionally being used outside annotation contexts in a few places. The PR fixes those up, with TODOs for later. I have installed everything and run tests in a 3.7 env and will continue to develop under 3.7 for the time being. cc @magnatelee

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@
 legate/core/install_info.py
 /build
 /legion
-/install
+/install*
 config.mk
 /docs/legate/core/build
 *.egg-info
--- a/legate/core/legion.py
+++ b/legate/core/legion.py
@@ -17,7 +17,15 @@ from __future__ import annotations
 import os
 import struct  # For packing and unpacking C data into/out-of futures
 import weakref
-from typing import TYPE_CHECKING, Any, Iterator, Optional, Sequence, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+    Union,
+)
 
 if TYPE_CHECKING:
     from .context import Context
@@ -1524,7 +1532,8 @@ class FieldID:
         return self._type
 
 
-FieldListLike = Union[int, FieldID, list[int], list[FieldID]]
+# todo: (bev) use list[...] when feasible
+FieldListLike = Union[int, FieldID, List[int], List[FieldID]]
 
 
 class Region:
--- a/legate/core/projection.py
+++ b/legate/core/projection.py
@@ -14,7 +14,7 @@
 #
 from __future__ import annotations
 
-from typing import Any, Callable, Optional, Union
+from typing import Any, Callable, Optional, Tuple, Union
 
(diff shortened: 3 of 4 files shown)


magnatelee:
@manopapad thanks for catching the bug. @bryevdv thanks for a quick fix! I think it's bad that we're claiming we support Python 3.7 and not actually testing with it.

@marcinz @m3vaz I believe we just pick whatever version of Python that satisfies the conda constraints for the CI image. How hard is it to run tests with more than one Python version? I think at the very minimum we should test with two versions of Python, the oldest one and probably the "default" one that comes with the latest conda. Any thoughts?

magnatelee:
A side note: #156 can make it much easier for us to test our code with multiple Python versions; we then just need to switch from one Python env to another. One caveat though is that certain features that launch Python tasks cannot be tested if the Python version Realm was compiled with is different from the interpreter's.

manopapad:
> I believe we just pick whatever version of Python that satisfies the conda constraints for the CI image.

No, currently we are explicitly asking for python 3.8 in the CI images. On my local environment I'm using 3.7, so I assume CI images with 3.7 will also work fine, but I could be wrong.

> How hard is it to run tests with more than one Python version? I think at the very minimum we should test with two versions of Python, the oldest one we're claiming to support and probably the "default" one that comes with the latest conda. Any thoughts?

Given that we don't conditionally use stuff from later python versions, IMHO it should be enough to run CI on just one image, using our minimal supported python version.

magnatelee:
> No, currently we are explicitly asking for python 3.8 in the CI images. On my local environment I'm using 3.7, so I assume CI images with 3.7 will also work fine, but I could be wrong.

Then, I have to ask why we were asking for Python 3.8 and not 3.7... We should make it consistent with our minimum supported version at least.

> Given that we don't conditionally use stuff from later python versions, IMHO it should be enough to run CI on just one image, using our minimal supported python version.

I want to remind you that the shutdown hang we had to fix for the previous release showed up only in the later Python versions. Issues like that might be rare enough that we don't need to worry too much, but we also don't have any control on what the Python folks decide to change down the road. So, if it's not too costly, I feel we should test at least the oldest and the latest Python versions we support.

bryevdv:
@magnatelee all of the above considered, it also just occurs to me that the tests that would catch compatibility issues are in the cunumeric repo, not in this one. Unless I am missing something, all the CI here does is build? Is there a reason not to add a basic unit test suite here directly? Even just importing all the modules would be a start.

magnatelee:
@bryevdv You're right. The CI for this repo only checks if the code builds correctly. I suggested to @marcinz that any change to the core should also run tests for at least one of the downstream libraries, but I don't think that workflow is implemented yet. A unit test suite for Legate core sounds like a good idea. We just haven't had enough bandwidth to do all this.

> Any objection to merging this now, to get others unstuck?
I have no objection to the merge.

PR #168, opened 2022-03-23
added allocator

mfoerste4:
added allocator that can be used within thrust execution policies in order to provide memory via DeferredBuffer.


src/core/utilities/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+  DeferredBufferAllocator(Legion::Memory::Kind kind);
+  virtual ~DeferredBufferAllocator();
+
+ public:
+  typedef char value_type;
+  char* allocate(size_t bytes);
+  void deallocate(char* ptr, size_t n);
+
+ private:
+  Legion::Memory::Kind target_kind{Legion::Memory::Kind::SYSTEM_MEM};
```
magnatelee, line 40: Let's use the same convention as the others for member names; please append `_` to the class members.

src/core/utilities/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include "core/data/buffer.h"
+
+#include <unordered_map>
+
+namespace legate {
+
+class DeferredBufferAllocator {
```
magnatelee, line 25: Maybe `IntraTaskAllocator` is a better name I guess? We don't want developers to use this for buffers that are returned to unbound stores.

src/core/data/allocator.cc
```diff
@@ -0,0 +1,55 @@
...
+  for (auto& pair : buffers_) { pair.second.destroy(); }
+  buffers_.clear();
+}
+
+char* IntraTaskAllocator::allocate(size_t bytes)
+{
+  if (bytes == 0) return nullptr;
+
+  // Use 16-byte alignment
+  bytes = (bytes + 15) / 16 * 16;
```
magnatelee, line 34: Let's make the alignment configurable. We can take this as a constructor argument.

Also, let's pass the alignment to `DeferredBuffer`'s constructor: https://github.com/StanfordLegion/legion/blob/stable/runtime/legion.h#L3223, instead of computing the aligned size manually here. You would need to change `create_buffer` to take the alignment.

src/core/data/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+ public:
+  using ByteBuffer = Buffer<int8_t>;
+
+ public:
+  IntraTaskAllocator() = default;
+  IntraTaskAllocator(Legion::Memory::Kind kind);
+  virtual ~IntraTaskAllocator();
+
+ public:
+  typedef char value_type;
```
magnatelee, line 35: Let's use `using` instead.

src/core/data/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+  char* allocate(size_t bytes);
+  void deallocate(char* ptr, size_t n);
```
magnatelee, line 37: Is this the signature that Thrust expects? Otherwise, I think we should change it to `void (void*, size_t)`

src/core/data/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+  using ByteBuffer = Buffer<int8_t>;
+
+ public:
+  IntraTaskAllocator() = default;
+  IntraTaskAllocator(Legion::Memory::Kind kind);
+  virtual ~IntraTaskAllocator();
+
+ public:
+  typedef char value_type;
+  char* allocate(size_t bytes);
```
magnatelee, line 36: The same applies here. You may want to change the return type to `void*`, unless Thrust forbids it.

src/core/data/allocator.h
magnatelee: Also, is the contract for a custom allocator for Thrust? Otherwise, we shouldn't add this member.

src/core/data/allocator.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include "core/data/buffer.h"
+
+#include <unordered_map>
+
+namespace legate {
+
+class IntraTaskAllocator {
```
magnatelee, line 25: I have second thoughts on the naming, and this really should be `ScopedAllocator`, as that's what this allocator does. Also, we should make a warning that allocations are valid only within the scope that this allocator is alive; i.e., as soon as this allocator goes out of scope and its destructor runs, memory space for the allocations can be recycled.

src/core/data/allocator.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "core/data/allocator.h"
+
+namespace legate {
+
+IntraTaskAllocator::IntraTaskAllocator(Legion::Memory::Kind kind) : target_kind_(kind) {}
+
+IntraTaskAllocator::~IntraTaskAllocator()
+{
+  for (auto& pair : buffers_) { pair.second.destroy(); }
```
magnatelee, line 25: Like I said in the previous comment, this will make the allocations escaped the scope unsafe to use. I think we should add an optional argument, say `scoped`, to the constructor that can control this behavior. When `scoped` is true, the destructor destroys all buffers in the destructor, otherwise, it lets them alive, allowing allocations to escape the scope. The latter is not necessarily unsafe, given that these buffers will be destroyed anyway when the task finishes. I think the default should be `scoped=true`.

src/core/data/allocator.h
mfoerste4: Yes, this is what thrust expects.

src/core/data/allocator.h
mfoerste4: Yes, this is what thrust expects as far as i understand. I did not get it to work with void

src/core/data/allocator.h
mfoerste4: Same as above

src/core/data/allocator.h
magnatelee: Ok. That still isn't a good excuse to use `char*`, as this isn't just meant for Thrust and using `char*` will cause trouble in other contexts. I think you should change it to `void*` and make an adaptor/wrapper to appease Thrust.

src/core/data/allocator.h
magnatelee: The same applies here.

src/core/data/allocator.cc
mfoerste4: 16byte seems to be default for DeferredBuffers anyway - so this manual adjustment is not needed. Do we want to define a default in create_buffer as well or do we want additional constructors with the alignment parameter?

magnatelee: You can make the argument optional in both the allocator's constructor and `create_buffer`, with the default value being 16.

src/core/data/allocator.h
mfoerste4: I removed all type specifications from the generic allocator. The consumer now has to specify a small wrapper class that derives from it and adds the types.

src/core/data/allocator.h
mfoerste4: Agreed. I did change the naming and added a parameter to control destruction behavior. What kind of warning are you thinking about? I suppose we don't want a runtime warning, right?

magnatelee: By warning I meant adding a comment to this class.

src/core/data/allocator.cc
```diff
@@ -0,0 +1,56 @@
...
+
+  buffers_[ptr] = buffer;
+  return ptr;
+}
+
+void ScopedAllocator::deallocate(void* ptr)
+{
+  ByteBuffer buffer;
+  auto finder = buffers_.find(ptr);
+  if (finder == buffers_.end()) return;
```
magnatelee, line 50: I this should be an error instead of being ignored silently. The allocator is not supposed to deallocate something that wasn't created by it.

install.py
```diff
@@ -236,6 +236,7 @@ def build_legion(
     cuda,
     arch,
     openmp,
+    march,
```
magnatelee, line 239: This and other changes in install.py and legate.mk are from another PR that this PR shouldn't include. Please clean them off.

mfoerste4: The rebase is somehow messed up. When I do a manual diff of the two heads (22.05 and this branch) these files don't show up as the corresponding changeset has already been merged (even before I did the rebase). I can't revert this from my branch as this would also remove it from the main branch when merged. I can create a new branch with cherry-picked changesets, but this would require a new PR as well.

PR #181, opened 2022-04-02
Add more type annotations

bryevdv:
This PR adds type annotations to:
```
legate.core.corelib
legate.core.install_info
legate.core.partition
legate.core.shape
legate.core.transform
```

Some Notes:

* `typing_extensions` was added to dependencies to support the use of `@overload` on `Shape.__getitem__`. Without it. it would be necessary to litter explicit `cast` calls everywhere a `Shape` was indexed
* As long as it is added, `TypeAlias` and `Protocol` (see comments below) were also used
* There was an issue with `get_inverse_transform`, sometimes it returned `AffineTransform` and other times `tuple[AffineTransform, int]`. Codepaths were updated to only return `AffineTransform` but special attention should focused on this for review.
* I turned `Transform` into a `Protocol` since it seemed as though using it as an defined abstract base had been explicitly avoided. Or perhaps it was intended as a Mixin? In any case its `serialize` method assumed a `self._runtime` but not all subclasses initialize that. I moved the code out to subclasses as appropriate since it was only a line or two. We could factor the "code" serialization into a helper that accepts a runtime if desired, though.
* `TransformStack` and `IdentityTransform` now inherit the protocol

cc @magnatelee


legate/core/corelib.py
```diff
@@ -15,39 +15,42 @@
...
         config = ResourceConfig()
         config.max_tasks = self._lib.LEGATE_CORE_NUM_TASK_IDS
         config.max_projections = self._lib.LEGATE_CORE_MAX_FUNCTOR_ID
         config.max_shardings = self._lib.LEGATE_CORE_MAX_FUNCTOR_ID
         return config
 
-    def destroy(self):
-        self._lib.legate_shutdown()
+    def destroy(self) -> None:
+        if self._lib:
```
bryevdv, line 55: mypy doesn't really have any way to know a post-init call to `initialize` happened. This seemed reasonable, or we could test and  raise an error.

legate/core/partition.py
```diff
@@ -118,55 +131,59 @@ def __eq__(self, other):
...
+    def __hash__(self) -> int:
+        if self._hash is not None:
+            return self._hash
+
+        self._hash = hash(
+            (
+                self.__class__,
+                self._tile_shape,
+                self._color_shape,
+                self._offset,
```
bryevdv, line 166: This could probably be reverted but all things considered I prefer the early return / less indentation

legate/core/shape.py
```diff
@@ -14,38 +14,48 @@
...
+Extentable: TypeAlias = Union["Shape", int, Iterable[int]]
+
+
+def _cast_tuple(value: Extentable, ndim: int) -> tuple[int, ...]:
     if isinstance(value, Shape):
         return value.extents
-    elif isinstance(value, tuple):
-        return value
+    elif isinstance(value, Iterable):
+        return tuple(value)
```
bryevdv, line 33: With these changes, `_cast_tuple` can be used in place of duplicated code in `Shape.__init__`

legate/core/shape.py
```diff
@@ -55,57 +65,71 @@ def extents(self):
...
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.extents)
 
+    def __iter__(self) -> Iterator[int]:
+        return iter(self.extents)
+
+    def __contains__(self, value: object) -> bool:
+        return value in self.extents
```
bryevdv, line 98: Some places expect `Sized[Shape]` and other expect `Iterable[Shape]`. The simplest protocol with both is `Collection[Shape]` but that also requires `__contains__`.

legate/core/shape.py
```diff
@@ -14,38 +14,48 @@
...
+from typing import TYPE_CHECKING, Iterable, Iterator, Optional, Union
 
+from typing_extensions import TypeAlias, overload
 
-def _cast_tuple(value, ndim):
+if TYPE_CHECKING:
+    from . import IndexSpace
+    from .runtime import Runtime
+
+Extentable: TypeAlias = Union["Shape", int, Iterable[int]]
```
bryevdv, line 26: Happy to know of a better name for this

legate/core/transform.py
```diff
@@ -382,62 +450,61 @@ def convert(self, partition):
...
-                == 0
-            ):
+            color_shape = partition.color_shape[
+                self._dim + 1 : self._dim + self._shape.ndim
+            ]
+            offset = partition.offset[
+                self._dim + 1 : self._dim + self._shape.ndim
+            ]
+
+            if color_shape.volume() == 1 and offset.sum() == 0:
```
bryevdv, line 507: In general I will try to optimize to make conditions short/readable on their own (esp with  `black` formatting and a short line limit)

legate/core/partition.py
magnatelee: Sounds good to me.

legate/core/shape.py
magnatelee: `ExtentListLike`?

legate/core/shape.py
magnatelee: That's great.

legate/core/types.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, dtype: Any, size_in_bytes: int, code: int) -> None:
         self._redop_ids: dict[int, int] = {}
 
     @property
-    def type(self) -> _Dtype:
+    def type(self) -> Any:
```
magnatelee, line 83: Have you encountered a case where this doesn't return a `_Dtype`?

magnatelee:
@bryevdv left several comments. Otherwise, looks good to me. Here are my answers to some of your comments:

> There was an issue with get_inverse_transform, sometimes it returned AffineTransform and other times tuple[AffineTransform, int]. Codepaths were updated to only return AffineTransform but special attention should focused on this for review.

I think that was due to context insensitivity of mypy's flow analysis. Both `Transform` and `TransformStack` have that method, which is a pure coincidence, but `Transform` is used only by `TransformStack`, which is a client-facing abstraction. This means that if you later start adding annotations to `Store`, you will need to make sure that `_transform` of every `Store` only points to either `TransformStack` or `IdentityTransform`. 

> I turned Transform into a Protocol since it seemed as though using it as an defined abstract base had been explicitly avoided. Or perhaps it was intended as a Mixin? In any case its serialize method assumed a self._runtime but not all subclasses initialize that. I moved the code out to subclasses as appropriate since it was only a line or two. We could factor the "code" serialization into a helper that accepts a runtime if desired, though.

It should have been an ABC, though it was like a mixin when I wrote it. The refactoring you did looks good.

legate/core/corelib.py
bryevdv: change to raise a `RuntimeError ` in https://github.com/nv-legate/legate.core/pull/181/commits/ae84ae5938bc75ba6646b2f96d0fb75931e7891b

legate/core/shape.py
bryevdv: is `Shape` a "list like" ? Maybe just `ExtentLike`?

bryevdv: https://github.com/nv-legate/legate.core/pull/181/commits/a1fb957ba4c02513af6915c7ff712f44270d2bad

PR #188, opened 2022-04-07
Adding conda build recipe files

marcinz:
Files copied from the utils repository.


conda/conda-build/meta.yaml
```diff
@@ -0,0 +1,63 @@
+{% set name = "legate-core" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ environ.get('GIT_DESCRIBE_TAG', environ.get('GIT_FULL_HASH', '')) }}
```
m3vaz, line 5: Is there ever a situation where you wouldn't have a tag, but would have the hash?

conda/conda-build/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    The Legate project endeavors to democratize computing by 
+    making it possible for all programmers to leverage the power 
+    of large clusters of CPUs and GPUs by running the same code 
+    that runs on a desktop or a laptop at scale.
+  doc_url: https://github.com/nv-legate/legate.core
+  dev_url: https://github.com/nv-legate/legate.core
+
+extra:
+  recipe-maintainers:
+    - mavaz
```
m3vaz, line 63: m3vaz

conda/conda-build/meta.yaml
m3vaz: PEP386 compliance for version numbers, stash hash in the build string?

m3vaz (changes requested):

conda/conda-build/meta.yaml
marcinz: Yeah, I changed this already. Will commit soon.

conda/conda-build/meta.yaml
```diff
@@ -0,0 +1,77 @@
+{% set name = "legate-core" %}
+{% if package_version is defined %}
+    {% set version = package_version %}
+{% elif 'dev' in environ.get('GIT_DESCRIBE_TAG', '') %}
+    {% set version = environ.get('GIT_DESCRIBE_TAG', '') ~ environ.get('GIT_DESCRIBE_NUMBER', '') %}
```
m3vaz, line 5: Can you add a comment here describing the expected format for the tag?

marcinz: Done. I think that we should have more detail on the workflow in a separate document, but here at least we can say what tags are expected.

conda/conda-build/meta.yaml
```diff
@@ -0,0 +1,80 @@
+{% set name = "legate-core" %}
+{% if package_version is defined %}
+    {% set version = package_version %}
+# The tag must be in one of the two formats (where N is a digit):
+# vNN.NN.NN
+# vNN.NN.NN.dev
+{% elif 'dev' in environ.get('GIT_DESCRIBE_TAG', '') %}
+    {% set version = environ.get('GIT_DESCRIBE_TAG', '') ~ environ.get('GIT_DESCRIBE_NUMBER', '') %}
```
m3vaz, line 8: Remember to strip the preceding "v" from the tag so that the version number is valid

conda/conda-build/meta.yaml
```diff
@@ -0,0 +1,80 @@
+{% set name = "legate-core" %}
+{% if package_version is defined %}
+    {% set version = package_version %}
+# The tag must be in one of the two formats (where N is a digit):
+# vNN.NN.NN
+# vNN.NN.NN.dev
+{% elif 'dev' in environ.get('GIT_DESCRIBE_TAG', '') %}
+    {% set version = (environ.get('GIT_DESCRIBE_TAG', '') ~ environ.get('GIT_DESCRIBE_NUMBER', '')).lstrip('v') %}
+{% else %}
+    {% set version = environ.get('GIT_DESCRIBE_TAG', '').lstrip('v') %}
```
m3vaz, line 10: Either this conditional should have a default statement to handle if it's not running under a git repo, or the environ.get() in the fallback should have a default value.

marcinz: Hmm, I think that if we are running in one of these situations, we should provide package version through `package_version`. I don't think that there is really a good default. Let me know what do you think.

m3vaz: hrm, agreed. We could just set it to 0.0.0?

marcinz: We could do that, but I would personally prefer an error message that would remind me to set `package_version`. WDYT? Unfortunately, for cunumeric we don't have a choice since we need a default to avoid errors in the initial pass where the environment vars are not defined yet. That's why we need a "fake" default there, but it also can prevent error messages.

marcinz: I tried to remove the defaults, but that does not work:

```
TypeError: argument of type 'NoneType' is not iterable
```

So it seems that without the default, the output of `environ.get` ends up being `None` in the first pass. So we do have to keep the default, even if it is simply `''`. I will elevate the default into a separate variable.

m3vaz: Sounds good. Other than that LGTM.

marcinz: Should be fixed. Take a look and confirm if ready for merging.

PR #195, opened 2022-04-12
Bryanv/arg protocol

### LINKED ISSUE #192: Protocol/ABC for launcher types
refs: https://github.com/nv-legate/legate.core/pull/189#discussion_r846573918, https://github.com/nv-legate/legate.core/pull/189#discussion_r846573957, https://github.com/nv-legate/legate.core/pull/189#discussion_r846573970

Clean up this union type:
```
 Sequence[Union[ScalarArg, RegionFieldArg, FutureStoreArg]]
```

bryevdv:
Fixes #192 

This PR adds a `Protocol` for launcher "arg" types in order to remove complicated sum types (https://github.com/nv-legate/legate.core/pull/195/commits/1a2360d8d75c37e92e10e2eb44f82b1c6939d6fa)

While addressing this, some some other unrelated typing issues were uncovered. Quick fixes for these are included as well, in https://github.com/nv-legate/legate.core/commit/3cffb02794cc86e58000d341dfe30b534ea1ce14


legate/core/launcher.py
```diff
@@ -916,7 +912,7 @@ def build_single_task(self, argbuf: BufferBuilder) -> SingleTask:
 
     def execute(
         self, launch_domain: Rect, redop: Optional[int] = None
-    ) -> Union[Future, FutureMap]:
+    ) -> FutureMap:
```
magnatelee, line 915: If we want to do this, then we should also remove `redop` and the if branch in the body to make sure there's no code that can ever produce `Future`.

magnatelee:
left one comment. looks good to me otherwise.

legate/core/launcher.py
bryevdv: Weird, my comment here got lost. I had asked if this needed to be split, or if there is another solution. The reason for the change is that `Reduce` does this:
```
            weights = launcher.execute(launch_domain)
            ...
            weighted = Weighted(self._runtime, launch_shape, weights)
```
but `Weighted` does expect a `Future`
> legate/core/operation.py:787:62: error: Argument 3 to "Weighted" has incompatible type "Union[Future, FutureMap]"; expected "FutureMap"  [arg-type]

bryevdv: I think there might not actually be anywhere that passed `redop` though, in which case just removing it will be simple

bryevdv: https://github.com/nv-legate/legate.core/pull/195/commits/7e9c5d3c73c79c9e92e74296a25e65a766aa2216

bryevdv:
I think we resolved the one comment together, so I will go ahead and merge this

PR #196, opened 2022-04-14
More typing

bryevdv:
This PR adds type annotations to 
* `legate.core`
* `legate.core.communicator`
* `legate.core.io`

After this PR all that remain is `runtime.py`, `solver.py` and `store.py`


legate/core/io.py
```diff
@@ -185,41 +199,42 @@ def ingest(
...
+        legion.legion_sharding_functor_invert(
+            sid,
+            shard,
+            domain,
+            domain,
+            total_shards,
+            points_ptr,
+            points_size,
+        )
+        points = []
```
magnatelee, line 229: maybe it's outside the scope of this PR, but I'd prefer a for-comprehension here than an explicit loop calling `append`.

legate/core/legate.py
```diff
@@ -80,7 +80,7 @@ def stores(self) -> list[Store]:
     @staticmethod
     def from_stores(
         dtype: pyarrow.DataType,
-        stores: list[Store],
+        stores: list[Optional[Store]],
```
magnatelee, line 83: It's actually a bit surprising that you figured this out just by making mypy happy. was there any call site that required this type (which btw is correct)?

legate/core/store.py
```diff
@@ -1197,7 +1197,7 @@ def compute_key_partition(self, restrictions):
                 partition = Tiling(self._runtime, tile_shape, launch_shape)
             return partition
 
-    def compute_projection(self, proj_fn=None, launch_ndim=None):
+    def compute_projection(self, proj_fn=None, launch_ndim=None) -> Any:
```
magnatelee, line 1200: I'm pretty sure this returns integers.

legate/core/store.py
```diff
@@ -1225,7 +1225,9 @@ def compute_projection(self, proj_fn=None, launch_ndim=None):
     def find_restrictions(self):
         return self._transform.convert_restrictions(self._storage.restrictions)
 
-    def find_or_create_legion_partition(self, partition, complete=False):
+    def find_or_create_legion_partition(
+        self, partition, complete=False
+    ) -> Any:
```
magnatelee, line 1230: if we could use `legion.legion_logical_partition_t` here, that's the right type.

legate/core/legate.py
bryevdv: Line 143 of `io.py` does this
```
  array = Array(dtype, [None, store])
```

legate/core/store.py
bryevdv: Not presently. That will be better to sort out once any cffi vs cython dust settles.

legate/core/io.py
bryevdv: I'm generally a fan of going small-scale code gardening as it is noticed https://github.com/nv-legate/legate.core/pull/196/commits/83b6367b1feadcd79dc2d163a3dc22e8dcd82322

legate/core/store.py
bryevdv: https://github.com/nv-legate/legate.core/pull/196/commits/6a5897c32657ab74b13d989d99e01b2a577a5107

magnatelee (approved):

PR #197, opened 2022-04-14
Split ManualTask from other tasks

### LINKED ISSUE #193: Clean up incompat overrides for ManualTask
ref: https://github.com/nv-legate/legate.core/pull/189#discussion_r846574602

Several properties on `ManualTask` override base class property types in an unhappy way, requiring a `# type: ignore [assignment] ` for now.

bryevdv:
fixes #193 

This PR splits off a `PartSymTask` class (placeholder name) for tasks that handle 
```
        self._input_parts: list[PartSym] = []
        self._output_parts: list[PartSym] = []
        self._reduction_parts: list[PartSym] = []
```
So that `ManualTask` (which does not have/use these) can deviate in a clean way


legate/core/operation.py
```diff
@@ -337,7 +301,55 @@ def _add_communicators(
             launcher.add_communicator(handle)
 
 
-class AutoTask(Task):
+class PartSymTask(Task):
```
bryevdv, line 304: I think this is most most minimal correct decomposition for the existing leaf classes, but this name is not really semantically meaningful, so any suggestions are appreciated. Another alternative would be to collapse existing `Task` down into `Operation` so that this could be named just `Task`

magnatelee: I think this better be `AutoOperation` that directly inherits `Operation` and the `Task` should be a mix-in. Note that most of the methods in `Task` don't make sense for `Copy` and `Reduce` so they shouldn't be inheriting `Task`.

magnatelee: I think this better be `AutoOperation` that directly inherits `Operation` and the `Task` should be a mix-in. Note that most of the methods in `Task` don't make sense for `Copy` and `Reduce` so they shouldn't be inheriting `Task`.

bryevdv: mixins usually cause a bit of friction with `mypy` so we may require some explicit `ignore` here and there, but I will see what I can do

bryevdv: @magnatelee are you sure about `Reduce` ? It currently accepts a `task_id` in its `__init__`

magnatelee: `Reduce` certainly uses `task_id`, but I don't think it requires other properties and methods `Task` provides. I think it only depends on those for auto-partitioning that `AutoOperation` will provide.

bryevdv:
@magnatelee See if this work. Notes:

* Replaced ABC with Protocol because that is more ergonomic for mixins
* Reduce explicitly handles `self._task_id` to avoid being a `Task`

magnatelee (approved):

magnatelee:
LGTM. Make sure all cuNumeric tests pass with this change.

bryevdv:
@magnatelee :+1:  I always run tests locally before pushing a PR :)

PR #208, opened 2022-04-21
Simple stream pool for all Legate libraries

magnatelee:


src/core/cuda/stream_pool.h
```diff
@@ -0,0 +1,65 @@
...
+  StreamView(cudaStream_t stream) : valid_(true), stream_(stream) {}
+  ~StreamView();
+
+ public:
+  StreamView(const StreamView&)            = delete;
+  StreamView& operator=(const StreamView&) = delete;
+
+ public:
+  StreamView(StreamView&&);
+  StreamView& operator=(StreamView&&);
```
manopapad, line 38: How do you expect we'd use move semantics on a StreamView?

src/core/cuda/stream_pool.cu
```diff
@@ -0,0 +1,72 @@
...
+
+StreamView::~StreamView()
+{
+  if (valid_ && Core::synchronize_stream_view) {
+#ifdef DEBUG_LEGATE
+    CHECK_CUDA_STREAM(stream_);
+#else
+    CHECK_CUDA(cudaStreamSynchronize(stream_));
+#endif
+  }
```
manopapad, line 32: Can you explain the intent behind `synchronize_stream_view` mode? Based on how `StreamView` is used in cuNumeric, it looks like we will be creating one view at the beginning of the task and automatically destroying it at exit. With this assumption it looks like `synchronize_stream_view` mode is meant to ensure that the stream is automatically sync'd at exit, but we already get this through Realm's automatic context sync. The debug case does more, by also checking for runtime errors through `cudaPeekAtLastError`, but in that case why not do that always when debug mode is active, instead of also requiring `synchronize_stream_view`?

All in all, why not this instead?
```
 StreamView::~StreamView()
 {
 #ifdef DEBUG_LEGATE
   if (valid_) {
     CHECK_CUDA_STREAM(stream_);
   }
 #endif
}
```

lightsighter:
This is only until a certain "green feature" comes along right?

src/Makefile
```diff
@@ -44,6 +44,14 @@ NVCC_FLAGS ?=
 
 LD_FLAGS ?=
 
+ifeq ($(strip $(DEBUG)),1)
+
+CC_FLAGS += -DDEBUG_LEGATE
+
+NVCCC_FLAGS += -DDEBUG_LEGATE
```
manopapad, line 51: Typo: `NVCC_FLAGS`

src/Makefile
```diff
@@ -44,6 +44,14 @@ NVCC_FLAGS ?=
 
 LD_FLAGS ?=
 
+ifeq ($(strip $(DEBUG)),1)
```
manopapad, line 47: Not an issue in this PR, but I realized we need to be careful that we don't read `DEBUG_LEGATE` in header files, as those headers may be used in the compilation of another downstream library like cunumeric, where `DEBUG_LEGATE` is not defined. A more robust solution would be to record the status of such variables (that are controlled by compile-time flags like `DEBUG`) in a generated `legate_defines.h` header, like legion and realm do with `legion/realm_defines.h` -- also see #8.

src/core/cuda/stream_pool.h
magnatelee: People shouldn't move stream views unless they know what they are doing. The intent is to make sure the stream is synchronized only once within the scope.

src/core/cuda/stream_pool.cu
magnatelee: This PR is actually a prerequisite for my follow-up PR on the NVTX integration. The goal is to have NVTX ranges capture the entire kernel execution and that needs a stream synchronize before each NVTX range is closed. So the RAII on streams would do  that work automatically, as long as the gpu tasks are using stream views "correctly". Alternatively, we could do this in Realm; basically we can close the outstanding NVTX range in the context synchronizer. I don't want to take that route yet, so this is what I came up with.

Note that we don't necessarily want to always synchronize at the end of a scope, and we normally use NVTX ranges for profiling in release mode. So I think we do need `synchornize_stream_view`. A side benefit of being able to toggle the behavior from the outside is that we can easily isolate the cost of host-side execution.

Technically, I can call device synchronize instead before closing each range. I actually did that in the first attempt, but there were some NVTX ranges that report much longer time than their corresponding kernel execution time, and this made me suspect there's some driver-level serialization when multiple threads are calling device synchronize at the same time. (That said, calling stream synchronize concurrently from multiple threads isn't free of such serialization...)

magnatelee:
> This is only until a certain "green feature" comes along right?

Like I said, this is to make NVTX ranges correctly represent execution time for GPU tasks. It is orthogonal to what we do with the feature that you're mentioning.

src/Makefile
manopapad: Correction, this *is* in an issue in this PR. The `CHECK_CUDA_STREAM` macro was moved here from cunumeric, and cunumeric's `cuda_help.h` includes legate.core's `cuda_help.h`. But when cunumeric is being compiled with `--debug` it doesn't define `DEBUG_LEGATE` but `DEBUG_CUNUMERIC`, meaning the instances of `CHECK_CUDA_STREAM` in cunumeric tasks won't do anything.

magnatelee: You're right, and I think that issue is well within this PR, as I moved `CHECK_CUDA` and `CHECK_CUDA_STREAM` to the core, and they change their behavior based on the presence of `DEBUG_LEGATE`. Let me think about a solution.

magnatelee: Ok. The logic that adds`DEBUG_LEGATE` is moved to `legate.mk` that is shared by all Legate libraries: https://github.com/nv-legate/legate.core/pull/208/files#diff-12c8b10d84c75f3c0d91d58a3c3b8f55caea9427b83f34723ac9c60f8245042bR89 I believe this fixes the issue. @manopapad What do you think?

src/legate.mk
```diff
@@ -86,6 +86,14 @@ endif
 NVCC_FLAGS += -g
 endif
 
+ifeq ($(strip $(DEBUG)),1)
```
manopapad, line 89: Note that cunumeric sets `DEBUG` (and thus `DEBUG_LEGATE`) independently of legate.core. We will have to be careful not to have `DEBUG_LEGATE` control things like struct layout.

src/Makefile
manopapad: This works, but note that cunumeric sets `DEBUG` (and thus `DEBUG_LEGATE`) independently of legate.core. We will have to be careful not to have `DEBUG_LEGATE` control things like struct layout.

manopapad (approved):

PR #209, opened 2022-04-21
NVTX integration

magnatelee:
This PR makes an initial effort to integrate NVTX in Legate core. This creates an NVTX ranges for each task invocation, and with #208, the range can optionally represent the task's total execution time.


magnatelee:
This PR has some overlapping changes with #208, which will disappear once #208 is merged.

PR #214, opened 2022-04-29
Implement collective communicator for multi-node CPUs

eddy16112:
Provide CPU-only collective routines including alltoall, alltoallv and allgather for the case when there are multiple ranks per node. 

We adopt the similar API design from NCCL and MPI. Here are the description of APIs:
- `collInit` and `collFinalize` : similar to MPI_Init and MPI_Finalize, which are called by the top level task of each node. 
- `collGetUniqueId`: similar to `ncclGetUniqueId`, only rank 0 need to call it to get a unique id. The unique id will be used to distinguish communicators. 
- `collCommCreate` and `collCommDestroy`: similar to `ncclCommInitRank` and `ncclCommDestroy`, only ranks participated in the communicator need to call them to initialize/finalize a communicator. There are thread-safe. 
- `collAllgather`, `collAlltoall` and `collAlltoallv`: collective routines called by each rank within a communicator. 

It supports both with and without `LEGATE_USE_GASNET`. When enabling `LEGATE_USE_GASNET`, we use multi-thread MPI with tagging to implement collective communications. When disabling `LEGATE_USE_GASNET`, we use a shared buffer to allow each rank(thread) within the same node to exchange data.


src/core.mk
```diff
@@ -17,6 +17,8 @@
 # General source files
 GEN_CPU_SRC	= core/legate_c.cc                 \
 							core/comm/comm.cc                \
+							core/comm/comm_cpu.cc            \
+							core/comm/coll.cc				 \
```
magnatelee, line 21: Please match the formatting

src/core.mk
```diff
@@ -47,6 +61,7 @@ INSTALL_HEADERS = legate.h                        \
 									legate_preamble.h               \
 									core/legate_c.h                 \
 									core/comm/communicator.h        \
+									core/comm/coll.h				\
```
magnatelee, line 64: another formatting issue

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+
+  int total_size = global_comm->global_comm_size;
+
+  int sendtype_extent = collGetDtypeSize(sendtype);
+  int recvtype_extent = collGetDtypeSize(recvtype);
+  assert(sendtype_extent == recvtype_extent);
+
+  int global_rank = global_comm->global_rank;
+
+  void* sendbuf_tmp = NULL;
```
magnatelee, line 51: let's use `nullptr` instead.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+
+  // MPI_IN_PLACE
+  if (sendbuf == recvbuf) {
+    sendbuf_tmp = (void*)malloc(sendtype_extent * sendcount);
```
magnatelee, line 55: why is the casting necessary here?

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+    assert(sendbuf_tmp != NULL);
+    memcpy(sendbuf_tmp, recvbuf, sendtype_extent * sendcount);
+  } else {
+    sendbuf_tmp = const_cast<void*>(sendbuf);
```
magnatelee, line 59: it'd be more tasteful if you didn't have to use `const_cast`.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  global_comm->comm->buffers[global_rank] = sendbuf_tmp;
+  __sync_synchronize();
+
+  int recvfrom_global_rank;
+  for (int i = 0; i < total_size; i++) {
+    recvfrom_global_rank = i;
```
magnatelee, line 67: if `recvfrom_global_rank` is always `i`, then why don't you just name the loop variable `recvfrom_global_rank`? I don't think this extra assignment really necessary.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+    while (global_comm->comm->buffers[recvfrom_global_rank] == NULL)
```
magnatelee, line 68: first of all, use `nullptr`. second, this `while` loop looks pretty scary. I assume either the expression is never null or some other thread is assigning the field for forward progress. Please leave some comments on why this is a safe thing to do.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+      ;
+    char* src = (char*)global_comm->comm->buffers[recvfrom_global_rank];
```
magnatelee, line 70: let's use c++ castings.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+    char* dst = (char*)recvbuf + (ptrdiff_t)recvfrom_global_rank * recvtype_extent * recvcount;
+#ifdef DEBUG_PRINT
```
magnatelee, line 72: I thought we agree to use the `DEBUG_LEGATE` macro.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,98 @@
...
+    log_coll.info("i: %d === global_rank %d, dtype %d, copy rank %d (%p) to rank %d (%p)",
```
magnatelee, line 73: using `info` for debugging messages is always a bad idea. please use `debug` or `spew`.

src/core/comm/allgather_thread_mpi.cc
```diff
@@ -0,0 +1,67 @@
...
+  int total_size = global_comm->global_comm_size;
+
+  MPI_Datatype mpi_sendtype = collDtypeToMPIDtype(sendtype);
+
+  MPI_Aint lb, sendtype_extent, recvtype_extent;
+  MPI_Type_get_extent(mpi_sendtype, &lb, &sendtype_extent);
+
+  int global_rank = global_comm->global_rank;
+
+  void* sendbuf_tmp = NULL;
```
magnatelee, line 45: use `nullptr`

src/core/comm/allgather_thread_mpi.cc
```diff
@@ -0,0 +1,67 @@
...
+
+  // MPI_IN_PLACE
+  if (sendbuf == recvbuf) {
+    sendbuf_tmp = (void*)malloc(sendtype_extent * sendcount);
```
magnatelee, line 49: I don't think you need a casting.

src/core/comm/allgather_thread_mpi.cc
```diff
@@ -0,0 +1,67 @@
...
+    assert(sendbuf_tmp != NULL);
```
magnatelee, line 50: use `nullptr`

src/core/comm/alltoall_thread_local.cc
```diff
@@ -0,0 +1,105 @@
...
+  assert(sendtype == recvtype);
+
+  int total_size = global_comm->global_comm_size;
+
+  int sendtype_extent = collGetDtypeSize(sendtype);
+  int recvtype_extent = collGetDtypeSize(recvtype);
+
+  int global_rank = global_comm->global_rank;
+
+  void* sendbuf_tmp = NULL;
```
magnatelee, line 52: I've already seen this pattern three times. wouldn't it be better to extract this into a function?

src/core/comm/allgather_thread_local.cc
eddy16112: There is no way to avoid it. I tried to set the `sendbuf_tmp `to `const void*`, but later, when I do pointer arithmetic and `free(sendbuf_tmp)`, I will need to do the `const_cast`.

src/core/comm/allgather_thread_local.cc
eddy16112: done

src/core/comm/allgather_thread_local.cc
eddy16112: fixed

src/core/comm/allgather_thread_local.cc
eddy16112: done

src/core/comm/allgather_thread_local.cc
eddy16112: done

src/core/comm/allgather_thread_local.cc
eddy16112: fixed.

src/core/comm/allgather_thread_mpi.cc
eddy16112: fixed

src/core/comm/allgather_thread_mpi.cc
eddy16112: fixed

src/core/comm/allgather_thread_mpi.cc
eddy16112: fixed

src/core/comm/allgather_thread_local.cc
eddy16112: fixed

src/core/comm/allgather_thread_local.cc
eddy16112: Yeah, it is scary. I put some comments on the code.
The idea behind it is that each thread will put its `sendbuf `to `buffers[thread_id]`, and the while loop is to check if other threads has write to the buffer.

src/core/comm/alltoall_thread_local.cc
eddy16112: I cleaned it up https://github.com/eddy16112/legate.core/blob/coll/src/core/comm/allgather_thread_local.cc#L40-L54
I just feel like it is not a good way to put this 10 lines of code into a function, because all these 5 parameters are used later. If I put them into a function, then we will need to give 7-8 inputs and 5 outputs.

src/core.mk
eddy16112: Fixed. Somehow the format is OK in the VS code, and I can only see the wrong format in vim.

src/core.mk
eddy16112: fixed.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+                  const int sendcounts[],
+                  const int sdispls[],
+                  CollDataType sendtype,
+                  void* recvbuf,
+                  const int recvcounts[],
+                  const int rdispls[],
+                  CollDataType recvtype,
+                  CollComm global_comm)
+{
+  log_coll.print("Alltoallv: global_rank %d, mpi_rank %d, unique_id %d, comm_size %d",
```
magnatelee, line 169: I don't think `print` is the right logging level for things like this. it should be `debug` instead.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+  log_coll.print("Gather: global_rank %d, mpi_rank %d, unique_id %d, comm_size %d",
+                 global_comm->global_rank,
+                 global_comm->mpi_rank,
+                 global_comm->unique_id,
+                 global_comm->global_comm_size);
+#if defined(LEGATE_USE_GASNET)
+  return collGatherMPI(
+    sendbuf, sendcount, sendtype, recvbuf, recvcount, recvtype, root, global_comm);
+#else
+  printf("Not implemented\n");
```
magnatelee, line 221: should be `fprintf(stderr, "Not implemented\n");`. I feel we shouldn't even add this to the API if we don't support it in all cases.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+{
+  current_unique_id = 0;
+#if defined(LEGATE_USE_GASNET)
+  int provided, res, init_flag = 0;
+  MPI_Initialized(&init_flag);
+  if (!init_flag) {
+    res = MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
+    assert(res == MPI_SUCCESS);
+  } else {
+    log_coll.print(
```
magnatelee, line 273: do we actually hit this case? I feel we should just abort in this case, as it clearly indicates that this code isn't used as intended.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+  *id = current_unique_id;
+  current_unique_id++;
+  assert(current_unique_id <= MAX_NB_COMMS);
+  return CollSuccess;
+}
+
+#if defined(LEGATE_USE_GASNET)
+MPI_Datatype collDtypeToMPIDtype(CollDataType dtype)
+{
+  if (dtype == CollDataType::CollInt8) {
```
magnatelee, line 324: use a switch statement instead.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+
+int collGenerateAlltoallTag(int rank1, int rank2, CollComm global_comm)
+{
+  // tag: seg idx + rank_idx + tag
+  // int send_tag = ((sendto_global_rank * 10000 + global_rank) * 10 + ALLTOALL_TAG) * 10 +
+  // global_comm->unique_id; // which dst seg it sends to (in dst rank) int recv_tag = ((global_rank
+  // * 10000 + recvfrom_global_rank) * 10 + ALLTOALL_TAG) * 10 + global_comm->unique_id; // idx of
+  // current seg we are receving (in src/my rank)
+#if 1
+  int tag = (rank1 * 10000 + rank2) * CollTag::MAX_TAG + CollTag::ALLTOALL_TAG;
```
magnatelee, line 356: two comments: 1) I suggest you use shift and bitwise operators for the encoding here and other places. 2) please give a meaningful name to the constant `10000`.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+#else
+  // still under testing
```
magnatelee, line 358: please remove this

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,92 @@
...
+#include "legion.h"
+
+namespace legate {
+namespace comm {
+namespace coll {
+
+using namespace Legion;
+extern Logger log_coll;
+
+int collAllgatherLocal(const void* sendbuf,
```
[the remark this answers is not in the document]
eddy16112: thanks, I have already simplified all the internal function names.

src/core/comm/allgather_thread_local.cc
```diff
@@ -0,0 +1,92 @@
...
+                       int sendcount,
+                       CollDataType sendtype,
+                       void* recvbuf,
+                       int recvcount,
```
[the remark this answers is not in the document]
eddy16112: I think you are right, we do not need to make the API the same as MPI. I have fixed it.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+static std::vector<MPI_Comm> mpi_comms;
+#else  // undef LEGATE_USE_GASNET
+static std::vector<ThreadComm> thread_comms;
+#endif
+
+static std::atomic<int> current_unique_id(0);
+
+static bool coll_inited = false;
+
+static constexpr int const MAX_NB_COMMS = 100;
```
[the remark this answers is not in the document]
eddy16112: It will throw an error telling increasing the `MAX_NB_COMMS`. I am thinking about using `getenv` to make it user configurable, so it will be easier for users to increase the size. I explain the reason why we need this restriction in the next comment.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+
+enum CollTag : int {
+  BCAST_TAG     = 0,
+  GATHER_TAG    = 1,
+  ALLTOALL_TAG  = 2,
+  ALLTOALLV_TAG = 3,
+  MAX_TAG       = 10,
+};
+
+static std::vector<MPI_Comm> mpi_comms;
```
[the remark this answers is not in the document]
eddy16112: This is a list of `MPI_Comm` managed by us, not the cpu comm (`CollComm`) exposed to users. When using the `ncclCommInitRank` to create a comm, only the ranks within the comm need to call it, however, MPI is much more complicated: 1. all the ranks (even the one not in the comm) have to call the MPI communicator creation function; 2. even though MPI is thread safe, only 1 thread per node can call the MPI comm creation function. So I asked the Open MPI folks at UTK how to implement something like the `ncclCommInitRank`, they suggested me to create a list of `MPI_Comm` (each one is a copy of MPI_COMM_WORLD) during the library initialization, later, when using `collCommCreate` to create a `CollComm`, we just need to get a `MPI_Comm` from the list. Later, when the `CollComm` is finalized, we can return the `MPI_Comm` back. This makes the life much easier, and the we can use the `collCommCreate` the same as the `ncclCommInitRank`

src/core/comm/coll.cc
eddy16112: Yes, we do. GASNet may init MPI, which is before the `collInit` is called.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+#include "legion.h"
+
+namespace legate {
+namespace comm {
+namespace coll {
+
+using namespace Legion;
+Logger log_coll("coll");
+
+#if defined(LEGATE_USE_GASNET)
```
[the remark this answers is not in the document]
eddy16112: done

src/core/comm/alltoall_thread_mpi.cc
```diff
@@ -0,0 +1,240 @@
...
+namespace legate {
+namespace comm {
+namespace coll {
+
+using namespace Legion;
+extern Logger log_coll;
+
+#define ALLTOALL_USE_SENDRECV
+
+static int collAlltoallMPIInplace(void* recvbuf,
```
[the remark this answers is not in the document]
eddy16112: It was my first version, now I have removed it.

src/core/comm/coll.cc
eddy16112: Fixed.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+{
+  int tag = rank * CollTag::MAX_TAG + CollTag::GATHER_TAG;
+  assert(tag < INT_MAX && tag > 0);
+  return tag;
+}
+
+#else  // undef LEGATE_USE_GASNET
+size_t collGetDtypeSize(CollDataType dtype)
+{
+  if (dtype == CollDataType::CollInt8 || dtype == CollDataType::CollChar) {
```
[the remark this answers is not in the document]
eddy16112: Thank you. I have fixed it.

src/core/comm/alltoall_thread_mpi.cc
```diff
@@ -0,0 +1,240 @@
...
+                       sendto_mpi_rank,
+                       send_tag,
+                       dst,
+                       recvcount,
+                       mpi_recvtype,
+                       recvfrom_mpi_rank,
+                       recv_tag,
+                       global_comm->comm,
+                       &status);
+    assert(res == MPI_SUCCESS);
```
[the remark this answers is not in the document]
eddy16112: Thanks, I have replaced assert with check mpi

src/core/comm/coll.cc
```diff
@@ -0,0 +1,451 @@
...
+  ALLTOALLV_TAG = 3,
+  MAX_TAG       = 10,
+};
+
+static std::vector<MPI_Comm> mpi_comms;
+#else  // undef LEGATE_USE_GASNET
+static std::vector<ThreadComm> thread_comms;
+#endif
+
+static std::atomic<int> current_unique_id(0);
```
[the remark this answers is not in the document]
eddy16112: So all the `ncclGetUniqueId` call will be serialized? even if we have task parallel in the future?

src/core/comm/coll.cc
eddy16112: I plan to leave it as print for now until we finalize the PR. I will mark it.

src/core/comm/coll.cc
eddy16112: partial fixed.

src/core/comm/coll.cc
eddy16112: cleaned

src/core/comm/coll.cc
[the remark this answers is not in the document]
magnatelee: as long as we map all unique id generation tasks to the same processor, they will be atomically executed thanks to the programming model. There's a question of whether their order can be swapped, but for now we use execution fences that prevent that from happening.

src/core/comm/coll.cc
magnatelee: Let's fix it now and remove the WIP prefix from the title. I think this PR is almost ready for the merge.

src/core/comm/coll.cc
magnatelee: Your use of "may" confuses me. Does GASNet always initialize MPI or can it sometimes not initialize MPI before this function is called? If it's the former, we don't need this branch. Otherwise, I think we should understand when that can happen, and eliminate the uncertainty if possible.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+  coll_inited = true;
+  return CollSuccess;
+}
+
+int collFinalize(void)
+{
+  assert(coll_inited == true);
+  coll_inited = false;
+#ifdef LEGATE_USE_GASNET
+  for (int i = 0; i < MAX_NB_COMMS; i++) { CHECK_MPI(MPI_Comm_free(&mpi_comms[i])); }
```
magnatelee, line 248: So this is a step that I think should be performed normally by the finalize task of this communicator. I don't think it's a good design to defer all communicator destructions to the end of the program execution.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+  thread_comms.clear();
+  return CollSuccess;
+#endif
+}
+
+int collGetUniqueId(int* id)
+{
+  *id = current_unique_id;
+  current_unique_id++;
+  assert(current_unique_id <= MAX_NB_COMMS);
```
magnatelee, line 262: If you insist on setting the max number of communicators, two things should happen: 1) the value should be configurable by a compiler flag. 2) this code should raise a clear error message instead of asserting false.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+  // example:
+  // 00 | 01 | 02 | 03
+  // 10 | 11 | 12 | 13
+  // 20 | 21 | 22 | 23
+  // 30 | 31 | 32 | 33
+  // 01's send_tag = 10, 10's recv_tag = 10, match
+  // 12's send_tag = 21, 21's recv_tag = 21, match
+
+  int tag;
+  constexpr int const max_ranks = 10000;
```
magnatelee, line 301: Is there a reason to make the max number of ranks this small? I don't think this is a conservative bound.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+  tag                           = rank1 * max_ranks + rank2;
```
magnatelee, line 302: I remember I left a comment about the encoding scheme. Any comment on that? I guess this integer arithmetic is fine, but I feel bit operators would be more natural and also more efficient.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+
+  // still under testing, will be used once if runs out of tags
```
magnatelee, line 304: What testing are you still doing? If you're done with it, please remove these comments.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,389 @@
...
+  return tag;
+}
+
+#else  // undef LEGATE_USE_GASNET
+size_t getDtypeSize(CollDataType dtype)
+{
+  size_t size = 0;
+  switch (dtype) {
+    case CollDataType::CollInt8:
+    case CollDataType::CollChar: size = sizeof(char); break;
```
magnatelee, line 352: You can directly return the value, instead of going through this assignment; i.e.,
```
    case CollDataType::CollChar: return sizeof(char);
```
Also, please use braces for the body of each case.

src/core/comm/alltoall_thread_mpi.cc
```diff
@@ -0,0 +1,240 @@
...
+
+  MPI_Aint lb, sendtype_extent, recvtype_extent;
+  MPI_Type_get_extent(mpi_sendtype, &lb, &sendtype_extent);
+  MPI_Type_get_extent(mpi_recvtype, &lb, &recvtype_extent);
+
+  void* sendbuf_tmp = const_cast<void*>(sendbuf);
+
+  // MPI_IN_PLACE
+  if (sendbuf == recvbuf) {
+    // Not sure which way is better
```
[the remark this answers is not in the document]
eddy16112: By the way, do we need inplace collective? If not, I can completely remove all inplace code.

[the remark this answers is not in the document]
magnatelee: I guess we would do inplace allgathers, but I don't think we'd use an inplace all-to-all.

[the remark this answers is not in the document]
eddy16112: done

src/core/comm/coll.cc
eddy16112: fixed

src/core/comm/coll.cc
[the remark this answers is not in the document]
eddy16112: fixed.

src/core/comm/coll.cc
eddy16112: fixed.

src/core/comm/coll.cc
eddy16112: Reply together with the previous 3 comments.
when max_ranks=1000, the max tag calculated by the equation is 1,000,100,000, while the max int32 is 2,147,483,647. If I set the max_ranks to 20000, the tag could be overflow. 
This is kind of like to have a hash function for pair<rank1, rank2>, I have tested several hash functions below, but not sure if they will have problems. So now I am just using the easiest one and if ranks > 10000, we can figure it out later.

src/core/comm/coll.cc
eddy16112: Yes, I did not use a compiler flag, instead, people can do `export LEGATE_MAX_COMMS=x` to set the value without re-compiling the code, and I have already update the error message to tell people how to set it.

src/core/comm/coll.cc
eddy16112: Actually, I am not sure about it. I have been used ibv conduit, and GASNet indeed initializes MPI. So I add this warning just to let people know if they see errors, it may be because of the MPI init.

src/core/comm/coll.cc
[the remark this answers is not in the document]
eddy16112: Yeah, but we can not do it using a single index launcher just like nccl, so for now, to make it easy, I just create several MPI_Comm during init and put them in the vector. I just do not want put too much time on it because after this release, I plan to sync with Mike's student to see if he has made any progress in the realm level collective, and I also would like to take a look at the ucx to see if we need to implement the collective without MPI. MPI is not rank-per-core friendly, and has limited number of tags which limits the max number of ranks we can run.

src/core/comm/coll.cc
eddy16112: The user-level `CollComm` is destroyed by the `collCommDestroy` call in the finalize task, but not the `MPI_Comm`. We can reuse the `MPI_Comm` for later created `CollComm`.

src/core/comm/coll.cc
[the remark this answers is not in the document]
magnatelee: I don't believe you can't do it with a single index launch, but if the plan is to switch over to UCX eventually, I agree that we should spend too much time on it.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,443 @@
...
+  global_comm->global_rank      = global_rank;
+  global_comm->status           = true;
+  global_comm->unique_id        = unique_id;
+#ifdef LEGATE_USE_GASNET
+  int mpi_rank, mpi_comm_size;
+  int *tag_ub, flag;
+  int compare_result;
+  MPI_Comm comm = mpi_comms[unique_id];
+  CHECK_MPI(MPI_Comm_compare(comm, MPI_COMM_WORLD, &compare_result));
+  assert(compare_result = MPI_CONGRUENT);
```
magnatelee, line 78: needs an equality check instead of an assignment. fyi, it's a good coding practice to put constant values on the LHS of equality checks to avoid mistakes like this.

src/core/comm/coll.cc
magnatelee: Is there a way to check if the MPI was initialized with `MPI_THREAD_MULTIPLE`? This warning message doesn't seem very useful, as the users have no way to fix it. Users will almost certainly use the install script that we provide, so if GASNet didn't initialize MPI with `MPI_THREAD_MULTIPLE`, it's actually our fault. And if we make sure GASNet always initializes MPI correctly, this warning message is unnecessary. I guess this is yet another issue that we won't have with UCX, right?

src/core/comm/coll.cc
magnatelee: Sure, but do we actually reuse the existing `MPI_Comm`? I don't think I see the code that does that.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,443 @@
...
+    }
+    case CollDataType::CollFloat: {
+      return sizeof(float);
+    }
+    case CollDataType::CollDouble: {
+      return sizeof(double);
+    }
+    default: {
+      log_coll.fatal("Unknown datatype");
+      assert(0);
```
magnatelee, line 413: This assertion is also redundant.

src/core/comm/coll.cc
magnatelee: Do tags need to be unique for all pairs of logical ranks? I thought messages sent to rank k can use the same set of tags as those sent to rank k + 1, as their destination ranks would distinguish between those two sets of messages.

src/core/comm/coll.cc
eddy16112: nice catch! I will update the assert to put constant value on the left. Thank you

src/core/comm/coll.cc
eddy16112: fixed

src/core/comm/coll.cc
eddy16112: Yes, because rank k and rank k+1 could be in the same node (MPI rank), actually what is why we need to use tag, while the original MPI implementation of collectives does not rely on tag. 
However, it is allowed sending to rank i and rank j to use the same pair as long as they are not in the same node, so I implemented a new way to compute the tag under the "still under testing"
`tag = rank1 % max_threads * max_ranks + rank2; `
I have also found 2 hash functions, and they are also under "still under testing", but they also have limitations.

magnatelee: By "rank" I meant the MPI rank. The crux of the problem is that we want to distinguish between messages from MPI rank `i` on MPI rank `j`, where the messages are sent by different logical ranks on MPI rank `i`. What I was saying is that the tags we pass don't need to be globally unique but just need to be different for different threads on MPI rank `i`. Isn't this true? Or does the MPI require that the tags be globally unique for some reason I don't understand? Last time I checked with a toy example the former was the case. An MPI rank is already associated with every MPI send or receive request, so it feels redundant that the tag encodes it somehow.

magnatelee: There's a possibility that the tags should be unique for all-to-all messages within the same MPI rank, but we can treat them specially.

src/core/comm/coll.cc
eddy16112: Actually I have the reuse code in another branch
https://github.com/eddy16112/legate.core/blob/coll-pool/src/core/comm/coll.cc#L278
Since I realized legate won't delete any communicators until the finalization of the library, so it is necessary to have the reuse code, which can increase the complexity.

magnatelee: Legate currently doesn't destroy any communicator during execution, but that can change in the future. And Legate's communicator manager is already pooling the communicators, so I don't think individual communicators should roll their own pooling logic like this. I'd rather have the `MPI_Comm` destroyed when the wrapping `CollComm` is destroyed.

src/core/comm/coll.cc
magnatelee: Ok, so that sounds like a real problem, and once we fixed that, we no longer need to have this check, as MPI will always get initialized correctly by Legate. Let me figure out where to add the init call.

src/core/comm/coll.cc
eddy16112: I think we can postpone this problem together with the vector of `MPI_Comm` because they are actually the same one.

src/core/comm/coll.cc
magnatelee: Let's go with the new tagging scheme. I agree that globally unique tags are easy to work with if you are still working on the communication routines, but I believe you already passed that point, so we better switch to a tagging scheme that would allow us to use practically infinitely many logical ranks.

eddy16112: I have update the tagging scheme
https://github.com/eddy16112/legate.core/blob/coll/src/core/comm/coll.cc#L336

magnatelee:
@eddy16112 what else do you need to fix for the merge? unless there's anything else you want to change, feel free to merge it

src/core/comm/coll.cc
eddy16112: I found the MPI function that can check if MPI_THREAD_MULTIPLE is used, and I have updated the code accordingly.

eddy16112:
> @eddy16112 what else do you need to fix for the merge? unless there's anything else you want to change, feel free to merge it

I am OK to merge it now, but it won't work until we add MPI_Init to legion_python.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,443 @@
...
+#endif
+}
+
+int collGetUniqueId(int* id)
+{
+  *id = current_unique_id;
+  current_unique_id++;
+  if (current_unique_id > MAX_NB_COMMS) {
+    log_coll.fatal(
+      "Please increase the LEGATE_MAX_CPU_COMMS by export LEGATE_MAX_CPU_COMMS=new number, current "
```
magnatelee, line 272: Chage `export LEGATE_MAX_CPU_COMMS=new number` to `\"export LEGATE_MAX_CPU_COMMS=(new number)\"`.

src/core/comm/coll.cc
```diff
@@ -0,0 +1,443 @@
...
+      "value "
```
magnatelee, line 273: Please fix this weird line formatting.

src/core/comm/coll.cc
eddy16112: fixed.

src/core/comm/coll.cc
eddy16112: I have fixed it, sometime the pre-commit can change the code into a weird format.

magnatelee:
> I am OK to merge it now, but it won't work until we add MPI_Init to legion_python.

#264 should address the problem.

legate/core/communicator.py
```diff
@@ -104,3 +104,51 @@ def _finalize(self, volume: int, handle: FutureMap) -> None:
...
+            library.LEGATE_CORE_INIT_CPUCOLL_MAPPING_TASK_ID
+        )
+        self._init_cpucoll = library.LEGATE_CORE_INIT_CPUCOLL_TASK_ID
+        self._finalize_cpucoll = library.LEGATE_CORE_FINALIZE_CPUCOLL_TASK_ID
+        self._tag = library.LEGATE_CPU_VARIANT
+        self._needs_barrier = False
+
+    def destroy(self) -> None:
+        Communicator.destroy(self)
+        self._runtime.issue_execution_fence(block=True)
```
magnatelee, line 125: Two comments: 1) please leave a comment why we need this blocking. 2) do this only if you have any CPU communicators. many programs don't ever create them.

eddy16112: This is the destructor of communicator, so if programs do not create them, the destroy function will not be called, right?

magnatelee: `Communicator.destroy` is invoked unconditionally once per communicator type. it's `Communicator._finalize` that is invoked for each created communicator. the fence is to block on the tasks launched by those `Communicator._finalize` calls, so you don't need it if no communicator is created. I assume you still need to `legate_cpucoll_finalize` even when no CPU communicator is created. but I believe that function can be invoked without the fence in that case.

eddy16112: Got it, I have pushed a fix for it. 
https://github.com/nv-legate/legate.core/pull/214/commits/d319db6326ce89f4c604c0f2236af44b3cbce6c0

magnatelee (approved):

magnatelee:
LGTM. Feel free to merge it when you can

eddy16112:
> LGTM. Feel free to merge it when you can

Thank you, I will update the description, and then merge it.

PR #223, opened 2022-05-09
Update build.sh with specific CPU targeting

m3vaz:
Updates build.sh with:
1) Set the std to c++17 to fix the type_traits templating issue
2) Builds up the install arguments in a variable and comments on each addition
3) Adds an install argument to target the CPU feature set equivalent to haswell


m3vaz:
There is no corresponding change in cunumeric since cunumeric inherits its args from the legate installation it is compiled against.

marcinz (approved):

manopapad (approved):
LGTM otherwise

PR #231, opened 2022-05-10
Revert "Switch conda recipe to rely on conda openblas"

m3vaz:
Reverts nv-legate/legate.core#230


manopapad:
Context: the `--with-openblas` option is supposed to be used on cunumeric, not legate.core. The reviewer (me) on the original issue https://github.com/nv-legate/legate.core/pull/230 should have caught it at code review time...

manopapad (approved):

PR #234, opened 2022-05-11
Launcher flags to configure Nsight Systems

magnatelee:
This PR introduces two new launcher flags for Nsight:

* `--nsys-targets`: takes a comma-separated list of APIs to trace with Nsight. The default is `cublas,cuda,cudnn,nvtx,ucx`.
* `--nsys-extra`: can be used to specify extra flags to pass to Nsight. More than one `--nsys-extra` can be used. For example, `--nsys-extra=-s --nsys-extra=cpu` enables CPU sampling.

### DIFF
--- a/legate.py
+++ b/legate.py
@@ -769,14 +769,14 @@ def driver():
         dest="nsys",
         action="store_true",
         required=False,
-        help="run Legate with Nsight",
+        help="run Legate with Nsight Systems",
     )
     parser.add_argument(
         "--nsys-targets",
         dest="nsys_targets",
         default="cublas,cuda,cudnn,nvtx,ucx",
         required=False,
-        help="Specify profiling targets for Nsight",
+        help="Specify profiling targets for Nsight Systems",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


legate.py
```diff
@@ -767,7 +769,22 @@ def driver():
         dest="nsys",
         action="store_true",
         required=False,
-        help="run Legate with nsys",
+        help="run Legate with Nsight",
```
manopapad, line 772: I would use "Nsight Systems", to disambiguate from "Nsight Compute".

magnatelee: ada8bf41c6b7f5d6778f77b95d6a8771c5e4cde1

manopapad:
@SeyedMir It has to be separate instances. If you give `--nsys-extra="-s cpu"` that will turn into the command line:
```
nsys "-s cpu" ...
```
I.e. we will pass the entire `"-s cpu"` as a single string, and nsys will complain about unrecognized option / non-existent file.

This is because we're using the "append" argparse action https://docs.python.org/3/library/argparse.html#action.

PR #236, opened 2022-05-12
Add script wrapper for zero code-change patching

bryevdv:
This PR adds a wrapper for legate that can patch specified libraries and then execute a given script with those patches already in place.
```
~/work/legate.core bryanv/zcc
dev38 ❯ lgpatch.py test.py -patch numpy -cunumeric:report:coverage
cuNumeric API coverage: 4/4 (100.0%)
```

See below for `test.py` example script that only imports numpy. 

### Questions

* Is `scripts/lgpatch.py` the best place/name for this? Should it be installed (as `lgpatch`?) in the install bin directory?
* Is `-patch:help` suffient for help? Legate intercepts `-h` and `--help` before we can get to them so we can't use those. 
* This does not use the trivial `cunumeric.patch()` because I wanted to have more error checking. But we could move that to `patch` and simplify here if desired (but then every new lib would be responsible for checking individually in their `patch()`) If we did this the usage would also switch to `-patch cunumeric` instead of the current `-patch numpy`.
* Alternatively, we could also just remove `cunumeric.patch` if is not needed after this. 

---

#### test.py
```
import numpy as np

def cholesky(n, dtype):
    input = np.eye(n, dtype=dtype)
    np.linalg.cholesky(input)

if __name__ == "__main__":
    cholesky(10, np.float32)
```

#### `-patch:help`
```
dev38 ❯ ./scripts/lgpatch.py -patch:help                                    
usage: lgpatch [-patch PATCH [PATCH ...]] PROG

Patch existing libraries with legate equivalents.

Currently the following patching can be applied:

    numpy -> cunumeric

positional arguments:
  PROG                  The legate program to run

optional arguments:
  -patch PATCH [PATCH ...]
                        Patch the specified libraries

Any additional command line arguments are passed on to PROG as-is
```

### DIFF
--- a/scripts/lgpatch.py
+++ b/scripts/lgpatch.py
@@ -1,0 +1,91 @@
+#! /usr/bin/env legate
+# Copyright 2021-2022 NVIDIA Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import sys
+import textwrap
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+
+KNOWN_PATCHES = {"numpy": "cunumeric"}
+
+newline = "\n"
+
+DESCRIPTION = textwrap.dedent(
+    f"""
+Patch existing libraries with legate equivalents.
+
+Currently the following patching can be applied:
+
+{newline.join(f'    {key} -> {value}' for key, value in KNOWN_PATCHES.items())}
+
+"""
+)
+
+EPILOG = """
+Any additional command line arguments are passed on to PROG as-is
+"""
+
+
+def parse_args():
+    parser = ArgumentParser(
+        prog="lgpatch",
+        description=DESCRIPTION,
+        allow_abbrev=False,
+        add_help=False,
+        epilog=EPILOG,
+        formatter_class=RawDescriptionHelpFormatter,
+    )
+    parser.add_argument(
+        "prog", metavar="PROG", help="The legate program to run"
+    )
+    parser.add_argument(
+        "-patch",
+        action="store",
+        nargs="+",
+        help="Patch the specified libraries",
+        default=[],
+    )
+
+    # legate parser intercepts "-h" and "--help" earlier
+    if "-patch:help" in sys.argv:
+        parser.print_help()
+        sys.exit()
+
+    return parser.parse_known_args()
+
+
+def do_patch(name: str) -> None:
+    if name not in KNOWN_PATCHES:
+        raise ValueError(f"No patch available for module {name}")
+    if name in sys.modules:
+        raise RuntimeError(f"Cannot patch {name} -- it is already loaded")
+
+    cuname = KNOWN_PATCHES[name]
+    try:
+        module = __import__(cuname)
+        sys.modules[name] = module
+    except ImportError:
+        raise RuntimeError(f"Could not import patch module {cuname}")
+
+
+if __name__ == "__main__":
+    args, extra = parse_args()
+
+    for name in args.patch:
+        do_patch(name)
+
+    sys.argv[:] = [args.prog] + extra
+
+    with open(args.prog) as f:
+        exec(f.read(), {"__name__": "__main__"})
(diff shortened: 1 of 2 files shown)


manopapad:
How does this handle the fact that we import numpy internally in cunumeric?

bryevdv:
> How does this handle the fact that we import numpy internally in cunumeric?

It doesn't but also doesn't need to AFAICT. It is replaced in `sys.modules` before any of the use code is ever executed. The module and ndarray wrapping/cloning also has already happened (using the real numpy) so if other parts of cunumeric use "numpy" they will end up with the wrapped/cloned objects (which already have internal references to real numpy set up as appropriate)

manopapad:
Got it, I missed the fact that `__import__(cunumeric)` gets executed in full before swapping it out for `numpy` in `sys.modules`. So as long as we load all `cunumeric` sub-libraries in `__init__.py`, and put all our internal `import numpy`s at global scope we should be fine.

This might still get dicey if we start distributing multiple libraries, and multiple of those need to internally import `numpy`. For example, say we have `cunumeric`, which replaces `numpy`, and `cufoo`, which replaces `foo`. And say that `cufoo` needs to import the real `numpy`. If you do:

```
lgpatch.py test.py -patch numpy -patch foo
```

Then AFAIU `cunumeric` will get loaded (with its internal pointers pointing properly at the actual `numpy`), then `cufoo` will get loaded, and when `cufoo` tries to `import numpy` it will instead get `cunumeric`.

At that point we could have the patch module maintain a pointer to all "real" instances of patched libraries, and we internally use that when doing such imports.

scripts/lgpatch.py
```diff
@@ -0,0 +1,91 @@
+#! /usr/bin/env legate
+# Copyright 2021-2022 NVIDIA Corporation
```
magnatelee, line 2: I guess the year could be just 2022?

magnatelee (approved):

bryevdv:
> This might still get dicey if we start distributing multiple libraries, and multiple of those need to internally import numpy.

Yes I do agree once we get to N=2 this will need to become a little more sophisticated.

scripts/lgpatch.py
bryevdv:
```suggestion
# Copyright 2022 NVIDIA Corporation
```

magnatelee:
> Is scripts/lgpatch.py the best place/name for this? Should it be installed (as lgpatch?) in the install bin directory?

I think the name is fine and it should be installed in the `bin` directory.

> Is -patch:help suffient for help? Legate intercepts -h and --help before we can get to them so we can't use those.

That's good for now. I think Legate intercepting that flag is a problem in other contexts too; for example, some micro-benchmarks have their own argument parsers, but `-h` gets intercepted and doesn't print the benchmarks' help. I think we should either pass `-h` to the executing program or use a different flag it.

> This does not use the trivial cunumeric.patch() because I wanted to have more error checking. But we could move that to patch and simplify here if desired (but then every new lib would be responsible for checking individually in their patch()) If we did this the usage would also switch to -patch cunumeric instead of the current -patch numpy.
> Alternatively, we could also just remove cunumeric.patch if is not needed after this.

I guess we could just remove `cunumeric.patch` once this PR gets merged.

> Yes I do agree once we get to N=2 this will need to become a little more sophisticated.

Can't we just import all listed modules first and then patch them?

bryevdv:
>  I think we should either pass -h to the executing program or use a different flag it.

Well the problem is that `legate` iteself has extensive command line arg help and expecting `legate -h` to provide that is not unreasonable for users. I'm not sure what the best compromise is. 

> Can't we just import all listed modules first and then patch them? 

Possibly. The concern is what @manopapad mentions:

> Then AFAIU cunumeric will get loaded (with its internal pointers pointing properly at the actual numpy), then cufoo will get loaded, and when cufoo tries to import numpy it will instead get cunumeric.

Which, to be honest, may be what is desired. But it depends on how the multiple downstream libs are designed to interact, whether there are any ordering issues for patching, etc.

manopapad:
> Well the problem is that legate iteself has extensive command line arg help and expecting legate -h to provide that is not unreasonable for users. I'm not sure what the best compromise is.

We could consider switching to a mode where any flag coming before the script name goes to legate/cunumeric, and any flag after the script name goes to the script:

```
bin/legate --nsys --profile examples/gemm.py --iters 20
```

Then there is no ambiguity. I think last time this was suggested it didn't go over well, but I forget why.

> Can't we just import all listed modules first and then patch them?

I think that would work most of the time. I can imagine it breaking if the user asks to patch `numpy` but not `foo`, then later imports `cufoo` directly, as that would still trigger the failure mode I wrote above. But that's a little far-fetched.

In any case, none of these discussions should stop this PR from getting merged.

bryevdv:
> Then there is no ambiguity. I think last time this was suggested it didn't go over well, but I forget why.

I think it's always a little surprising (as a user) but I would be :+1: on this personally. I do think it might be tricky getting `argparse` to play ball, though.

> In any case, none of these discussions should stop this PR from getting merged.

:+1: I just need to figure out how to get `lgpatch.py` installed as an executable in our setup. Once that is done this will be ready
=

PR #241, opened 2022-05-19
Use read-write privilege for scatter targets

magnatelee:
Legate core was using write-discard privilege for scatter targets, which make them not fetch valid values for pass-through entries. This PR is to fix that issue.

### DIFF
--- a/legate/core/_legion/operation.py
+++ b/legate/core/_legion/operation.py
@@ -450,6 +450,7 @@ class Copy(Dispatchable[None]):
         parent: Optional[Region] = None,
         tag: int = 0,
         redop: int = 0,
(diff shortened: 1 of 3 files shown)


manopapad:
I guess we don't need to update privileges for non-scatter copies, because the core can only handle full-region copies at the moment (no transforms)?

magnatelee:
> I guess we don't need to update privileges for non-scatter copies, because the core can only handle full-region copies at the moment (no transforms)?

Yes. Once we start passing transforms directly to Legion copies, some of the transforms make the copies effectively scatter even when no indirection field is given. I'll make a note on this.

manopapad (approved):

PR #246, opened 2022-05-25
Use phase barriers to work around the CUDA driver bug

magnatelee:
[nv-legate/cunumeric/#374](https://github.com/nv-legate/cunumeric/issues/374) turned out to be caused by a CUDA driver bug that has been fixed in CUDA driver >= 520. This PR is a workaround that introduces a barrier at the beginning of every NCCL task whenever the system is using a buggy driver.


install.py
```diff
@@ -421,6 +421,14 @@ def build_legion(
             + setup_py_flags,
             cwd=legion_python_dir,
         )
+    verbose_check_call(
```
manopapad, line 424: Should this be added to the `INSTALL_HEADERS` in https://github.com/StanfordLegion/legion/blob/master/runtime/runtime.mk#L1030?

magnatelee: I guess we should check with Mike but I don't think we actually want to expose that header, though it's necessary for the core @lightsighter any thoughts?

src/core/utilities/deserializer.h
```diff
@@ -89,6 +89,7 @@ class TaskDeserializer : public BaseDeserializer<TaskDeserializer> {
   void _unpack(RegionField& value);
   void _unpack(OutputRegionField& value);
   void _unpack(comm::Communicator& value);
+  void _unpack(Legion::PhaseBarrier& future);
```
manopapad, line 92: rename arg to `barrier` to match definition

legate/core/launcher.py
```diff
@@ -901,6 +906,11 @@ def build_task(
             req.proj.add(task, req, fields, _index_task_calls)
         for future in self._future_args:
             task.add_future(future)
+        if self._insert_barrier:
+            volume = launch_domain.get_volume()
+            arrival, wait = self._runtime.get_barriers(volume)
+            task.add_future(arrival)
+            task.add_future(wait)
```
manopapad, line 913: How come these are passed as futures and not scalars?

src/core/comm/comm_nccl.cu
```diff
@@ -170,6 +189,20 @@ void register_tasks(Legion::Machine machine,
   }
 }
 
+static const char* MIN_VERSION_WITH_HANG_FIX = "520";
+
+bool needs_barrier()
+{
+  CHECK_NVML(nvmlInit_v2());
```
manopapad, line 196: Could you add a couple of sentences here for future readers, describing the issue and workaround? It's clear by reading the code how the workaround functions, but it's not clear why it's needed.

manopapad (approved):
Some minor questions, otherwise looks good.

src/core/utilities/deserializer.h
magnatelee: 1cb2ff61bdc52cacad92ff1f2b227dc2daf85285

src/core/comm/comm_nccl.cu
magnatelee: 72cd10eb3abbfcdde9d6d0d8faab5289592e4764

legate/core/launcher.py
magnatelee: I haven't found a sane way to pack cdata objects in the argument buffer.

src/core/comm/comm_nccl.cu
```diff
@@ -170,6 +171,27 @@ void register_tasks(Legion::Machine machine,
   }
 }
 
+bool needs_barrier()
+{
+  int32_t ver;
+  auto status = cuDriverGetVersion(&ver);
+  if (status != CUDA_SUCCESS) {
+    const char* error_string;
+    cuGetErrorString(status, &error_string),
```
manopapad, line 180: Should be semicolon here instead of comma?

magnatelee: 12aef5f92ecfcf6be0eefd0191b2e0a9e80bafc0

PR #258, opened 2022-06-03
Exception support

magnatelee:
This PR adds exception support to Legate core. The goal is to allow Legate tasks in C++ to toss their exceptions to the Python land and have them re-thrown.


src/core/task/task.h
```diff
@@ -96,9 +98,26 @@ class LegateTask {
...
+    } catch (Realm::ApplicationException& e) {
+      throw e;
+    } catch (legate::TaskException& e) {
+      if (context.can_raise_exception()) {
+        context.make_all_unbound_stores_empty();
+        return context.pack_return_values_with_exception(e.index(), e.error_message());
+      } else
+        // This is an unexpected exception so better be rethrown
+        throw e;
+    }
```
manopapad, line 120: You may be able to use just `throw;` which reuses the same exception object, and can be used with abstract classes:

```
      catch (const legate::TaskException& e) {
        if (context.can_raise_exception()) {
          context.make_all_unbound_stores_empty();
          return context.pack_return_values_with_exception(e.index(), e.error_message());
        } else
          // This is an unexpected exception so better be rethrown
          throw;
      } catch (const std::exception& e) {
        throw;
      }
```

Come to think of it, do you need to handle all other exception classes here explicitly? Assuming the exception doesn't match `legate::TaskException`, won't it automatically be re-thrown to the parent function?

magnatelee: Initially I was trying to catch all non-Legate C++ exceptions except the Realm ones and re-throw them on the Python side with a generic Legate exception. Then I realized that that would effectively require the core to do exception handling for every task, which I don't think we want. So I gave up on that idea but forgot to remove the cases for Realm exceptions that are obsolete.

I updated the code so that it only handles Legate exceptions. In case they are raised by tasks that did not declare any Python exceptions, they are reported as errors now.

src/core/runtime/runtime.cc
```diff
@@ -118,6 +119,17 @@ static ReturnValues extract_scalar_task(const Task* task,
...
+/*static*/ void Core::report_unexpected_exception(const char* task_name,
+                                                  const legate::TaskException& e)
+{
+  log_legate.error(
+    "Task %s threw an exception \"%s\", but the task did not declare any exception. "
+    "Please specify a Python exception that you want this exception to be re-thrown with "
+    "using 'throws_exception'.",
+    task_name,
+    e.error_message().c_str());
+}
```
manopapad, line 131: Shouldn't we also `LEGATE_ABORT` at this point?

src/core/task/task.h
```diff
@@ -96,9 +98,19 @@ class LegateTask {
...
+      return context.pack_return_values();
+    } catch (legate::TaskException& e) {
+      if (context.can_raise_exception()) {
+        context.make_all_unbound_stores_empty();
+        return context.pack_return_values_with_exception(e.index(), e.error_message());
+      } else
+        // If a Legate exception is thrown by a task that does not declare any exception,
+        // this is a bug in the library that needs to be reported to the developer
+        Core::report_unexpected_exception(task_name(), e);
+    }
```
manopapad, line 113: clang warns here that
```
/Users/mpapadakis/legate.core/install/include/core/task/task.h:114:3: warning: non-void function does not return a value in all control paths [-Wreturn-type]
```

src/core/task/return.cc
```diff
@@ -30,6 +33,111 @@ using namespace Legion;
...
+struct JoinReturnedException {
+  using LHS = ReturnedException;
+  using RHS = LHS;
+
+  static const ReturnedException identity;
+
+  template <bool EXCLUSIVE>
+  static void apply(LHS& lhs, RHS rhs)
+  {
+    assert(EXCLUSIVE);
```
manopapad, line 45: Should this be guarded by `DEBUG_LEGATE`? Or possibly have two specializations, for `EXCLUSIVE=true` and `EXCLUSIVE=false`, where the latter just does `assert(false)`?

src/core/task/return.cc
```diff
@@ -87,4 +195,11 @@ void ReturnValues::legion_deserialize(const void* buffer)
   buffer_size_ = ptr - static_cast<const int8_t*>(buffer);
 }
 
+void register_exception_reduction_op(Runtime* runtime, const LibraryContext& context)
+{
+  auto redop_id = context.get_reduction_op_id(LEGATE_CORE_JOIN_EXCEPTION_OP);
+  auto* redop   = Realm::ReductionOpUntyped::create_reduction_op<JoinReturnedException>();
+  runtime->register_reduction_op(redop_id, redop, returned_exception_init, returned_exception_fold);
```
manopapad, line 202: Should this be `Runtime::register_reduction_op` ?

legate/core/runtime.py
```diff
@@ -883,6 +886,24 @@ def __init__(self, core_library: CoreLib) -> None:
...
+                legion.LEGATE_CORE_TUNABLE_MAX_PENDING_EXCEPTIONS,
+                ty.uint32,
+            )
+        )
+        self._precise_exception_trace: bool = (
+            self._max_pending_exceptions > 1
+            and bool(
+                self._core_context.get_tunable(
+                    legion.LEGATE_CORE_TUNABLE_PRECISE_EXCEPTION_TRACE,
+                    ty.uint32,
```
manopapad, line 900: not `ty.bool_` as with `LEGATE_CORE_TUNABLE_NCCL_NEEDS_BARRIER`?

src/core/task/task.h
magnatelee: a4463afbdcac4549d3d83618fadfea5f9f7535e7

src/core/task/return.cc
magnatelee: 7feb1761de1c7dc4c1fc2c73f682cb3682590794

legate/core/runtime.py
magnatelee: 47eeb80d968239bc44f4c9a65b2473b369b25f6c

src/core/runtime/runtime.cc
magnatelee: `Logger::error` kills the process, so no need to add `LEGATE_ABORT`.

src/core/task/return.cc
magnatelee: d91979381eafbeaa7bbd980b91c7db0251bf70a5

legate/core/runtime.py
```diff
@@ -1015,9 +1039,13 @@ def flush_scheduling_window(self) -> None:
         self._schedule(ops)
 
     def submit(self, op: Operation) -> None:
+        if op.can_raise_exception and self._precise_exception_trace:
```
manopapad, line 1042: IIUC, the only computation we save if we don't do precise exceptions is that we skip a stacktrace save when launching an operation that can throw.

Given that this saving is at the python control program only (i.e. not on the critical computation path), and only affects operations that can throw, do you think the savings is worth the inaccuracy?

legate/core/runtime.py
```diff
@@ -883,6 +886,24 @@ def __init__(self, core_library: CoreLib) -> None:
...
         ] = {}
 
+        self._max_pending_exceptions: int = int(
+            self._core_context.get_tunable(
+                legion.LEGATE_CORE_TUNABLE_MAX_PENDING_EXCEPTIONS,
+                ty.uint32,
+            )
+        )
+        self._precise_exception_trace: bool = (
+            self._max_pending_exceptions > 1
```
manopapad, line 896: I guess the point of this condition is that you are going to block directly after the operation is launched anyway (assuming you also flush the scheduling window), so the traceback is going to be almost the same, and thus you can save a stack walk.

Note that in this case the traceback will contain some extraneous frames in the bottom (like the ones you cut with `skip = 2` when calling `capture_stacktrace` elsewhere):

```
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 5, in <module>
    cn.linalg.cholesky(cn.random.random((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/operation.py", line 167, in execute
    self._context.runtime.submit(self)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1040, in submit
    self.raise_exceptions()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1383, in raise_exceptions
    pending.raise_exception()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/exception.py", line 47, in raise_exception
    raise exn.with_traceback(self._tb)
cunumeric.linalg.exception.LinAlgError: Matrix is not positive definite
```

Since you went to the trouble of throwing such frames away in the "precise" case, you may care to handle this case better as well.

manopapad:
Note that if I hardcode `self._precise_exception_trace = True` in `runtime.py`, then the printed stacktrace I get in the end is not printed correctly. It looks like the saved stacktrace is "tacked on" to the stacktrace at the point where the final exception is raised:

```
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 5, in <module>
    cn.linalg.cholesky(cn.random.random((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/operation.py", line 167, in execute
    self._context.runtime.submit(self)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1040, in submit
    self.raise_exceptions()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1383, in raise_exceptions
    pending.raise_exception()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/exception.py", line 47, in raise_exception
    raise exn.with_traceback(self._tb)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 5, in <module>
    cn.linalg.cholesky(cn.random.random((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/operation.py", line 167, in execute
    self._context.runtime.submit(self)
cunumeric.linalg.exception.LinAlgError: Matrix is not positive definite
```

src/core/runtime/runtime.cc
manopapad: I see. In that case the `abort()` statement in the definition of `LEGATE_ABORT` is extraneous.
https://github.com/nv-legate/legate.core/blob/branch-22.07/src/legate_defines.h#L23

legate/core/runtime.py
magnatelee: The Python code is already a sequential bottleneck that limits the smallest problem size we can run, so I don't think we want to put any overhead on top of what we already have. Plus, the stacktrace generation is actually a pretty expensive operation, so I'd rather ask the users to mean it if they wanted it.

magnatelee:
> It looks like the saved stacktrace is "tacked on" to the stacktrace at the point where the final exception is raised:

I haven't found a way to get rid of those extra frames, as Python doesn't seem to allow me to do that. Note that the stacktrace I raise a pending exception with ([self._tb](https://github.com/magnatelee/legate.core/blob/exception-pr/legate/core/exception.py#L47)) only contains the stacktrace I capture. I don't think there's a way to completely replace the exception's stack with this one.

legate/core/runtime.py
magnatelee: 8cb62e4764e4bd4921fd2a1b86e848cb888bda32 should remove all legate.core frames.

manopapad:
> I don't think there's a way to completely replace the exception's stack with this one.

I think that's true. This used to be possible in Python 2 apparently, with the 3-argument `raise` statement, but no longer in Python 3 (probably for the better). However, I suggest we change the code like this:

```
--- a/legate/core/exception.py
+++ b/legate/core/exception.py
@@ -44,4 +44,4 @@ class PendingException:
         (exn_index, error_size) = struct.unpack("iI", buf[1:9])
         error_message = buf[9 : 9 + error_size].decode()
         exn = self._exn_types[exn_index](error_message)
-        raise exn.with_traceback(self._tb)
+        raise exn from Exception("Original location").with_traceback(self._tb)
```

which gives us somewhat better output:

```
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 5, in <module>
    cn.linalg.cholesky(cn.ones((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
Exception: Original location

The above exception was the direct cause of the following exception:

  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 9, in <module>
    cn.linalg.cholesky(cn.ones((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/operation.py", line 167, in execute
    self._context.runtime.submit(self)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1040, in submit
    self.raise_exceptions()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1383, in raise_exceptions
    pending.raise_exception()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/exception.py", line 47, in raise_exception
    raise exn from Exception("Original location").with_traceback(self._tb)
cunumeric.linalg.exception.LinAlgError: Matrix is not positive definite
```

legate/core/runtime.py
manopapad: That didn't do anything for me
```
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 434, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 5, in <module>
    cn.linalg.cholesky(cn.random.random((10,10)))
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 106, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 88, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 73, in cholesky
    return _cholesky(a)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/linalg.py", line 528, in _cholesky
    output._thunk.cholesky(input._thunk, no_tril=no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 75, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 1869, in cholesky
    cholesky(self, src, no_tril)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 183, in cholesky
    potrf_single(context, output.base)
  File "/Users/mpapadakis/cunumeric/cunumeric/linalg/cholesky.py", line 55, in potrf_single
    task.execute()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/operation.py", line 167, in execute
    self._context.runtime.submit(self)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1048, in submit
    self.raise_exceptions()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/runtime.py", line 1391, in raise_exceptions
    pending.raise_exception()
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/exception.py", line 47, in raise_exception
    raise exn.with_traceback(self._tb)
cunumeric.linalg.exception.LinAlgError: Matrix is not positive definite
```

magnatelee:
> However, I suggest we change the code like this:

Thanks for the suggestion. I tweaked your code and replaced `Exception("Original location")` with the original exception class and  error message, as I think the generic exception would be confusing; it'd be weird if the original stacktrace didn't contain what the original exception was.

legate/core/runtime.py
magnatelee: that's because what you're seeing is not the captured stacktrace, but one directly from Python. I updated the code to apply the precise stacktrace tunable uniformly, so hopefully the stacktrace would be less confusing.

legate/core/runtime.py
```diff
@@ -1334,6 +1374,20 @@ def get_barriers(self, count: int) -> tuple[Future, Future]:
...
+        future: Future,
+        tb: Optional[TracebackType] = None,
+    ) -> None:
+        exn = PendingException(exn_types, future, tb)
+        self._pending_exceptions.append(exn)
+
+    def raise_exceptions(self) -> None:
+        for pending in self._pending_exceptions:
+            pending.raise_exception()
+        self._pending_exceptions = []
```
manopapad, line 1389: Move this before the for-loop:

```
     def raise_exceptions(self) -> None:
-        for pending in self._pending_exceptions:
-            pending.raise_exception()
+        pending_exceptions = self._pending_exceptions
         self._pending_exceptions = []
+        for pending in pending_exceptions:
+            pending.raise_exception()
```

With the current implementation, if any of the `pending` operations throws, then the `_pending_exceptions` list doesn't get cleaned, so even if the exception is caught and execution proceeds, the exception will be thrown again at shutdown time, when legate calls `raise_exceptions` again.

magnatelee: A good catch. b860e8d9479b5c5fb4eb0cd3535bed237d4e8c12 I wonder whether we should add an API call for clients to manually block on exceptions. I feel client would need a programmable way to control how exceptions would be handled. that would also make exception tests for client libraries easier to write. for example, cuNumeric could (re-)export the blocking API for testing purposes, which would allow exception tests to always pass regardless of the configuration.

manopapad (approved):

src/core/runtime/runtime.cc
magnatelee: @manopapad I was wrong about the logger and `Logger::error` actually doesn't kill the process, so I added `LEGATE_ABORT`: c8abdf7065679ce9f6e79d5b32e2dbbd2c96af99

PR #274, opened 2022-06-15
use mpicc to compile the code when gasnet is enabled

eddy16112:


src/legate.mk
```diff
@@ -35,10 +35,37 @@ endif
 
 RM	:= rm
 
+# use mpi{cc,cxx,f90} compiler wrappers if USE_MPI=1 and we're not on a Cray system
```
magnatelee, line 38: `USE_MPI` should be `USE_GASNET`

eddy16112: Thanks, fixed now.

magnatelee (approved):

src/legate.mk
```diff
@@ -35,10 +35,37 @@ endif
...
-INC_FLAGS += -I$(MPI_HOME)/include
-LD_FLAGS += -L$(MPI_HOME)/lib -lmpi
+  ifeq (${CRAYPE_VERSION},)
+    # OpenMPI check
+    ifneq ($(strip $(shell __INTEL_POST_CFLAGS+=' -we10006' $(CC) -showme:compile 2>&1 > /dev/null; echo $$?)),0)
+      # MPICH check
+      ifneq ($(strip $(shell __INTEL_POST_CFLAGS+=' -we10006' $(CC) -show 2>&1 > /dev/null; echo $$?)),0)
+        export OMPI_CC  	:= $(CC)
+        export MPICH_CC  	:= $(CC)
+        CC					:= mpicc
```
manopapad, line 47: Can you fix the alignment here?

eddy16112: good catch, somehow the alignment is OK in my vscode. Is there a way to let pre-commit catch the format issue of makefile?

manopapad: I think we'll be moving away from GNU Make soon, so I wouldn't worry about this, but maybe we should consider adding auto-formatting for our upcoming cmakefiles @trxcllnt @marcinz ?

PR #275, opened 2022-06-15
let gasnet to handle mpi init and finalize

eddy16112:
I had a case that GASNet will crash if I finalize the MPI in the `collFinalize` function, so we will just let GASNet to handle the initialization and finalization of MPI.

### DIFF
--- a/src/core/comm/coll.cc
+++ b/src/core/comm/coll.cc
@@ -251,7 +251,10 @@ int collInit(int argc, char* argv[])
   int provided, init_flag = 0;
   CHECK_MPI(MPI_Initialized(&init_flag));
   if (!init_flag) {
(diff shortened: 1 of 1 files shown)


src/core/comm/coll.cc
```diff
@@ -291,7 +294,10 @@ int collFinalize(void)
   mpi_comms.clear();
   int fina_flag = 0;
   CHECK_MPI(MPI_Finalized(&fina_flag));
-  if (fina_flag == 0) { CHECK_MPI(MPI_Finalize()); }
+  if (fina_flag == 1) {
+    log_coll.fatal("MPI should not been finalized");
```
magnatelee, line 298: Typo: `"MPI should not have been finalized"`

eddy16112: Thanks, it should have been fixed now.

magnatelee (approved):

PR #280, opened 2022-06-20
Get rid of the LEGATE_MAX_CPU_COMMS

eddy16112:
The communicators are created on demand now.

### DIFF
--- a/legate/core/communicator.py
+++ b/legate/core/communicator.py
@@ -17,6 +17,8 @@ from __future__ import annotations
 from abc import ABC, abstractmethod, abstractproperty
 from typing import TYPE_CHECKING
 
+import numpy as native_np
+
 from . import FutureMap, Point, Rect
 from .launcher import TaskLauncher as Task
 
@@ -147,6 +149,12 @@ class CPUCommunicator(Communicator):
             side_effect=True,
         )
         cpucoll_id = task.execute_single()
+        cpucoll_id.wait()
+        coll_uid = native_np.frombuffer(
+            cpucoll_id.get_buffer(), dtype=native_np.int32
+        )[0]
+        self._runtime.issue_execution_fence(block=True)
+        self._runtime.core_library.legate_cpucoll_initcomm(coll_uid)
         task = Task(self._context, self._init_cpucoll_mapping, tag=self._tag)
         mapping_table_fm = task.execute(Rect([volume]))
         task = Task(self._context, self._init_cpucoll, tag=self._tag)
--- a/legate/core/communicator.py
+++ b/legate/core/communicator.py
@@ -153,7 +153,6 @@ class CPUCommunicator(Communicator):
         coll_uid = native_np.frombuffer(
             cpucoll_id.get_buffer(), dtype=native_np.int32
         )[0]
-        self._runtime.issue_execution_fence(block=True)
         self._runtime.core_library.legate_cpucoll_initcomm(coll_uid)
         task = Task(self._context, self._init_cpucoll_mapping, tag=self._tag)
         mapping_table_fm = task.execute(Rect([volume]))
--- a/legate/core/communicator.py
+++ b/legate/core/communicator.py
@@ -14,10 +14,9 @@
 #
 from __future__ import annotations
 
+import struct
 from abc import ABC, abstractmethod, abstractproperty
 from typing import TYPE_CHECKING
-
-import numpy as native_np
 
 from . import FutureMap, Point, Rect
 from .launcher import TaskLauncher as Task
@@ -113,7 +112,6 @@ class CPUCommunicator(Communicator):
         super().__init__(runtime)
         library = runtime.core_library
 
-        self._init_cpucoll_id = library.LEGATE_CORE_INIT_CPUCOLL_ID_TASK_ID
         self._init_cpucoll_mapping = (
             library.LEGATE_CORE_INIT_CPUCOLL_MAPPING_TASK_ID
         )
@@ -142,22 +140,14 @@ class CPUCommunicator(Communicator):
         return self._needs_barrier
 
     def _initialize(self, volume: int) -> FutureMap:
-        task = Task(
-            self._context,
-            self._init_cpucoll_id,
-            tag=self._tag,
-            side_effect=True,
-        )
-        cpucoll_id = task.execute_single()
-        cpucoll_id.wait()
-        coll_uid = native_np.frombuffer(
-            cpucoll_id.get_buffer(), dtype=native_np.int32
-        )[0]
-        self._runtime.core_library.legate_cpucoll_initcomm(coll_uid)
+        cpucoll_uid = self._runtime.core_library.legate_cpucoll_initcomm()
+        print(type(cpucoll_uid))
+        buf = struct.pack("i", cpucoll_uid)
+        cpucoll_uid_f = self._runtime.create_future(buf, len(buf))
         task = Task(self._context, self._init_cpucoll_mapping, tag=self._tag)
         mapping_table_fm = task.execute(Rect([volume]))
         task = Task(self._context, self._init_cpucoll, tag=self._tag)
-        task.add_future(cpucoll_id)
+        task.add_future(cpucoll_uid_f)
         for i in range(volume):
             f = mapping_table_fm.get_future(Point([i]))
             task.add_future(f)
--- a/legate/core/communicator.py
+++ b/legate/core/communicator.py
@@ -141,7 +141,6 @@ class CPUCommunicator(Communicator):
 
     def _initialize(self, volume: int) -> FutureMap:
         cpucoll_uid = self._runtime.core_library.legate_cpucoll_initcomm()
-        print(type(cpucoll_uid))
         buf = struct.pack("i", cpucoll_uid)
         cpucoll_uid_f = self._runtime.create_future(buf, len(buf))
         task = Task(self._context, self._init_cpucoll_mapping, tag=self._tag)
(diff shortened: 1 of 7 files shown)


legate/core/communicator.py
```diff
@@ -17,6 +17,8 @@
 from abc import ABC, abstractmethod, abstractproperty
 from typing import TYPE_CHECKING
 
+import numpy as native_np
```
magnatelee, line 20: use `np`

legate/core/communicator.py
```diff
@@ -147,6 +149,12 @@ def _initialize(self, volume: int) -> FutureMap:
             side_effect=True,
         )
         cpucoll_id = task.execute_single()
+        cpucoll_id.wait()
```
magnatelee, line 152: this is very bad. why do you need to block on the task? this will stall the whole downstream pipeline.

legate/core/communicator.py
```diff
@@ -147,6 +149,12 @@ def _initialize(self, volume: int) -> FutureMap:
...
+        coll_uid = native_np.frombuffer(
+            cpucoll_id.get_buffer(), dtype=native_np.int32
+        )[0]
+        self._runtime.issue_execution_fence(block=True)
```
magnatelee, line 156: this is also very bad. why do you need to block the entire execution here?

legate/core/communicator.py
```diff
@@ -147,6 +149,12 @@ def _initialize(self, volume: int) -> FutureMap:
...
+        self._runtime.core_library.legate_cpucoll_initcomm(coll_uid)
```
magnatelee, line 157: hoist this to a task, then you don't need to remove all the previous blocking calls.

src/core/comm/coll.cc
```diff
@@ -50,18 +50,15 @@ enum CollTag : int {
 
 static int mpi_tag_ub = 0;
 
-static std::vector<MPI_Comm> mpi_comms;
+static std::vector<MPI_Comm> mpi_comms(0);
```
magnatelee, line 53: this feels like an unnecessary micro-optimization that doesn't do any good. I wouldn't mind keeping around the default allocation for this vector even when no cpu communicators are used.

src/core/comm/coll.cc
```diff
@@ -270,16 +264,9 @@ int collInit(int argc, char* argv[])
   CHECK_MPI(MPI_Comm_get_attr(MPI_COMM_WORLD, MPI_TAG_UB, &tag_ub, &flag));
   assert(flag);
   mpi_tag_ub = *tag_ub;
-  // create mpi comms
-  mpi_comms.resize(MAX_NB_COMMS, MPI_COMM_NULL);
-  for (int i = 0; i < MAX_NB_COMMS; i++) { CHECK_MPI(MPI_Comm_dup(MPI_COMM_WORLD, &mpi_comms[i])); }
+  mpi_comms.clear();
```
magnatelee, line 267: I don't see why this is necessary. is this because you're unsure that the statically declared vector is not empty?

src/core/comm/coll.cc
```diff
@@ -270,16 +264,9 @@ int collInit(int argc, char* argv[])
...
 #else
-  thread_comms.resize(MAX_NB_COMMS);
-  for (int i = 0; i < MAX_NB_COMMS; i++) {
-    thread_comms[i].ready_flag = false;
-    thread_comms[i].buffers    = nullptr;
-    thread_comms[i].displs     = nullptr;
-  }
+  thread_comms.clear();
```
magnatelee, line 269: same comment here. why is this necessary?

src/core/comm/coll.cc
```diff
@@ -290,34 +277,46 @@ int collFinalize(void)
   assert(coll_inited == true);
   coll_inited = false;
 #ifdef LEGATE_USE_GASNET
-  for (int i = 0; i < MAX_NB_COMMS; i++) { CHECK_MPI(MPI_Comm_free(&mpi_comms[i])); }
+  for (int i = 0; i < mpi_comms.size(); i++) { CHECK_MPI(MPI_Comm_free(&mpi_comms[i])); }
```
magnatelee, line 280: let's use a range loop here, instead of a C style loop. (i.e., `for (MPI_Comm& mpi_comm : mpi_comms) { ... }`)

src/core/comm/coll.cc
eddy16112: I just want to make sure the vector is empty.

src/core/comm/coll.cc
eddy16112: Yes, I am not sure if the vector will be empty by default. Maybe I can do a check instead of using the clear.

legate/core/communicator.py
eddy16112: I just want to distinguish the numpy with cunumeric, so I rename it as native_np.

src/core/comm/coll.cc
eddy16112: fixed.

src/core/comm/coll.cc
eddy16112: fixed.

src/core/comm/coll.cc
eddy16112: fixed, will push it after the sapling is back to normal.

legate/core/communicator.py
eddy16112: I removed the `issue_execution_fence`, but kept the `wait`. It is difficult to convert this function into a task because the `MPI_Comm_dup` inside `legate_cpucoll_initcomm` needs to be called by the main thread of each node. Actually the `MPI_Comm_dup` acts like a `MPI_Barrier`, so it will block the calling thread of all nodes, so I do not think it is necessary to convert it into an index task. As long as I remove the `issue_execution_fence`, it will not block the Legion runtime to progress tasks on other threads.

legate/core/communicator.py
magnatelee: we don't import cunumeric in legate core, so there's no need for such a distinction.

legate/core/communicator.py
magnatelee: > It is difficult to convert this function into a task because the MPI_Comm_dup inside legate_cpucoll_initcomm needs to be called by the main thread of each node.

I don't think this is 100% accurate. I believe what you really want is to call that function only once per node. note that there's no such thing as "main thread" in the legate model. even the python task you're calling that function in isn't really the main thread, but just one of the threads managed by the runtime.

if you want to launch something per node instead of per processor, you just need to adjust the launch domain and update the mapper to do the right thing.

> Actually the MPI_Comm_dup acts like a MPI_Barrier, so it will block the calling thread of all nodes, so I do not think it is necessary to convert it into an index task.

there's a big difference between blocking in one of the leaf tasks and doing it in the main python task; the former only blocks execution of the downstream tasks, which is the right thing to do, but the latter blocks the submission of the downstream tasks and unrolling of the program's control flow, which should be avoided at all cost. I think you should convert it to an index task not to stall the runtime pipeline. it defeats the purpose of deferred execution.

> As long as I remove the issue_execution_fence, it will not block the Legion runtime to progress tasks on other threads.

no this is not true. what `issue_execution_fence(block=True)` really does is that it blocks on a future returned by the fence. so it's basically the same thing as you calling `wait` on a future from the task. so both the blocking on a fence and the `wait` call on a future block the progress of the program.

src/core/comm/coll.cc
magnatelee: the default constructor empties the vector, so why do you want to do this explicitly?

src/core/comm/coll.cc
```diff
@@ -264,9 +264,9 @@ int collInit(int argc, char* argv[])
   CHECK_MPI(MPI_Comm_get_attr(MPI_COMM_WORLD, MPI_TAG_UB, &tag_ub, &flag));
   assert(flag);
   mpi_tag_ub = *tag_ub;
-  mpi_comms.clear();
+  assert(mpi_comms.empty() == true);
```
magnatelee, line 267: isn't `empty` already returning a boolean value? what's the point of this comparison?

src/core/comm/coll.cc
```diff
@@ -286,7 +286,7 @@ int collFinalize(void)
     LEGATE_ABORT;
   }
 #else
-  for (int i = 0; i < thread_comms.size(); i++) { assert(thread_comms[i].ready_flag == false); }
+  for (ThreadComm& thread_comm : thread_comms) { assert(thread_comm.ready_flag == false); }
```
magnatelee, line 289: I'd use a usual bang operator for negation, though this is just a nitpick.

src/core/comm/coll.cc
```diff
@@ -299,10 +299,18 @@ int collGetUniqueId(int* id)
   return CollSuccess;
 }
 
-int collInitComm(int id)
+int collInitComm(void)
```
magnatelee, line 302: let's not use `void`.

src/core/comm/coll.cc
```diff
@@ -506,5 +515,5 @@ extern "C" {
 
 void legate_cpucoll_finalize(void) { legate::comm::coll::collFinalize(); }
 
-void legate_cpucoll_initcomm(int id) { legate::comm::coll::collInitComm(id); }
+int legate_cpucoll_initcomm(void) { return legate::comm::coll::collInitComm(); }
```
magnatelee, line 518: again, let's not use `void` in the signature for a function that takes no argument.

magnatelee:
left some nitpicking comments. otherwise looks good.

legate/core/communicator.py
```diff
@@ -140,17 +140,13 @@ def needs_barrier(self) -> bool:
...
 
     def _initialize(self, volume: int) -> FutureMap:
-        task = Task(
-            self._context,
-            self._init_cpucoll_id,
-            tag=self._tag,
-            side_effect=True,
-        )
-        cpucoll_id = task.execute_single()
+        cpucoll_uid = self._runtime.core_library.legate_cpucoll_initcomm()
```
magnatelee, line 143: fyi, technically, this code doesn't work when control replication is turned off, but no communicators would work either.

PR #282, opened 2022-06-21
WIP: Towards using Collective Instances

ipdemes:
What has been changed:

map_inline: if collective instances required , make all instances collective
map_task (legate_create_store):
  - for reduction instances only: 
    a) make reduction instances (with REDUCTION privilege or broadcasted ( read-only without projection) collective
    b) add tags to instances with REDUCTION privilege without projections

instance caching :  do not cache collective instances. (TODO: add separate caching logic for collective instances)

Since collective instances are not cached, performance is worse than before. At this point, I would like to get feedback on the approach taken.

### DIFF
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -661,7 +661,7 @@ void BaseMapper::map_task(const MapperContext ctx,
 
     for (auto failed_acquire : failed_acquires) {
       auto affected_mappings = instances_to_mappings[failed_acquire];
-      instances_to_mappings.erase(failed_acquire);
+      if (!failed_acquire.is_collective_instance()) instances_to_mappings.erase(failed_acquire);
 
       for (auto& mapping_idx : affected_mappings) {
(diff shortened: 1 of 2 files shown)


src/core/mapping/base_mapper.cc
```diff
@@ -766,17 +767,58 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
   LayoutConstraintSet layout_constraints;
   mapping.populate_layout_constraints(layout_constraints);
 
+  // QUESTION: is it safe to assume that colocated stores should have the same
+  // privileges / projection/ reduction???
+
```
ipdemes, line 772: @magnatelee : the question above is for you

magnatelee:
> Since collective instances are not cached, performance is worse than before. 

Did you actually measure the performance?

src/core/mapping/base_mapper.cc
ipdemes: But would it be ever possible that collective instances are colocated with not collective ?

PR #287, opened 2022-06-24
Add some types needed for work on cunumeric

bryevdv:
Needed to type cunumeric `runtime.py`


legate/core/context.py
```diff
@@ -243,7 +243,7 @@ def dispatch_single(self, op: Dispatchable[T]) -> T:
     def create_store(
         self,
         ty: Any,
-        shape: Optional[Shape] = None,
+        shape: Optional[Union[Shape, tuple[int, ...]]] = None,
```
bryevdv, line 246: We should probably pull `NdShape` down into legate, I will do that in a future PR

magnatelee (approved):

PR #291, opened 2022-07-01
fix the coll for non-mpi case.

eddy16112:
I used to get the address of a vector element and assign it to Communicator of each rank, but this is wrong, because vector is a dynamic data structure,  when it is resized, the address can be changed.

### DIFF
--- a/src/core/comm/coll.cc
+++ b/src/core/comm/coll.cc
@@ -52,7 +52,7 @@ static int mpi_tag_ub = 0;
 
 static std::vector<MPI_Comm> mpi_comms;
 #else  // undef LEGATE_USE_GASNET
-static std::vector<ThreadComm *> thread_comms;
+static std::vector<ThreadComm*> thread_comms;
 #endif
 
 static int current_unique_id = 0;
@@ -286,7 +286,7 @@ int collFinalize()
     LEGATE_ABORT;
   }
 #else
-  for (ThreadComm* thread_comm : thread_comms) { 
+  for (ThreadComm* thread_comm : thread_comms) {
     assert(!thread_comm->ready_flag);
     free(thread_comm);
   }
@@ -322,7 +322,7 @@ int collInitComm()
 #else
   assert(thread_comms.size() == id);
   // create thread comm
-  ThreadComm *thread_comm = (ThreadComm *)malloc(sizeof(ThreadComm));
+  ThreadComm* thread_comm = (ThreadComm*)malloc(sizeof(ThreadComm));
   thread_comm->ready_flag = false;
   thread_comm->buffers    = nullptr;
   thread_comm->displs     = nullptr;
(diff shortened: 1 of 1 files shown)


src/core/comm/coll.cc
```diff
@@ -102,24 +102,24 @@ int collCommCreate(CollComm global_comm,
...
+      thread_comms[global_comm->unique_id]->displs[i]  = nullptr;
     }
     __sync_synchronize();
-    thread_comms[global_comm->unique_id].ready_flag = true;
+    thread_comms[global_comm->unique_id]->ready_flag = true;
   }
   __sync_synchronize();
-  volatile ThreadComm* data = &(thread_comms[global_comm->unique_id]);
-  while (data->ready_flag != true) { data = &(thread_comms[global_comm->unique_id]); }
-  global_comm->comm = &(thread_comms[global_comm->unique_id]);
```
eddy16112, line 122: The reason of this PR is because of this line. When the thread_comms vector get resized, the address is changed.

src/core/comm/coll.cc
```diff
@@ -52,7 +52,7 @@ static int mpi_tag_ub = 0;
 
 static std::vector<MPI_Comm> mpi_comms;
 #else  // undef LEGATE_USE_GASNET
-static std::vector<ThreadComm> thread_comms;
+static std::vector<ThreadComm*> thread_comms;
```
ipdemes, line 55: does this vector ever get relocated? Did you try running this multi-node?

eddy16112: Line 54 shows this vector is used for single node run only.

ipdemes:
@eddy16112 : LGTM

src/core/comm/coll.cc
ipdemes: I see, Ok

eddy16112:
> @eddy16112 : LGTM

Thank you

src/core/comm/coll.cc
magnatelee: yeah and even if the address was stable, it'd be bad for performance to copy the whole structs.

src/core/comm/coll.cc
```diff
@@ -319,10 +322,10 @@ int collInitComm()
 #else
   assert(thread_comms.size() == id);
   // create thread comm
-  ThreadComm thread_comm;
-  thread_comm.ready_flag = false;
-  thread_comm.buffers    = nullptr;
-  thread_comm.displs     = nullptr;
+  ThreadComm* thread_comm = (ThreadComm*)malloc(sizeof(ThreadComm));
```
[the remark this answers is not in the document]
magnatelee: another comment: if these thread communicators are solely owned by the core, then you can use `std::unique_ptr` instead. then you don't have to manually deallocate them.

PR #293, opened 2022-07-06
NVTX: Use RangePush and Domain

evanramos-nvidia:
This branch makes two changes:

- Replaces the use of the nvtxRangeStart/RangeEnd API with RangePush/RangePop
- Tags NVTX events with a domain specific to legate.core for easy filtering

### DIFF
--- a/src/core/runtime/runtime.cc
+++ b/src/core/runtime/runtime.cc
@@ -45,6 +45,11 @@ static const char* const core_library_name = "legate.core";
 
 /*static*/ bool Core::has_socket_mem = false;
 
+#ifdef LEGATE_USE_CUDA
+/*static*/ const char* const nvtx::Range::domainName = "Legate";
+/*static*/ nvtxDomainHandle_t nvtx::Range::domain;
+#endif
+
 /*static*/ void Core::parse_config(void)
 {
 #ifndef LEGATE_USE_CUDA
@@ -83,6 +88,10 @@ static const char* const core_library_name = "legate.core";
   parse_variable("LEGATE_EMPTY_TASK", use_empty_task);
   parse_variable("LEGATE_SYNC_STREAM_VIEW", synchronize_stream_view);
   parse_variable("LEGATE_LOG_MAPPING", log_mapping_decisions);
+
+#ifdef LEGATE_USE_CUDA
+  nvtx::Range::initialize();
+#endif
 }
 
 static void extract_scalar_task(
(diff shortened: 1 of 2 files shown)


magnatelee:
@evanramos-nvidia thanks for the PR. can you briefly explain me how Domains manifest in Nsight timelines? I was trying to find the domain name of my ranges, but couldn't, probably because I'm not using the latest Nsight?

src/core/runtime/runtime.cc
```diff
@@ -76,6 +80,10 @@ static const char* const core_library_name = "legate.core";
 
   const char* sync_stream_view = getenv("LEGATE_SYNC_STREAM_VIEW");
   if (sync_stream_view != nullptr && atoi(sync_stream_view) > 0) synchronize_stream_view = true;
+
+#ifdef LEGATE_USE_CUDA
+  nvtx::Range::initialize(core_library_name);
```
magnatelee, line 85: I think we would want a separate `Domain` per Legate library, instead of putting all ranges under the same domain. A prerequisite for that is a centralized registry of all legate libraries, which doesn't exist in this branch. we can certainly merge this PR and tackle the issue in a follow-on PR, as the registry can take some time to implement properly.

evanramos-nvidia:
> @evanramos-nvidia thanks for the PR. can you briefly explain me how Domains manifest in Nsight timelines? I was trying to find the domain name of my ranges, but couldn't, probably because I'm not using the latest Nsight?

No problem! Domains are listed on the left side of the timeline view. I've highlighted one of the instances of the Legate domain in this screenshot:

Screenshot from 2022-07-18 17-57-25

src/core/runtime/runtime.cc
evanramos-nvidia: After consulting with @jcohen-nvidia, we think your idea would be better suited to the Categories feature in NVTX rather than creating separate domains, which would come with a performance cost and also complicate the implementation of a toggle switch for tracing Legate in Nsight Systems.

evanramos-nvidia:
I updated my branch to use `Legate` as the domain name string rather than `legate.core`, in line with other tracing functionality in Nsys.

src/core/runtime/runtime.cc
magnatelee: I repeat here what I said in the meeting to remind ourselves: I still think each of Legate libraries should have a separate `Domain` , as they are physically different libraries with separate APIs that the user may or may not choose to trace, and each library might want to use its own set of categories for the domain operations. We can merge this PR as-is for now and I can make a follow-on PR for a per-library `Domain` (which would naturally fit to this `LibraryContext`: https://github.com/nv-legate/legate.core/blob/branch-22.07/src/core/runtime/context.h#L65-L105).

magnatelee (approved):

PR #306, opened 2022-07-20
Broaden add_scalar_arg type

bryevdv:
Original type for `add_scalar_arg` was too narrow, discovered in https://github.com/nv-legate/cunumeric/pull/438

### DIFF
--- a/legate/core/launcher.py
+++ b/legate/core/launcher.py
@@ -120,7 +120,7 @@ class ScalarArg:
         self,
         core_types: ty.TypeSystem,
         value: Any,
-        dtype: Union[DTType, tuple[DTType, ...]],
+        dtype: Union[DTType, tuple[DTType]],
         untyped: bool = True,
     ) -> None:
         self._core_types = core_types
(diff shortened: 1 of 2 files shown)


legate/core/launcher.py
```diff
@@ -747,7 +747,7 @@ def __del__(self) -> None:
     def add_scalar_arg(
         self,
         value: Any,
-        dtype: DTType,
+        dtype: Union[DTType, tuple[DTType, ...]],
```
manopapad, line 750: I think the current usage is more restricted than this; only singleton tuples are supported. Therefore, I think we could strengthen the type to `Union[DTType, tuple[DTType]]` -- assuming mypy is ok with that.

bryevdv: Thanks for the clarification https://github.com/nv-legate/legate.core/pull/306/commits/97c35c87714929bf02c28b9f26901e5480b4cacf

manopapad (approved):

PR #307, opened 2022-07-27
Required core changes for cunumeric#474

manopapad:


bryevdv (approved):

legate/core/store.py
```diff
@@ -988,6 +988,9 @@ def storage(self) -> Union[RegionField, Future]:
             )
         return self._storage.data
 
+    def get_root(self) -> Storage:
```
magnatelee, line 991: I'm not comfortable to expose `Storage`. (for the same reason, I think we should get rid of the `storage` property.) can we instead have a query method like `has_same_storage_as` that takes another Store?

manopapad: 892a93f796cf8b890940dff727974c031f9eaf92

magnatelee (approved):

PR #308, opened 2022-07-29
add mypy pre-commit hook

bryevdv:
This PR adds a [pre-commit hook for mypy](https://github.com/pre-commit/mirrors-mypy). It should run `mypy legate` whenever there are relevant file changes.

### DIFF
--- a/legate/core/corelib.py
+++ b/legate/core/corelib.py
@@ -1,4 +1,5 @@
 # Copyright 2021-2022 NVIDIA Corporation
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


manopapad (approved):
Works for me!

legate/core/corelib.py
```diff
@@ -1,4 +1,3 @@
-# Copyright 2021-2022 NVIDIA Corporation
 #
```
manopapad, line 1: I assume we still want a copyright header?

bryevdv: Oh weird, not sure why/how that happened

bryevdv, line 1:
```suggestion
# Copyright 2021-2022 NVIDIA Corporation
```

bryevdv: fixed (two commits, tried to be to clever with GH web ui suggestion)

PR #311, opened 2022-07-31
src: switch to abolute include path for pthread_barrier.h

rohany:
This commit switches the relative include of "pthread_barrier.h" to an
absolute path off of core/ so that client libraries that include
"core/comm.h" can find the file.

### DIFF
--- a/src/core.mk
+++ b/src/core.mk
@@ -61,7 +61,7 @@ INSTALL_HEADERS = legate.h                        \
 									legate_preamble.h               \
 									core/legate_c.h                 \
 									core/comm/communicator.h        \
-									core/comm/coll.h		        \
+									core/comm/coll.h                \
 									core/comm/pthread_barrier.h     \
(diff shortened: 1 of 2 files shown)


src/core.mk
```diff
@@ -61,7 +61,8 @@ INSTALL_HEADERS = legate.h                        \
 									legate_preamble.h               \
 									core/legate_c.h                 \
 									core/comm/communicator.h        \
-									core/comm/coll.h		\
+									core/comm/coll.h		        \
```
magnatelee, line 64: can you align this backslash with the others?

magnatelee (approved):

magnatelee:
left a minor comment. looks good to me otherwise.

src/core.mk
rohany: Hmm, it looked aligned on my editor. Let me take a look again

manopapad (approved):

PR #316, opened 2022-08-02
Update docs to match cunumeric

bryevdv:
Quick PR to get the `legate.core` docs building at all, and closer to the structure of the `cunumeric` docs.  This PR:

* updates to use pydata sphinx theme (to match cunumeric)
* removes unused `changelog.rst`
* adds symlinks for README and CONTRIBUTING (to match cunumeric)
* removes `git.rst` (would propose adding to a "dev guide" at a later date, if desirerd)
* updates API ref guide structure to use autosummary (to match cunumeric)
* minor verbiage tweaks for missing refs or to match cunumeric
* removes autostructify extension which does not seem worth the bother to keep

With these changes the docs should build with any of the current conda envs.


docs/legate/core/source/api/index.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. currentmodule:: legate.core
+
+.. toctree::
+   :maxdepth: 1
+
+.. autosummary::
+   :toctree: generated/
+
+    Point
```
magnatelee, line 13: Note that the classes listed here aren't necessarily what we want to expose to the developers, as they are Legion abstractions and not Legate's. I'm fine with merging this PR, but keep in mind that we will need to revisit this API reference in the future.

magnatelee (approved):

docs/legate/core/source/api/index.rst
bryevdv: :+1:  For sure, just duplicates the pre-existing structure and gets it back to successfully building.

PR #319, opened 2022-08-04
avoiding division by 0 in slicing

ipdemes:


magnatelee:
I feel we shouldn't be in this code path if the tile shape is empty. can you instead add a condition that the store is empty to this if statement? https://github.com/nv-legate/legate.core/blob/branch-22.10/legate/core/store.py#L1278 that will prevent a tile partition from ever being created for an empty store.

ipdemes:
@magnatelee : I have tried adding `or self.size ==0` (adding logic for size method), but this didn't help. I actually want slicing to happen (I call project on (0,1) shape)

magnatelee:
@ipdemes if that's the case, then I think you should instead change how the slicing is done. any projection on an empty store can reuse the storage of the original store, instead of creating a sub-store.

ipdemes:
@magnatelee OK. I tried to reuse storage when size==0 in my latest commit. Please have a look

legate/core/store.py
```diff
@@ -950,6 +950,15 @@ def ndim(self) -> int:
...
             return self._shape.ndim
 
+    @property
+    def size(self) -> int:
+        s = 1
+        if self._ndim == 0:
+            return s
+        for p in self.shape:
+            s *= p
+        return s
```
bryevdv, line 960: FYI `math.prod` was added in 3.8 
```
In [2]: math.prod((1,2,3))
Out[2]: 6

In [3]: math.prod(())
Out[3]: 1
```
(assuming shape is empty when ndim is zero)

legate/core/store.py
```diff
@@ -727,10 +727,10 @@ def slice(self, tile_shape: Shape, offsets: Shape) -> Storage:
         # region in all dimensions of the parent region, then we'll make
         # a disjoint tiled partition with as many children as possible
         shape = self.get_root().extents
+
```
magnatelee, line 730: please revert this bogus change

legate/core/store.py
```diff
@@ -727,10 +727,10 @@ def slice(self, tile_shape: Shape, offsets: Shape) -> Storage:
...
         can_tile_completely = (offsets % tile_shape).sum() == 0 and (
             shape % tile_shape
         ).sum() == 0
-
```
magnatelee, line 733: please revert this bogus change

legate/core/store.py
magnatelee: this is the code from when the minimum python version was 3.7. now we drop python 3.7 from the list of supported versions, we can and should replace this with `math.prod`. (and I suspect this isn't the only place where we roll a custom `prod` implementation.)

legate/core/store.py
```diff
@@ -1119,6 +1131,10 @@ def slice(self, dim: int, sl: slice) -> Store:
             )
 
         size = self.shape[dim]
+
+        if size == 0 and sl == slice(None):
```
magnatelee, line 1135: I think we can change this to `if size == 0 or sl == slice(None):`, as a slicing is no-op with either of them.

FYI, it looks like NumPy allows slice indices to be out-of-bounds, which would end up resulting an empty slice, whereas we raise an exception below. This is something we should fix after this PR is merged.

legate/core/store.py
ipdemes: Yes, this was basically copy-paste from cunumeirc. I have fixed it with the latest commit

legate/core/store.py
ipdemes: makes sense.

magnatelee (approved):

PR #323, opened 2022-08-08
Add CMake build for C++ and scikit-build infrastructure for Python package installation

jjwilke:
Clone of https://github.com/nv-legate/legate.core/pull/198 with special branch name so CI will test the legate.core and cunumeric PR's together.


trxcllnt (approved):

conda/environment-test-3.10.yml
```diff
@@ -1,14 +1,20 @@
...
   # build
-  - cutensor>=1.3.3
+  - git
   - nccl
+  - make
+  - zlib
+  - cmake
+  - ninja
+  - c-compiler
+  - cxx-compiler
```
bryevdv, line 15: - cmake>=3.24
  - ninja
  - c-compiler
  - cxx-compiler
  - gcc_linux-64 # [linux64]
  - sysroot_linux-64==2.17 # [linux64]

bryevdv, line 15:
```suggestion
  - cmake>=3.24
  - ninja
  - c-compiler
  - cxx-compiler
  - gcc_linux-64 # [linux64]
  - sysroot_linux-64==2.17 # [linux64]
```

conda/environment-test-3.8.yml
```diff
@@ -1,14 +1,20 @@
...
   # build
-  - cutensor>=1.3.3
+  - git
   - nccl
+  - make
+  - zlib
+  - cmake
+  - ninja
+  - c-compiler
+  - cxx-compiler
```
bryevdv, line 15:
```suggestion
  - cmake>=3.24
  - ninja
  - c-compiler
  - cxx-compiler
  - gcc_linux-64 # [linux64]
  - sysroot_linux-64==2.17 # [linux64]
```

conda/environment-test-3.9.yml
```diff
@@ -1,14 +1,20 @@
...
   # build
-  - cutensor>=1.3.3
+  - git
   - nccl
+  - make
+  - zlib
+  - cmake
+  - ninja
+  - c-compiler
+  - cxx-compiler
```
bryevdv, line 15:
```suggestion
  - cmake>=3.24
  - ninja
  - c-compiler
  - cxx-compiler
  - gcc_linux-64 # [linux64]
  - sysroot_linux-64==2.17 # [linux64]
```

conda/environment-test-3.10.yml
```diff
@@ -1,14 +1,22 @@
...
+name: legate-core-test
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - python=3.10
 
   # build
-  - cutensor>=1.3.3
+  - git
```
manopapad, line 8: Is git really necessary here?

conda/environment-test-3.10.yml
```diff
@@ -1,14 +1,22 @@
...
   - nccl
+  - make
+  - zlib
+  - cmake>=3.24
+  - ninja
```
manopapad, line 13: Do we need both make and ninja?

conda/environment-test-3.10.yml
```diff
@@ -1,14 +1,22 @@
...
+  - c-compiler
+  - cxx-compiler
+  - gcc_linux-64 # [linux64]
+  - sysroot_linux-64==2.17 # [linux64]
```
manopapad, line 17: Are we doing something to remove the `-O?` (and other) flags that conda-forge's compiler package injects into `CXXFLAGS` and friends? See https://github.com/conda-forge/ctng-compiler-activation-feedstock/issues/80

If we're not, then these compilers don't allow us to do debug builds, and thus are not useful for development.

conda/environment-test-3.10.yml
```diff
@@ -17,6 +25,7 @@ dependencies:
   - pyarrow>=5
   - scipy
   - typing_extensions
+  - llvm-openmp
```
manopapad, line 28: Does the gcc package not come with openmp support? Do we need to also pull llvm here?

conda/environment-test-3.10.yml
```diff
@@ -1,14 +1,23 @@
...
+  - git
   - nccl
+  - make
+  - zlib
+  - cmake>=3.24
+  - ninja
+  - openmpi
+  - c-compiler
+  - cxx-compiler
+  - gcc_linux-64 # [linux64]
```
manopapad, line 17: Do we need to specify gcc specifically? Especially if later we request llvm-openmp

conda/environment-test-3.10.yml
trxcllnt: This is to align with the package [build and runtime dependency](https://github.com/jjwilke/legate.core/blob/add-cmake-pr/conda/conda-build/meta.yaml#L92), which @m3vaz said we're using because it's faster than the regular `openmp` package.

conda/environment-test-3.10.yml
trxcllnt: I don't know, see also: https://github.com/nv-legate/legate.core/pull/323#discussion_r957830930

conda/environment-test-3.10.yml
trxcllnt: CMake will append [our flags](https://github.com/jjwilke/legate.core/blob/add-cmake-pr/CMakeLists.txt#L70-L80) after the ones specified by conda, and ours will take precedence. There may be extra flags we need to remove, though I haven't had any issues so far.

conda/environment-test-3.10.yml
trxcllnt: Some people may prefer make over ninja, but I'm fine going w/ only ninja.

conda/environment-test-3.10.yml
trxcllnt: Git is used during the build to clone dependencies from source so they can be built.

conda/environment-test-3.10.yml
m3vaz: @trxcllnt @manopapad So openmp support is a bit messy. There are several openmp packages and I want to make sure we're clear on which we're talking about.
`openmp` is a dead package, and ***nothing*** should point at it.
`llvm-openmp` is generally a drop-in replacement for the old `openmp` package
`libgomp` is the GNU openmp package

There are a couple of reasons to prefer `llvm-openmp` over `libgomp`:
1) For reasons, gcc will output symbols that work with either `libgomp` or `llvm-openmp`, but clang will only produce symbols that work with `llvm-openmp`. 
2) Last I checked, `llvm-openmp` provides a wider platform support matrix (including support for Mac)
3) `libgomp` is not fork-safe

Does that clarify the situation?
This may also help: https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp

conda/environment-test-3.10.yml
m3vaz: IIRC, the compiler does not necessarily come with a specific openmp implementation, and package maintainers/users have a choice of which openmp implementation to use.

cmake/thirdparty/get_legion.cmake
```diff
@@ -0,0 +1,193 @@
...
+
+      set(_nvcc_flags ${CUDA_NVCC_FLAGS})
+      if(NOT "${_nvcc_flags}" MATCHES "-std=")
+        list(APPEND _nvcc_flags "-std=c++${_cuda_std}")
+      endif()
+
+      # Get the `stubs/libcuda.so` path so we can set CMAKE_LIBRARY_PATH for FindCUDA.cmake
+
+      # Prefer users' CUDA_PATH envvar (if set)
+      set(_cuda_stubs "$ENV{CUDA_PATH}")
```
manopapad, line 89: Can we check in all cases that `<candidate-dir>/stubs/libcuda.so` exists before accepting it? e.g. something like
```
does ENV{CUDA_PATH}/stubs/libcuda.so exist?
if yes then _cuda_stubs += ENV{CUDA_PATH}/stubs otherwise:
does ENV{CUDA_LIB_PATH}/stubs/libcuda.so exist?
if yes then _cuda_stubs += ENV{CUDA_LIB_PATH}/stubs otherwise:
...
```

Issue #329, opened 2022-08-12
In-bounds but empty slices not supported


manopapad:
The following program, run in deferred mode:

```
x = cn.ones((2,))
y = x[1:1]
print(y)
```

crashes inside the core with:

```
Traceback (most recent call last):
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 440, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 47, in <module>
    y = x[1:1]
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 110, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 123, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 912, in __getitem__
    return ndarray(shape=None, thunk=self._thunk.get_item(key))
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 798, in get_item
    result = self._get_view(key)
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 708, in _get_view
    store = store.slice(dim + shift, k)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/store.py", line 1159, in slice
    storage = self._storage.slice(
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/store.py", line 732, in slice
    can_tile_completely = (offsets % tile_shape).sum() == 0 and (
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/shape.py", line 191, in __mod__
    return Shape(tuple(a % b for (a, b) in zip(lh, rh)))
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/shape.py", line 191, in <genexpr>
    return Shape(tuple(a % b for (a, b) in zip(lh, rh)))
ZeroDivisionError: integer division or modulo by zero
```

ipdemes:
I am trying to ix this issue on CuNumeric side

ipdemes:
fixed in https://github.com/nv-legate/cunumeric/pull/574

PR #332, opened 2022-08-16
fixing issue #329

ipdemes:
This PR is fixing https://github.com/nv-legate/legate.core/issues/329


magnatelee:
@ipdemes I don't like the tile shape is ever being empty. I feel the client library is supposed to deal with these corner cases and not the core. the same applies to [your other PR](https://github.com/nv-legate/legate.core/pull/334).

ipdemes:
closing since the fix will go to CuNumeric

Issue #317, opened 2022-08-03
Cannot inline-map 0d RegionField-backed Stores


manopapad:
This program:

```
import cunumeric as cn
idx = cn.zeros((3,))[2:3].reshape(())
print(idx)
```

fails with:

```
Traceback (most recent call last):
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 440, in legion_python_main
    run_path(args[start], run_name='__main__')
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legion_top.py", line 255, in run_path
    exec(code, module.__dict__, module.__dict__)
  File "a.py", line 9, in <module>
    print(idx)
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 110, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 1579, in __str__
    return str(self.__array__())
  File "/Users/mpapadakis/cunumeric/cunumeric/coverage.py", line 110, in wrapper
    return func(*args, **kwargs)
  File "/Users/mpapadakis/cunumeric/cunumeric/array.py", line 671, in __array__
    return self._thunk.__numpy_array__()
  File "/Users/mpapadakis/cunumeric/cunumeric/deferred.py", line 283, in __numpy_array__
    alloc = self.base.get_inline_allocation(self.context)
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/store.py", line 1221, in get_inline_allocation
    return self._storage.get_inline_allocation(
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/store.py", line 771, in get_inline_allocation
    return self.data.get_inline_allocation(
  File "/Users/mpapadakis/legate.core/install/lib/python3.8/site-packages/legate/core/store.py", line 346, in get_inline_allocation
    func = getattr(
  File "/Users/mpapadakis/opt/miniconda3/envs/legate/lib/python3.8/site-packages/cffi/api.py", line 912, in __getattr__
    make_accessor(name)
  File "/Users/mpapadakis/opt/miniconda3/envs/legate/lib/python3.8/site-packages/cffi/api.py", line 907, in make_accessor
    raise AttributeError(name)
AttributeError: legion_physical_region_get_field_accessor_array_0d_with_transform
```

magnatelee:
[...]

PR #340, opened 2022-08-20
Correctly inline map 0d stores

magnatelee:
The inline mapping code had been crashing when the store was 0D. This PR fixes that bug.


magnatelee:
cc @manopapad this and https://github.com/nv-legate/cunumeric/pull/551 fix #317.

PR #349, opened 2022-08-25
Misc minor cleanup

bryevdv:
This PR addresses some issues reported by a local run of CodeQL using the full LGTM query set.

### DIFF
--- a/legate/core/projection.py
+++ b/legate/core/projection.py
@@ -67,7 +67,7 @@ class ProjExpr:
         return (
             self._dim == other._dim
             and self._weight == other._weight
-            and self._offset == self._offset
+            and self._offset == other._offset
         )
 
     def __mul__(self, other: int) -> ProjExpr:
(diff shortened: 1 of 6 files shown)


legate/core/projection.py
```diff
@@ -67,7 +67,7 @@ def __eq__(self, other: object) -> bool:
         return (
             self._dim == other._dim
             and self._weight == other._weight
-            and self._offset == self._offset
+            and self._offset == other._offset
```
bryevdv, line 70: I assume this was just a copy-pasta bug

legate/core/solver.py
```diff
@@ -80,14 +80,14 @@ def record(self, var1: T, var2: T) -> None:
...
+            self._merge(var1, var2)
+        elif found1:  # first only
             self._update(var1, var2)
-        elif found2:
+        elif found2:  # second only
             self._update(var2, var1)
-        else:
-            self._merge(var1, var2)
+        else:  # neither
+            self._add(var1, var2)
```
bryevdv, line 90: Previously the else case (both true) would never be reached because either of the previous branches would trigger first.  cc @magnatelee

magnatelee: good catch!

magnatelee (approved):

PR #352, opened 2022-08-30
fix cpu communicator for omp

eddy16112:


rohany:
Change looks fine to me, but I defer to @magnatelee for the final review.

legate/core/communicator.py
```diff
@@ -120,7 +120,16 @@ def __init__(self, runtime: Runtime) -> None:
         )
         self._init_cpucoll = library.LEGATE_CORE_INIT_CPUCOLL_TASK_ID
         self._finalize_cpucoll = library.LEGATE_CORE_FINALIZE_CPUCOLL_TASK_ID
-        self._tag = library.LEGATE_CPU_VARIANT
+        num_omps = int(
```
magnatelee, line 123: let's move this to `Runtime` and make the communicator use it, as I think it's better to consolidate all tunable lookups in one place for code maintenance.

magnatelee:
left a minor change request. otherwise looks good.

legate/core/communicator.py
eddy16112: I have pushes a new commit for it. Besides omps, I also add tunable lookups for CPUs and GPUs.

magnatelee (approved):

PR #353, opened 2022-08-31
A custom task wrapper for efficient handling of return values

magnatelee:
The core has been using Legion's generic task wrapper for its return value handling, which is necessary for the general case but inefficient for the majority of cases with one or even no return value. This PR replaces that generic wrapper with a tailor-made wrapper for Legate tasks that can optimize away the return value handling for common cases.


magnatelee:
> Just a note that (AFAIU) even with this change we shouldn't expect to see scalar values remain completely on framebuffer. The buffer_ created for a non-read-only FutureWrapper is placed on zero-copy memory, so any reduce_future_maps will presumably happen there, and the first task which reads the reduction result will pull it into the framebuffer.

Yes, I was aware of this issue and just pushed another change to use framebuffer allocations for return values in the common case (i.e., where there's no exception and only a single return value). Please take a look.

src/core/mapping/core_mapper.cc
```diff
@@ -374,6 +378,10 @@ void CoreMapper::map_future_map_reduction(const MapperContext ctx,
                                           FutureMapReductionOutput& output)
 {
```
manopapad, line 379: We probably don't want to do this in the case of `LEGATE_CORE_JOIN_EXCEPTION_OP` (in that case we probably want to let Legion do the default behavior, of placing a single value on system memory).

src/core/task/return.cc
```diff
@@ -199,6 +197,66 @@ void ReturnValues::legion_deserialize(const void* buffer)
...
+  if (kind == Processor::TOC_PROC) {
+    // FIXME: We don't currently have a good way to defer the return value packing on GPUs,
+    //        as doing so would require the packing to be chained up with all preceding kernels,
+    //        potentially launched with different streams, within the task. Until we find
+    //        the right approach, we simply synchronize the device before proceeding.
+    CHECK_CUDA(cudaDeviceSynchronize());
+
+    size_t return_size = legion_buffer_size();
+
+    auto mem_kind = (return_values_.size() == 1 && !has_exception_)
```
manopapad, line 218: We probably also want to skip this optimization if we're dealing with a task that has a single output region (and no other return values).

manopapad (approved):
No issues found, some improvements suggested.

I was actually thinking we could go one step further, and place the `UntypedDeferredValue`s for `FutureWrapper` buffers directly on framebuffer on construction, let the task body operate exclusively on framebuffer, then on task exit return the `UntypedDeferredValue` instance directly if we're only dealing with a single non-exception-bearing future. I think in that case we wouldn't even need to explicitly `cudaDeviceSync`, since that would be covered automatically by Realm's context sync.

The downside being that now `ReturnValues::legion_serialize` cannot assume that all pointers are on host-accessible memory; we would need to keep track of which pointers came from `FutureWrapper` framebuffer instances, and `cudaMemppy` as appropriate when packing.

I could look into this on a follow-up PR.

src/core/mapping/core_mapper.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/353/commits/72bc605e5130966e586739c4907a4707a12f608d

magnatelee:
> I was actually thinking we could go one step further, and place the UntypedDeferredValues for FutureWrapper buffers directly on framebuffer on construction, let the task body operate exclusively on framebuffer, then on task exit return the UntypedDeferredValue instance directly if we're only dealing with a single non-exception-bearing future. I think in that case we wouldn't even need to explicitly cudaDeviceSync, since that would be covered automatically by Realm's context sync.

I had the same idea, but wasn't sure how much work it would be so I didn't want to include that in this PR. But I decided to finish what I started and here's the change: https://github.com/nv-legate/legate.core/pull/353/commits/d3bc30e3c7e7524325d8d3c284213271b8e17135

src/core/data/store.cc
```diff
@@ -123,12 +133,27 @@ FutureWrapper::FutureWrapper(
...
-    auto p_init_value = initialize ? future_.get_buffer(mem_kind) : nullptr;
-    buffer_           = UntypedDeferredValue(field_size, mem_kind, p_init_value);
+    auto proc     = Processor::get_executing_processor();
+    auto mem_kind = proc.kind() == Processor::Kind::TOC_PROC ? Memory::Kind::GPU_FB_MEM
+                                                             : Memory::Kind::SYSTEM_MEM;
+    if (initialize) {
+      auto p_init_value = future_.get_buffer(mem_kind);
+#ifdef LEGATE_USE_CUDA
+      if (mem_kind == Memory::Kind::GPU_FB_MEM) {
+        // TODO: This should be done by Legion
```
manopapad, line 146: I guess `UntypedDeferredValue(field_size, mem_kind, p_init_value)` doesn't work if `mem_kind == Memory::Kind::GPU_FB_MEM`? Is this something we should report to be fixed in Legion?

src/core/data/store.cc
```diff
@@ -123,12 +133,27 @@ FutureWrapper::FutureWrapper(
...
+        buffer_ = UntypedDeferredValue(field_size, mem_kind);
+        AccessorWO<int8_t, 1> acc(buffer_, field_size, false);
+        auto stream = cuda::StreamPool::get_stream_pool().get_stream();
+        CHECK_CUDA(
+          cudaMemcpyAsync(acc.ptr(0), p_init_value, field_size, cudaMemcpyDeviceToDevice, stream));
```
manopapad, line 151: Do we also need to deviceSync in this case, in case the task performs cuda operations on a different stream?

src/core/data/store.cc
```diff
@@ -163,15 +188,16 @@ void FutureWrapper::initialize_with_identity(int32_t redop_id)
   assert(redop->sizeof_lhs == field_size_);
 #endif
   auto identity = redop->identity;
-  memcpy(ptr, identity, field_size_);
+#ifdef LEGATE_USE_CUDA
+  if (buffer_.get_instance().get_location().kind() == Memory::Kind::GPU_FB_MEM) {
+    auto stream = cuda::StreamPool::get_stream_pool().get_stream();
+    CHECK_CUDA(cudaMemcpyAsync(ptr, identity, field_size_, cudaMemcpyHostToDevice, stream));
```
manopapad, line 194: Similar to above, should we deviceSync here?

src/core/data/store.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/353/commits/7f07fa116925cb78585f3ae4b161d64c8b4bb148

src/core/data/store.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/353/commits/7f07fa116925cb78585f3ae4b161d64c8b4bb148

src/core/task/return.cc
```diff
@@ -137,17 +137,66 @@ ReturnValue ReturnedException::pack() const
...
+
+template <bool PACK_SIZE>
+int8_t* pack_return_value(int8_t* target, const ReturnValue& value)
+{
+  if constexpr (PACK_SIZE) {
+    *reinterpret_cast<uint32_t*>(target) = value.second;
+    target += sizeof(uint32_t);
+  }
+
+  AccessorWO<int8_t, 1> acc(value.first, value.second, false);
```
manopapad, line 158: AccessorRO?

src/core/task/return.cc
```diff
@@ -137,17 +137,66 @@ ReturnValue ReturnedException::pack() const
...
+
+template <bool PACK_SIZE>
+int8_t* pack_return_value(int8_t* target, const ReturnValue& value, cuda::StreamView& stream)
+{
+  if constexpr (PACK_SIZE) {
+    *reinterpret_cast<uint32_t*>(target) = value.second;
+    target += sizeof(uint32_t);
+  }
+
+  AccessorWO<int8_t, 1> acc(value.first, value.second, false);
```
manopapad, line 173: AccessorRO?

src/core/task/return.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/353/commits/7a62c098103d6e1e5c1cedf9593e1e274df4dd8b

src/core/task/return.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/353/commits/7a62c098103d6e1e5c1cedf9593e1e274df4dd8b

manopapad (approved):

src/core/data/store.cc
magnatelee: It doesn't work because `p_init_value` points to a framebuffer allocation whereas the runtime expects it to be a host allocation. We could potentially add another constructor that takes the initial value in a future and make the runtime issue a realm copy between the backing instances. @lightsighter what do you think?

PR #350, opened 2022-08-26
Instance manager improvements

magnatelee:
Mappers used separate instance managers, which prevented them from reusing instances created by others. This PR fixes that problem simply by sharing the same instance manager with all Legate mappers.

Update: this PR has been updated to contain changes from #351 and other improvements to the instance manager.

PR #355, opened 2022-09-01
Use Legion primitives to coordinate accesses to the shared instance manager

magnatelee:
#350 used `std::mutex` to synchronize accesses to the shared instance manager, which turns out to be problematic as the blocking behavior is invisible to the runtime and eventually leads to hangs. This PR replaces that mutex with Legion's locking mechanisms that are compatible with the runtime's thread management.

### DIFF
--- a/src/core/mapping/instance_manager.h
+++ b/src/core/mapping/instance_manager.h
@@ -152,8 +152,7 @@ class InstanceManager {
   void erase(Instance inst);
 
  public:
-  void lock();
-  void unlock();
+  Legion::Mapping::LocalLock& lock() { return manager_lock_; }
 
  public:
   static InstanceManager* get_instance_manager();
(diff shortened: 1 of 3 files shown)


src/core/mapping/instance_manager.h
```diff
@@ -152,8 +152,7 @@ class InstanceManager {
   void erase(Instance inst);
 
  public:
-  void lock();
-  void unlock();
+  Legion::Mapping::LocalLock& lock() { return manager_lock_; }
```
rohany, line 155: I wouldn't name this `lock()`, as then `instance_manager->lock()` implies that a lock is being taken (and could lead to a bug in the future). `get_lock()` or something similar would be better.

rohany:
I have confirmed that this gets rid of my hangs.

src/core/mapping/instance_manager.h
magnatelee: yeah, `lock()` seems ambiguous as it's unclear whether it's holding the lock or simply returning it. I renamed it with `manager_lock()`. FYI, I reserved `get_*` names for getters that involve some computation and return fresh values and not references.

PR #358, opened 2022-09-02
Synchronize instance manager accesses from mapper calls other than `map_task`

magnatelee:


manopapad:
I am seeing segfaults on erase as well, that also needs to be covered?

magnatelee:
> I am seeing segfaults on erase as well, that also needs to be covered?

should be covered as of https://github.com/nv-legate/legate.core/pull/358/commits/92fa673ada9fc13a1625eeb27424c572471585d3

rohany:
I can confirm that this fixes crashes that I have been seeing as well.

PR #361, opened 2022-09-07
Cycle detection check

manopapad:
Uploading the script I used to find the most recent case of reference cycles in the legate.core object graph. Here is how I used it:

```
import cunumeric as cn
from legate.core.cycle_detector import print_cycles
a = cn.arange(10000) % 100
b = cn.arange(10000) // 100
c = cn.random.rand(100, 100)
c[a, b]
del c
del b
del a
print_cycles()
```

And here is the output:

```
looking for cycles involving 0x7f79a8043ee0, of type <class 'legate.core.store.RegionField'>
found cycle!
  tail:
    0x7f79a8043ee0: <class 'legate.core.store.RegionField'>
     ^ ["_data"]
    0x7f7a058cca00: <class 'dict'>
     ^ .__dict__
    0x7f79a80509a0: <class 'legate.core.store.Storage'>
     ^ ["_storage"]
    0x7f7a05885200: <class 'dict'>
     ^ .__dict__
    0x7f79a8050760: <class 'legate.core.store.Store'>
     ^ [0]
    0x7f79975703c0: <class 'tuple'>
     ^
    0x7f79e8bd5a30: <class 'frame'>
  cycle:
    0x7f79e8bd5a30: <class 'frame'>
     ^ .tb_frame
    0x7f7a0592ad80: <class 'traceback'>
     ^ .tb_next
    0x7f7a0592a400: <class 'traceback'>
     ^ .tb_next
    0x7f79aa1921c0: <class 'traceback'>
     ^ .tb_next
    0x7f79aa192f80: <class 'traceback'>
     ^ .tb_next
    0x7f79aa1922c0: <class 'traceback'>
     ^ .tb_next
    0x7f79aa192340: <class 'traceback'>
     ^ .tb_next
    0x7f79aa192940: <class 'traceback'>
     ^ .tb_next
    0x7f79aa192100: <class 'traceback'>
     ^ .tb_next
    0x7f79aa192840: <class 'traceback'>
     ^ ["_tb"]
    0x7f7a054e3300: <class 'dict'>
     ^ .__dict__
    0x7f79a8043bb0: <class 'legate.core.operation.AutoTask'>
     ^
    0x7f79e8bd5a30: <class 'frame'>
[4 more similar cycles after this]
```

Please feel free to suggest minor or significant changes, or a different tool or approach. What I want is to incorporate this check within our CI system ASAP, so we never have to deal with this kind of regression again.

I expect that we will want to have a special mode in CI, where we run with the minimal amount of resources, disable GC, skip all of our explicit `gc.collect()` calls in legate.core and cunumeric, and do one cycle check at application exit.


lightsighter:
Reference cycle detection, there are other (worse) kinds of cycles ;)

manopapad:
@bryevdv @marcinz I added a commandline flag, to perform cycle check at exit. You can see it at work by doing:

```
LEGATE_PRECISE_EXCEPTION_TRACE=1 legate a.py -legate:cycle-check
```

where `a.py` is:

```
import cunumeric as cn
a = cn.arange(10000) % 100
b = cn.arange(10000) // 100
c = cn.random.rand(100, 100)
c[a, b]
```

Because the check is performed inside a finalizer, I cannot `sys.exit(1)` to signal failure. Can we still add this check to the CI, even though the program appears to exit normally even if cycles are found? If not, I could cache the cycle check result and act on it later inside `_cleanup_legate_runtime`.

bryevdv:
@manopapad I think you'd have to capture all the output and look for `found cycle!` or whatever marker signifies that

manopapad:
@bryevdv What is the best way to unit-test this? E.g. I would want to verify that the following program:

```
from legate.core import get_legate_runtime, types as ty
runtime = get_legate_runtime()
context = runtime.core_context
store = context.create_store(ty.int64, shape=(4, 4))
x = [store]
x.append(x)
```

when run as follows:

```
legate a.py -legate:cycle-check
```

produces output like the following:

```
looking for cycles involving 0x7f8bb0768f40, of type <class 'legate.core.store.Store'>
found cycle!
  tail:
    0x7f8bb0768f40: <class 'legate.core.store.Store'>
     ^ [0]
    0x7f8bb2c5cfc0: <class 'list'>
  cycle:
    0x7f8bb2c5cfc0: <class 'list'>
     ^ [1]
    0x7f8bb2c5cfc0: <class 'list'>
```

I don't think the current unit testing setup allows you to test the entire output of `legate`.

bryevdv:
@manopapad I think the current test driver has the ability to add per-test flags, what about running that code in a test that adds that flag, and then you could use the pytest `capsys` fixture to capture and examine stdout.

manopapad:
Wrote a unit test, @bryevdv can you take a look?

bryevdv:
Test looks good!

bryevdv (approved):

PR #368, opened 2022-09-20
InstanceManager segfault fixes

manopapad:
Fixes various segfaults in the InstanceManager, and strengthens the sanity checks.

The subsumption checks are no longer used, and can be removed if there is no further need for them.

### DIFF
--- a/src/core/mapping/instance_manager.cc
+++ b/src/core/mapping/instance_manager.cc
@@ -148,8 +148,8 @@ struct construct_overlapping_region_group_fn {
 
       // NOTE: It is critical that we maintain the invariant that if at least one region is mapped
       // to a group in the instances_ table, that group is still present on the groups_ table, and
-      // thus there's at least one shared_ptr remaining to it. Otherwise we run the risk that a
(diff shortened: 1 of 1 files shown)


src/core/mapping/instance_manager.cc
```diff
@@ -241,8 +260,21 @@ bool InstanceSet::erase(PhysicalInstance inst)
       it++;
   }
 
-  for (auto& group : filtered_groups)
-    for (auto& region : group->regions) groups_.erase(region);
+  std::vector<Region> filtered_regions;
```
magnatelee, line 263: I think we want to use a `std::set` to keep the filtered regions, as some filtered groups might have an intersection. (I believe calling `erase` with a non-existent key is safe but not efficient.)

magnatelee:
@manopapad left a minor comment. otherwise looks good.

src/core/mapping/instance_manager.cc
manopapad: 2c0cfb7c26a73c386d777083b46f779679b21a9a

magnatelee (approved):

PR #369, opened 2022-09-20
Add Fill operation

manopapad:
Single Fills only for now; not sure if IndexFills are necessary.


legate/core/launcher.py
```diff
@@ -1111,3 +1112,59 @@ def execute(
...
+        self._point = point
+
+    def build_single_fill(self) -> Fill:
+        assert self._lhs.kind is not Future
+        assert self._value.kind is Future
+        if TYPE_CHECKING:
+            assert isinstance(self._lhs.storage, RegionField)
+            assert isinstance(self._value.storage, Future)
+        fill = Fill(
+            self._lhs.storage.region,
```
manopapad, line 1155: This assumes that the `Storage` attached to a `Store` is always an exact match for the extent which that `Store` represents (i.e. any attached transforms that restrict the extent of the region have already been applied through partition operations on the attached `Storage`). I suspect this is true, but it's not trivially obvious from the code.

magnatelee:
> Single Fills only for now; not sure if IndexFills are necessary.

Mike reminded that a single fill on a root region would coalesce the equivalence set, which later should be refined. So we should support index fills as well and I don't think it would be terribly difficult to add.

magnatelee:
@manopapad can you add provenance tracking for fills? https://github.com/nv-legate/legate.core/commit/44cc9418d17d2d6222bbc05706cd5cb21ed5d4c6

manopapad:
> @manopapad can you add provenance tracking for fills? https://github.com/nv-legate/legate.core/commit/44cc9418d17d2d6222bbc05706cd5cb21ed5d4c6

12c9b2c

magnatelee (approved):

PR #370, opened 2022-09-21
Provenance tracking

magnatelee:
This PR adds provenance tracking for tasks, copies, and fills.


legate/core/context.py
```diff
@@ -38,6 +48,24 @@
 T = TypeVar("T")
 
 
+class AnyCallable(Protocol):
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
```
bryevdv, line 52: I thought this already existed somewhere but maybe only in `cunumeric`?

bryevdv (approved):

PR #376, opened 2022-09-21
Add suggested fixes for known build issues

jjwilke:


jjwilke:
This needs to be changed to both capture output and continue to pipe output to the terminal. I don't recall an easy way to do this with Python.

install.py
```diff
@@ -24,11 +24,33 @@
...
 
 # Flush output on newlines
 sys.stdout.reconfigure(line_buffering=True)
 
 
+class bcolors:
+    OKGREEN = "\033[92m"
+    WARNING = "\033[93m"
+    FAIL = "\033[91m"
+    ENDC = "\033[0m"
```
bryevdv, line 37: Could imitate the API I've used elsewhere for consistency

https://github.com/nv-legate/legate.core/blob/5a73122b7516f6b8e786d0c649522223aedc20fe/legate/driver/ui.py#L67-L68

(just manually, w/o using `colorama`)

bryevdv:
@jjwilke it seems non-trivial enought that there is an entire package for doing it https://github.com/pycontribs/subprocess-tee

Maybe we could simplify it and vendor the parts we need. Otherwise, you can capture stdout and then just print it back out yourself, but obviously that won't update the console "in real time"

bryevdv (dismissed):

Issue #374, opened 2022-09-21
Check if ninja is available, fall back to gmake if not


manopapad:

jjwilke:
I'm just going to `shutil.which("ninja")` and set the default to `None` if ninja is not available.

manopapad:
Fixed by #379

PR #379, opened 2022-09-22
Only set default CMake generator if Ninja is available: Issue #374

jjwilke:


install.py
```diff
@@ -439,6 +439,12 @@ def validate_path(path):
 
 
 def driver():
+    ninja_path = shutil.which("ninja")
+    if ninja_path is None:
```
bryevdv, line 443: could use walrus here

```suggestion
    if (ninja_path := shutil.which("ninja")) is None:
```

bryevdv (approved):

install.py
bryevdv: or alternatively, a ternary:
```
ninja_path = shutil.which("ninja")
cmake_generator_default = None if ninja_path is None else "Ninja
```

manopapad (approved):

manopapad:
I went ahead and applied the PR review comments; it would be good to merge this soon, since our default environment files are changing to not mandate ninja.

PR #381, opened 2022-09-22
Delay label checking on fresh PRs

marcinz:
When a PR is created, the authors will often set the labels soon after creation. We will have a "grace period" on a new pull request to allow setting labels after the PR is created.


marcinz:
@magnatelee Delay of 60s looks good or should we have more?

PR #384, opened 2022-09-23
Support building with GASNet-Ex and MPI backends

manopapad:
Port of https://github.com/nv-legate/legate.core/pull/322 for the new build system


manopapad:
@jjwilke this doesn't seem to be breaking anything, so I'm going to merge so we unblock the performance runs, but please do take a look and make sure I'm doing something silly

PR #383, opened 2022-09-23
Add errors when find_package(legion) and CPM_Legion_SOURCE conflict

jjwilke:

PR #387, opened 2022-09-25
Allow only one of --legion-dir and --legion-src-dir

jjwilke:


jjwilke:
Compare with PR #383. Either PR #383 or this PR should get merged.

manopapad:
I think this is the better PR of the two. Can we also add a check in `install.py` that the user shouldn't pass both `legion_dir` (which IIUC has the effect of redirecting the `findpackage(legion)` search to a user-defined directory) and `legion_src_dir`?

jjwilke:
@manopapad Added check

manopapad (approved):

manopapad:
@trxcllnt how does this interact with https://github.com/nv-legate/legate.core/pull/411 ?

trxcllnt (approved):

trxcllnt (changes requested):
I approved thinking this was a new PR, but I think this change was also made in #411. So maybe we can close this instead?

manopapad:
I removed the changes that were covered by https://github.com/nv-legate/legate.core/pull/411, but kept the sanity check in `install.py`.

PR #392, opened 2022-09-28
Add a pool of region managers with LRU eviction

magnatelee:
One side effect of PR #364 is that it eagerly collects region managers even when their fields can be reused immediately after their active field counts become zero. This makes the code create many fresh regions, most of which aren't necessary if the code was able to reuse managers with no active fields. This is problematic as the runtime doesn't perform region destruction as frequently as those fresh regions are created, which ends up limiting how far the runtime can run ahead.

To solve this problem, this PR introduces a pool of inactive region managers with some simple LRU eviction policy.


legate/core/runtime.py
```diff
@@ -1217,19 +1245,39 @@ def find_region_manager(self, region: Region) -> RegionManager:
...
         return self.region_managers_by_region[region]
 
-    def destroy_region_manager(
+    def revive_manager(self, region_mgr: RegionManager) -> None:
+        lru_managers: Deque[RegionManager] = deque()
+        for to_check in self.lru_managers:
+            if to_check is not region_mgr:
+                lru_managers.append(to_check)
+        assert len(lru_managers) < len(self.lru_managers)
+        self.lru_managers = lru_managers
```
manopapad, line 1254: Out of curiosity, did this not work?
```
self.lru_managers.remove(region_mgr)
```

magnatelee: that would have worked. (the original code this was copied from needed a loop, but then it has become unnecessary after the code went through some simplification.) will make a follow-on PR to clean this up after the release.

PR #395, opened 2022-09-28
Periodically flush region managers in the LRU queue

magnatelee:
Though PR #392 addresses the inefficiency from collecting inactive region managers too eagerly, it may let those inactive managers stick around until the program finishes. This PR introduces another heuristic that periodically flushes inactive managers kept in the LRU queue.

### DIFF
--- a/legate/core/runtime.py
+++ b/legate/core/runtime.py
@@ -825,6 +825,8 @@ class CommunicatorManager:
 class Runtime:
     _legion_runtime: Union[legion.legion_runtime_t, None]
     _legion_context: Union[legion.legion_context_t, None]
+    # TODO: do we want this to be configurable?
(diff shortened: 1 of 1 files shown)


manopapad (approved):
This is OK, but I would have preferred to see a separate counter on each "dead" RegionManager. Then each RegionManager would be guaranteed to last for N field allocations, whereas now it could get unlucky and be removed right away, because it happened to become empty right before an LRU cache flush was scheduled to be triggered.

I think such a feature would actually remove the need for an LRU mechanism altogether; just place dead RegionManagers in a regular list and tick them all up on each field allocation.

marcinz:
@magnatelee What is the status of this PR?

magnatelee:
@marcinz this PR was a bit rushed and I'll make a follow-on PR that subsumes this. let's keep this open until that happens.

PR #409, opened 2022-10-04
Legion bug WAR: don't instantiate futures on framebuffer

manopapad:


src/core/data/store.cc
```diff
@@ -136,8 +136,9 @@ FutureWrapper::FutureWrapper(
 #ifdef DEBUG_LEGATE
     assert(!initialize || future_.get_untyped_size() == field_size);
 #endif
-    auto proc     = Processor::get_executing_processor();
-    auto mem_kind = proc.kind() == Processor::Kind::TOC_PROC ? Memory::Kind::GPU_FB_MEM
+    auto proc = Processor::get_executing_processor();
+    // TODO: 2022-10-04: Work around a Legion bug, by not instantiating futures on framebuffer.
+    auto mem_kind = proc.kind() == Processor::Kind::TOC_PROC ? Memory::Kind::Z_COPY_MEM
```
magnatelee, line 141: you can use this function: https://github.com/nv-legate/legate.core/blob/branch-22.12/src/core/utilities/machine.cc#L23-L28

magnatelee:
@manopapad can we wrap this workaround with a C macro?

manopapad:
> @manopapad can we wrap this workaround with a C macro? we might want to be able to toggle the behavior even after the runtime fix.

Done.

src/core/mapping/base_mapper.cc
```diff
@@ -610,7 +610,11 @@ void BaseMapper::map_task(const MapperContext ctx,
 
     if (req_indices.empty()) {
       // This is a mapping for futures
-      output.future_locations.push_back(get_target_memory(task.target_proc, mapping.policy.target));
+      StoreTarget target = mapping.policy.target;
+#ifdef LEGATE_NO_FUTURES_ON_FB
```
manopapad, line 614: Note that this change is not strictly necessary to work around the assertion failure we're seeing currently. The creation of Future buffers on fbmem in `store.cc` is what's triggering it. However, since we know there's a general problem with Legion's handling of Futures instances on fbmem, I removed this case as well.

src/core/data/store.cc
manopapad: fixed

magnatelee (approved):

magnatelee:
@manopapad please port this pr to `branch-22.12` so that we can do perf testing while the runtime fix is being worked on

PR #411, opened 2022-10-04
Explicitly build Legion if `legion_dir` or `legion_src_dir` is not provided

trxcllnt:
This PR fixes an issue where building and installing `legate.core` succeeds, but re-building and installing `legate.core` again (without uninstalling it first) seemingly succeeds but removes the Legion libs.

Here's the output from two runs:
* [Build and install into a clean conda env](https://gist.github.com/trxcllnt/2c96a589a5eea79f71eb74184624055a#file-gistfile1-txt-L4812)
* [Build and install into the same conda env again (w/o uninstalling legate.core)](https://gist.github.com/trxcllnt/fc123699bd04847708fd1019d4a8e985#file-gistfile1-txt-L3835)

1. The first run builds and installs Legion as part of building the `legate.core` wheel.
2. The second run _doesn't_ build Legion (because the installed version from before [was found](https://gist.github.com/trxcllnt/fc123699bd04847708fd1019d4a8e985#file-gistfile1-txt-L189) by CMake)

As part of installing the wheel produced by step 2, pip first [removes the files](https://gist.github.com/trxcllnt/fc123699bd04847708fd1019d4a8e985#file-gistfile1-txt-L3811-L3817) added by the previous installation (step 1), but it doesn't re-install Legion (because Legion was found in an install location and not built as a side-effect of building `legate.core`.)

This PR forces `install.py` to tell CMake to always build Legion as part of building `legate.core` _unless_ the user explicitly passes `--with-legion` or `--legion-src-dir`.


bryevdv:
This seems good to me

bryevdv (approved):

bryevdv:
@trxcllnt I pushed the button to merge in the mypy fixes, should go green now :D

bryevdv:
merge ok'd by @trxcllnt

PR #412, opened 2022-10-04
Parallel task execution

ipdemes:
Adding some logic for asynchronous execution of the single tasks in Legate.
The option can be enabled by the `-legate:parallel_execution` flag


legate/core/runtime.py
```diff
@@ -69,6 +69,87 @@
...
+            )
+            result = IndexSpace(
+                self._legion_context, self._legion_runtime, handle=handle
+            )
+            return result
+        else:
+            return None
+
+
+class DeviceID:
```
ipdemes, line 142: @magnatelee I could have used `int` instead of DeviceID, if you think it would be better. I thought it would be nice to have a separate class for it so code is easier to understand/read.

magnatelee: I feel a type alias would be sufficient if readability was the goal. and if we mean to add an abstraction for devices, I want it to contain more metadata than just the id to be useful. (for example, #439 might need such an abstraction.) for this PR, I don't think that abstraction would be necessary though.

legate/core/runtime.py
```diff
@@ -1038,6 +1138,16 @@ def num_omps(self) -> int:
...
         return self._num_gpus
 
+    @property
+    def num_nodes(self) -> int:
+        return self._num_nodes
+
+    def machine_model(self) -> MachineModel:
+        return self._machine_model
+
+    def get_next_available_point(self) -> DeviceID:
```
magnatelee, line 1148: maybe we shouldn't add this shortcut but let the caller go through the machine model instead.

legate/core/store.py
```diff
@@ -106,6 +107,7 @@ def __init__(
         field: Field,
         shape: Shape,
         parent: Optional[RegionField] = None,
+        location: Optional[DeviceID] = None,
```
magnatelee, line 110: I find `Store` to be a more natural place to keep the location.

magnatelee (changes requested):

PR #430, opened 2022-10-11
remove --install-dir option

bryevdv:
This PR removes the `--install-dir` option from `./install.py` per previous team discussions.


manopapad:
Would it make sense to also remove the `--python-lib` option at this point? Since we only support installing under the currently active python environment, that would dictate which python library we're building against.

bryevdv:
That seems reasonable to me, @trxcllnt are there any reasons not to that we are not aware?

trxcllnt:
@bryevdv no, that sounds reasonable to me

trxcllnt (approved):

bryevdv:
@manopapad removed in https://github.com/nv-legate/legate.core/pull/430/commits/c7f4bc65e573b45d0979a04e085d76b38ccf2c24

Issue #422, opened 2022-10-07
User-controlled target processor kind


manopapad:
@rohany brought up a scenario where the user would want to constrain one phase of the application to run on CPUs, and then switch to GPUs.

My first hunch would be to allow the user to annotate parts of the code with a processor target, e.g. using a python context:

```
with legate.execute_on_cpus():
  x = make_input()
with legate.execute_on_gpus():
  y = process(x)
```

But maybe Rohan can explain his usecase a bit more, and we can see if there's a better way to support it.

rohany:
As manolis sketched, I have an application which has a data generation phase that is very memory hungry (especially for deferred allocations) and then a phase that has pretty stable memory usage (a solve). I would like to perform the data generation (and allocate more deferred memory) on the CPUs which have more memory, and then flip a switch to have the solve run on the GPUs. This seems like something that could be controlled by legate core, as individual libraries can control this pretty well but cannot control what other legate libraries will do.

magnatelee:
for everyone's benefit, here's a snippet that shows a tentative API design:
<img width="1354" alt="all devices = get machine()" src="https://user-images.githubusercontent.com/331175/197047561-926ba9f7-ef0e-47df-8fd4-3222a2912b7e.png">

though the API looks simple, the resource scoping has major consequences on how the core internally manages cached states (as #439 already explored some of it). To properly address them, a new scoping mechanism would be introduced to centralize all cached states sensitive to the resource configuration, such as the max degree of parallelism, key partitions, communicators, etc.

RAMitchell:
My 2 cents. Out of memory errors are our most significant barrier to running distributed XGBoost. We can tell users to launch more machines but we do want the utilisation of each machine to be high and this is more often limited by memory.

Allowing user-controlled target processors would be nice, but it would be even nicer if legate could automatically recognise that a task will exceed available GPU memory and run it on a CPU if there is more memory available. Maybe this is asking too much given that things like unbound store sizes are not known ahead of time?

What if I were able to label a task somehow as "memory bound", indicating that legate should do everything possible to avoid OOM.

magnatelee:
> Allowing user-controlled target processors would be nice, but it would be even nicer if legate could automatically recognise that a task will exceed available GPU memory and run it on a CPU if there is more memory available. Maybe this is asking too much given that things like unbound store sizes are not known ahead of time?

[...]

PR #439, opened 2022-10-14
legate/core: rough prototype of staged control of CPU-GPU execution

rohany:
Signed-off-by: Rohan Yadav <rohany@alumni.cmu.edu>

This is a rough prototype of some of the implementation behind https://github.com/nv-legate/legate.core/issues/422. An accompanying commit to change in cunumeric to get this to work out is https://github.com/rohany/cunumeric/commit/bed2f6a2dc41ae57afd1be61ff481bcbfa165411. The API can be improved, but I'd like to get feedback on how this is done and if there are any interactions with other parts in the codebase that I should be worried about. 

Additionally, I understand if NVIDIA is planning on doing something like this differently, or has a broader vision in mind to get this done. However, I'd like to get feedback on this approach regardless, as I'll integrate this into the branches I maintain for my own experiments.

### DIFF
--- a/legate/core/operation.py
+++ b/legate/core/operation.py
@@ -142,6 +142,19 @@ class Operation(OperationProtocol):
         self._all_parts: list[PartSym] = []
         self._launch_domain: Union[Rect, None] = None
         self._error_on_interference = True
+        # If we've been requested to limit execution to GPUs,
+        # set the corresponding tag on launched tasks.
+        # TODO (rohany): For some reason, I can't do this on
+        #  the Task parent class as the context isn't defined
+        #  yet. I don't understand this error...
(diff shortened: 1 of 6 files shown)


magnatelee:
@rohany thanks for taking a shot at this. I feel this PR can benefit from the machine model that #412 is about to add to the core, so I think it's not a bad idea to let #412 go first. also, I think the code should directly pass the target processor kind for a task to the mapper rather than relying on the mapper's default heuristic and occasionally making it deviate from it. that would align better with the current plan that the core runtime chooses task mappings and the mapper merely replays decisions (though it'd still do some of the instance management).

rohany:
Closing in favor of https://github.com/nv-legate/legate.core/pull/457

PR #442, opened 2022-10-17
Fix returned legion paths for editable install with separate legion b…

jjwilke:
When using a separate legion build with editable installs, this falls through to return presumed installed paths. This fixes? the logic by sending all Legion build tree configurations through a `finally` block.


bryevdv (approved):

bryevdv:
Checked this works with a "plain" editable install that I normally do
```
./install.py --verbose --openmp --cuda --arch turing --clean --editable
```

trxcllnt (approved):

PR #449, opened 2022-10-20
Make sure the buffer builder stays alive until the task is launched

magnatelee:


manopapad:
I don't see why this change would have the desired effect.

The `BufferBuilder` you pass to `build_task` doesn't need to be kept around after the `IndexTask` object is created, we just need to keep around what comes out of `argbuf.get_string()`, which is a fresh `bytes` object returned by `struct.pack()` (a normal Python object, subject to normal garbage collection rules), and `argbuf.get_size()`, which is an `int`. I would expect that it's the lifetime of those objects that must be managed, not the lifetime of the `BufferBuilder`, and `IndexTask` explicitly keeps around a pointer to its `data` buffer, which should keep `ffi.from_buffer(data)` valid https://github.com/nv-legate/legate.core/blob/branch-22.12/legate/core/_legion/task.py#L535.

Overall, it smells to me like possibly the `task` object is collected before `task.launcher` is executed, and thus the `task.data` pointer is not around to prevent collection of the data pointed to by the launcher.

magnatelee:
@manopapad I didn't see that we've been keeping that reference in the launcher (I don't recall since when we started doing that). you're right that this PR doesn't fix anything with that extra reference.

PR #460, opened 2022-10-31
Python optimization experiments

bryevdv:
cc @manopapad 

I tried lots of things, many of them did not have an effect that I could discern. Here are a few that seemed to make a difference in the Python cProfile output (i.e. something that was near the top of `tottime` dropped off after the change). It's hard to say whether there was any appreciable change in the legion prof output. With GPUs especially, Python proc still always seemed pegged at 100% on my machine.

I did also look at the example code itself. Just FYI, factoring out the three values at the top here seems to reliably shave off 10% of the execution time reported by legion prof
```python
def pressure_poisson_periodic(b, nit, p, dx, dy):
    pn = np.empty_like(p)

    # these were pulled to the top
    dx2 = dx ** 2
    dy2 = dy ** 2
    dxdynorm = (2 * (dx2 + dy2)) - dx2 * dy2 / (2 * (dx2 + dy2))

    for q in range(nit):
        pn = p.copy()
        p[1:-1, 1:-1] = (
            (pn[1:-1, 2:] + pn[1:-1, 0:-2]) * dy2
            + (pn[2:, 1:-1] + pn[0:-2, 1:-1]) * dx2
        ) / dxdynorm * b[1:-1, 1:-1]

        # Periodic BC Pressure @ x = 2
        p[1:-1, -1] = (
            (pn[1:-1, 0] + pn[1:-1, -2]) * dy2
            + (pn[2:, -1] + pn[0:-2, -1]) * dx2
        ) / dxdynorm * b[1:-1, -1]

        # Periodic BC Pressure @ x = 0
        p[1:-1, 0] = (
            (pn[1:-1, 1] + pn[1:-1, -1]) * dy2
            + (pn[2:, 0] + pn[0:-2, 0]) * dx2
        ) / dxdynorm * b[1:-1, 0]

        # Wall boundary conditions, pressure
        p[-1, :] = p[-2, :]  # dp/dy = 0 at y = 2
        p[0, :] = p[1, :]  # dp/dy = 0 at y = 0

    return p
```


legate/core/shape.py
```diff
@@ -148,55 +144,95 @@ def __eq__(self, other: object) -> bool:
             else:
                 return self.extents == other.extents
         elif isinstance(other, (int, Iterable)):
-            lh = _cast_tuple(self, self.ndim)
-            rh = _cast_tuple(other, self.ndim)
+            lh = self.extents
+            rh = (
```
manopapad, line 148: Can we roll the `other.extents if isinstance(other, Shape) else ...` part into `_cast_tuple` as well?

manopapad (approved):
This has indeed made a difference in our performance runs (Python execution is no longer the bottleneck).

legate/core/shape.py
bryevdv: That's the way it was, I was trying to avoid calling `_cast_tuple` at all when possible

bryevdv:
Good to hear! That was not evident to me on my local machine (or else by my interpretation of legion-prof)

PR #461, opened 2022-11-01
Call bind.sh unconditionally

bryevdv:
This PR changes the driver invocation to always add `bind.sh` unconditionally. This is so that a computed `LEGATE_RANK` environment variable can be set consistently, regardless of launch presence or kind. This will support a future PR that adds an option for per-rank Python cprofile output.

The new usage for `bind.sh` is:
```
bind.sh --launcher <launcher> [--cpus <spec>] [--gpus <spec>] [--mems <spec>] [--nics <spec>] -- <app> ...
```
The main differences are:
* `--launcher` required before the launcher value (`local`, `mpirun`, `srun`, or `jsrun`)
* delimeter `--` required before remaining "app" portion

These changes make the reported invocation less confusing and more robust, and should be fine since `bind.sh` is only invoked by the driver, and the driver has complete control over the args passed.  The necessary changes have been made to the driver (and related unit tests)

### DIFF
--- a/bind.sh
+++ b/bind.sh
@@ -13,18 +13,40 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# Usage: bind.sh --launcher <launcher> [--cpus <spec>] [--gpus <spec>] [--mems <spec>] [--nics <spec>] -- <app> ...
-# <spec> specifies the resources to bind each node-local rank to, with ranks
-# separated by /, e.g. 0,1/2,3/4,5/6,7 for 4 ranks per node.
+#
 
 set -euo pipefail
 
 help() {
-  echo "Usage: bind.sh -l | --launcher <mpirun,srun,jrun,local> [ -c | --cpus ] [ -g | --gpus ] [ -m | --mems ] [ -n | --nics ] -- <app>"
+  cat <<EOM
+Usage: bind.sh [OPTIONS]... -- APP...
+
+Options:
+  --launcher={mpirun|srun|jrun|auto|local}
+                    Launcher type, used to set LEGATE_RANK
+                    If 'auto', attempt to find the launcher rank automatically
+                    If 'local', rank is set to "0".
+  --cpus=SPEC       CPU binding specification, passed to numactl
+  --gpus=SPEC       GPU binding specification, used to set CUDA_VISIBLE_DEVICES
+  --mems=SPEC       Memory binding specification, passed to numactl
+  --nics=SPEC       Network interface binding specification, used to set
+                    all of: UCX_NET_DEVICES, NCCL_IB_HCA, GASNET_NUM_QPS,
+                    and GASNET_IBV_PORTS
+
+SPEC specifies the resources to bind each node-local rank to, with ranks
+separated by /, e.g. '0,1/2,3/4,5/6,7' for 4 ranks per node.
+
+APP is the application that will be executed by bind.sh, as well as any
+arguments for it.
+
+If --cpus or --mems is specified, then APP will be invoked with numactl.
+
+An explicit '--' separator should always come after OPTIONS and before APP.
+EOM
   exit 2
 }
 
+launcher=auto
 while :
 do
   case "$1" in
@@ -46,20 +68,22 @@ do
   shift 2
 done
 
-if [ -z "$launcher" ]; then
-  echo "bind.sh: -l / --launcher is a required argument"
-fi
-
 case "$launcher" in
-  local) rank=0 ;;
-  mpirun) rank="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
-  jsrun) rank="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
-  srun) rank="$SLURM_LOCALID" ;;
+  mpirun) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
+  jsrun ) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
+  srun  ) rank="${SLURM_LOCALID:-unknown}" ;;
+  auto  ) rank="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-unknown}}}" ;;
+  local ) rank="0" ;;
   *)
     echo "Unexpected launcher value: $launcher"
     help
     ;;
 esac
+
+if [[ "$rank" == "unknown" ]]; then
+    echo "Error: Could not determine node-local rank" 1>&2
+    exit 1
+fi
 
 export LEGATE_RANK="$rank"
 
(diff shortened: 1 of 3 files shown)


bind.sh
```diff
@@ -13,92 +13,110 @@
...
-    if [[ -n "${CPUS+x}" ]]; then
-        set -- --physcpubind "${CPUS[$IDX]}" "$@"
-    fi
-    if [[ -n "${MEMS+x}" ]]; then
-        set -- --membind "${MEMS[$IDX]}" "$@"
-    fi
-    set -- numactl "$@"
-elif [[ -n "${CPUS+x}" || -n "${MEMS+x}" ]]; then
-    echo "Warning: numactl is not available, cannot bind to cores or memories" 1>&2
+if [ $launcher != "local" ]; then
```
bryevdv, line 108: @manopapad please confirm, when launcher is "local", all of this should be skipped.

bind.sh
```diff
@@ -13,92 +13,110 @@
...
+    echo "Unexpected launcher value: $launcher"
+    help
+    ;;
 esac
-shift
-if [[ "$IDX" == "none" ]]; then
-    echo "Error: Cannot detect node-local rank" 1>&2
-    exit 1
+
+export LEGATE_RANK="$rank"
```
bryevdv, line 64: Is this env var name OK, or is there something better?

bryevdv: Also: do we want to also ultimately change the `nsys` and `nvprof` invocations to utilize this env var instead of the `%q` special case?

bind.sh
manopapad: We would like the binding to still happen, even if it's just a singleton invocation. As a matter of fact, `test.py` depends on this behavior.

bind.sh
```diff
@@ -13,92 +13,110 @@
...
-case "$1" in
-    mpirun) IDX="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
-    jsrun) IDX="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
-    srun) IDX="$SLURM_LOCALID" ;;
-    local) IDX=0 ;;
-    none) IDX="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-none}}}" ;;
+set -euo pipefail
+
+help() {
+  echo "Usage: bind.sh -l | --launcher <mpirun,srun,jrun,local> [ -c | --cpus ] [ -g | --gpus ] [ -m | --mems ] [ -n | --nics ] -- <app>"
```
manopapad, line 24: Add `<spec>` after `-c | --cpus`, `-g | --gpus` etc.

manopapad:
Can we print all error-like messages to stderr? Including the help message? Just add `1>&2` at the end of the `echo` command.

bind.sh
```diff
@@ -13,92 +13,110 @@
...
+
+if [ -z "$launcher" ]; then
+  echo "bind.sh: -l / --launcher is a required argument"
+fi
+
+case "$launcher" in
+  local) rank=0 ;;
+  mpirun) rank="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
+  jsrun) rank="$OMPI_COMM_WORLD_LOCAL_RANK" ;;
+  srun) rank="$SLURM_LOCALID" ;;
```
manopapad, line 57: We still want to have the "catch-all" case for `launcher == none`. This is what would happen if legate is invoked with external launching, i.e. the user calls `mpirun -N 3 legate` instead of `legate --launcher mpirun`. In that case there will be one instance of `legate` invoked on each process, and we have no idea what launcher was used to launch us, so we just need to check every variable we know about.

bind.sh
bryevdv: Right, I guess I was thinking of this case:
```
    if all(x is None for x in (cpu_bind, mem_bind, gpu_bind, nic_bind)):
        return ()
``` 
which suppressed `bind.sh` at all if no bind args were set. We could:

* detect this inside `bind.sh` and skip the `numactl` invocation
* detect this in the driver and pass `--launcher=none` to flag skipping `numactl`

Do you have a preference?

bind.sh
manopapad: The name is good. Yes, we can handle nsys and nvprof in the same script as cprofile, and use the same variable to name their output files.

bind.sh
bryevdv: :+1:  OK we will need to remove the check above to make it required

bind.sh
manopapad: `numactl` should always be used if `--cpu-bind` or `--mem-bind` is present (and `numactl` is available on the system), regardless of the value of `--launcher`. So I think we need to check for the presence of either of these options inside `bind.sh`.

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/461/commits/cb05f269f6a078b470b319e448f3e558d261aa64

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/461/commits/cb05f269f6a078b470b319e448f3e558d261aa64

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/461/commits/cb05f269f6a078b470b319e448f3e558d261aa64

legate/driver/command.py
```diff
@@ -30,18 +30,11 @@
...
-        return ()
-
     ranks = config.multi_node.ranks
 
     opts: CommandPart = (
         str(system.legate_paths.bind_sh_path),
+        "--launcher",
         "local"
         if launcher.kind == "none" and ranks == 1
         else str(launcher.kind),
```
manopapad, line 40: Note that this can pass "--launcher none", if we're externally launched on multiple ranks/nodes. It doesn't look like this is currently handled in `bind.sh` (it should do what is currently being done in the case of "auto").

bind.sh
```diff
@@ -17,88 +17,131 @@
...
+if [[ -n "${cpus+x}" || -n "${mems+x}" ]]; then
+  if command -v numactl &> /dev/null; then
+      if [[ -n "${cpus+x}" ]]; then
+          set -- --physcpubind "${cpus[$rank]}" "$@"
+      fi
+      if [[ -n "${mems+x}" ]]; then
+          set -- --membind "${mems[$rank]}" "$@"
+      fi
+      set -- numactl "$@"
+  elif [[ -n "${cpus+x}" || -n "${mems+x}" ]]; then
```
manopapad, line 142: you already checked this above ;-)

bryevdv: https://github.com/nv-legate/legate.core/pull/461/commits/6eef4c185b5038fad992352867009e85c7b34c56

legate/driver/command.py
bryevdv: https://github.com/nv-legate/legate.core/pull/461/commits/6eef4c185b5038fad992352867009e85c7b34c56 if I understand correctly, when the the driver launcher is "none" (i.e. usually unspecified, since that is the default) and there at multi-ranks, we want the "auto" behavior.

manopapad (approved):

manopapad (approved):

PR #464, opened 2022-11-03
Support for library specific annotations

magnatelee:
This PR adds an `Annotation` class with which client libraries can attach client specific annotations other than the provenance to operations. Here's an example of how an operator name added as an annotation in cuNumeric gets rendered:

image


legate/core/context.py
```diff
@@ -66,6 +66,35 @@ def find_last_user_frame(libname: str) -> str:
...
+class LibraryAnnotations:
+    def __init__(self) -> None:
+        self._entries: dict[str, str] = {}
+
+    @property
+    def empty(self) -> bool:
+        return len(self._entries) == 0
+
+    def __getitem__(self, key: str) -> Optional[str]:
+        return self._entries[key] if key in self._entries else None
```
bryevdv, line 78: also an option:
```suggestion
        return self._entries.get(key, None)
```

legate/core/context.py
```diff
@@ -66,6 +66,35 @@ def find_last_user_frame(libname: str) -> str:
...
+    def remove(self, key: str) -> None:
+        del self._entries[key]
+
+    def __str__(self) -> str:
+        return "|".join(
+            f"{key},{value}" for key, value in self._entries.items()
+        )
+
+    def __repr__(self) -> str:
+        return str(self)
```
bryevdv, line 92: Could move `__str__` implementation here, and then get `__str__` for free

legate/core/context.py
```diff
@@ -380,3 +417,17 @@ def decorator(func: AnyCallable) -> AnyCallable:
...
+
+
+class Annotation:
+    def __init__(self, context: Context, pairs: dict[str, str]) -> None:
+        self._annotation = context.annotation
+        self._pairs = pairs
+
+    def __enter__(self) -> None:
+        for key, value in self._pairs.items():
+            self._annotation[key] = value
```
bryevdv, line 429: I think `self._annotation.update(**pairs)` would also work

legate/core/context.py
```diff
@@ -66,6 +66,35 @@ def find_last_user_frame(libname: str) -> str:
...
+    def __str__(self) -> str:
+        return "|".join(
+            f"{key},{value}" for key, value in self._entries.items()
+        )
+
+    def __repr__(self) -> str:
+        return str(self)
+
+
+PROVENANCE_KEY = "Provenance"
```
bryevdv, line 95: I usually tack on some dunders for good measure if I am reserving a "special" key name that I don't want to collide with anything a user might supply

bryevdv: Though as long as there is already a class `LibraryAnnotations` couldn't the provenance be stored separately instead of in the same dict?

legate/core/context.py
magnatelee: https://github.com/nv-legate/legate.core/pull/464/commits/828d27955457b8038419ef37e13dac281118a7ac

magnatelee: `__get_item__` and `__set_item__` are removed

legate/core/context.py
magnatelee: https://github.com/nv-legate/legate.core/pull/464/commits/828d27955457b8038419ef37e13dac281118a7ac

legate/core/context.py
magnatelee: https://github.com/nv-legate/legate.core/pull/464/commits/828d27955457b8038419ef37e13dac281118a7ac

legate/core/context.py
magnatelee: https://github.com/nv-legate/legate.core/pull/464/commits/828d27955457b8038419ef37e13dac281118a7ac

magnatelee: took the suggestion that the provenance can be separate

legate/core/context.py
```diff
@@ -66,6 +67,36 @@ def find_last_user_frame(libname: str) -> str:
...
+    def remove(self, key: str) -> None:
+        del self._entries[key]
+
+    def __repr__(self) -> str:
+        pairs = (f"{key},{value}" for key, value in self._entries.items())
+        return "|".join(
+            pairs
+            if self._provenance is None
+            else chain(pairs, (f"Provenance,{self._provenance}",))
+        )
```
bryevdv, line 97: maybe a little simpler, and avoids the `itertools` import:

```suggestion
        if self._provenance is not None:
            pairs += (f"Provenance,{self._provenance}",)
        return "|".join(pairs)
```

bryevdv (approved):

legate/core/context.py
magnatelee: ok I didn't know you could add a tuple to a generator, but wouldn't that materialize the generator into some sequence?

magnatelee: looks like I can't even do it
```
TypeError: unsupported operand type(s) for +=: 'generator' and 'tuple'
```

magnatelee: I mean, I'm fine with materializing the generator into a list, I guess I was trying to be too clever...

bryevdv: oh right, you'd have to call `tuple` on it first, either way seems fine

PR #465, opened 2022-11-04
fix for -ll:networks none, we will init MPI if it has not been initialized

eddy16112:
If `-ll:networks none` is passed, then GASNet will not be initialized, in this case, we will init the MPI explicitly.

### DIFF
--- a/src/core/comm/coll.cc
+++ b/src/core/comm/coll.cc
@@ -245,19 +245,21 @@ int collInit(int argc, char* argv[])
   int provided, init_flag = 0;
   CHECK_MPI(MPI_Initialized(&init_flag));
   if (!init_flag) {
-    log_coll.fatal(
+    // TODO: check if -ll:networks none is enabled
+    log_coll.warning(
       "MPI has not been initialized, it should be initialized by "
-      "the networking backend");
+      "the networking backend, if -ll:networks none is passed, "
+      "then this warning can be safely ignored.");
+    int provided;
+    MPI_Init_thread(0, 0, MPI_THREAD_MULTIPLE, &provided);
+  }
+  int mpi_thread_model;
+  MPI_Query_thread(&mpi_thread_model);
+  if (mpi_thread_model != MPI_THREAD_MULTIPLE) {
+    log_coll.fatal(
+      "MPI has been initialized by others, but is not initialized with "
+      "MPI_THREAD_MULTIPLE");
     LEGATE_ABORT;
-  } else {
-    int mpi_thread_model;
-    MPI_Query_thread(&mpi_thread_model);
-    if (mpi_thread_model != MPI_THREAD_MULTIPLE) {
-      log_coll.fatal(
-        "MPI has been initialized by others, but is not initialized with "
-        "MPI_THREAD_MULTIPLE");
-      LEGATE_ABORT;
-    }
   }
   // check
   int *tag_ub, flag;
--- a/src/core/comm/coll.cc
+++ b/src/core/comm/coll.cc
@@ -244,13 +244,13 @@ int collInit(int argc, char* argv[])
 {
   current_unique_id = 0;
 #ifdef LEGATE_USE_NETWORK
-  int provided, init_flag = 0;
+  int init_flag = 0;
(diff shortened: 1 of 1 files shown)


src/core/comm/coll.cc
```diff
@@ -245,19 +245,21 @@ int collInit(int argc, char* argv[])
...
   if (!init_flag) {
-    log_coll.fatal(
+    // TODO: check if -ll:networks none is enabled
+    log_coll.warning(
       "MPI has not been initialized, it should be initialized by "
-      "the networking backend");
+      "the networking backend, if -ll:networks none is passed, "
+      "then this warning can be safely ignored.");
+    int provided;
+    MPI_Init_thread(0, 0, MPI_THREAD_MULTIPLE, &provided);
```
SeyedMir, line 254: I think you need to finalize MPI in collFinalize if it is initialized by collInit.

manopapad:
If `-ll:networks` is none, that means we're not running in multi-rank/multi-node mode, therefore there's only one node. In that case we shouldn't need to initialize the MPI communicator in the first place. Is that right?

manopapad:
OK, I see what's happening. Currently we decide whether we'll be using the MPI communicator at compile time, based on whether Realm was compiled with a network. However, now it's possible that Realm is compiled with a networking backend, but that backend is not loaded at runtime.

src/core/comm/coll.cc
```diff
@@ -245,19 +245,21 @@ int collInit(int argc, char* argv[])
   int provided, init_flag = 0;
   CHECK_MPI(MPI_Initialized(&init_flag));
   if (!init_flag) {
-    log_coll.fatal(
+    // TODO: check if -ll:networks none is enabled
+    log_coll.warning(
       "MPI has not been initialized, it should be initialized by "
-      "the networking backend");
+      "the networking backend, if -ll:networks none is passed, "
+      "then this warning can be safely ignored.");
```
manopapad, line 252: I would prefer to avoid emitting the warning altogether in that case. If there's a way to ask Realm whether a networking module is loaded, then if Realm answers "no" we know that the user passed `-ll:networks none`.

magnatelee:
I think the right course of action when no network is given is to activate the pthread-based communicator, not initializing the MPI one. A prerequisite is, as Manolis said, being able to query the chosen network module in a programmable way.

src/core/comm/coll.cc
eddy16112: good catch!, thank you

src/core/comm/coll.cc
eddy16112: Removed the warning. Now I will check if `LEGATE_NEED_NETWORK` is larger than 1, we should not init MPI ourselves, but throw an error to let people know they should not use `-ll:networks none` with more than one node.

src/core/comm/coll.cc
```diff
@@ -245,20 +247,28 @@ int collInit(int argc, char* argv[])
   int provided, init_flag = 0;
```
manopapad, line 247: You can remove `provided` from here, it's declared below.

src/core/comm/coll.cc
```diff
@@ -245,20 +247,28 @@ int collInit(int argc, char* argv[])
...
   if (!init_flag) {
-    log_coll.fatal(
-      "MPI has not been initialized, it should be initialized by "
-      "the networking backend");
-    LEGATE_ABORT;
-  } else {
-    int mpi_thread_model;
-    MPI_Query_thread(&mpi_thread_model);
-    if (mpi_thread_model != MPI_THREAD_MULTIPLE) {
+    char* network = getenv("LEGATE_NEED_NETWORK");
```
manopapad, line 250: `LEGATE_NEED_NETWORK` will be either undefined, 0 or 1, it's not the number of nodes.

manopapad:
Is this deprecated by https://github.com/nv-legate/legate.core/pull/468?

eddy16112:
> Is this deprecated by #468?

Not necessary. With this one. the QA team can start testing, but that one requires more time/work to be merged.

manopapad (approved):

PR #466, opened 2022-11-04
New collective

ipdemes:


src/core/mapping/base_mapper.cc
```diff
@@ -207,6 +207,28 @@ void BaseMapper::select_task_options(const MapperContext ctx,
...
+    } else if (req.tag == LEGATE_CORE_KEY_STORE_TAG) {
+      key_functor = find_legate_projection_functor(req.projection);
+      if (key_functor != nullptr && key_functor->is_collective()) {
+        output.check_collective_regions.insert(idx);
+        // std::cout <<"IRINA DEBUG is collective in mapper "<< task.get_task_name()<<std::endl;
+      }
+    }
+    idx++;
+  }
+
```
magnatelee, line 231:
```suggestion
  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
    auto& req = task.regions[idx];
    if (find_legate_projection_functor(req.projection)->is_collective())
      output.check_collective_regions.insert(idx);
  }
```

src/core/runtime/projection.cc
```diff
@@ -123,6 +123,23 @@ AffineFunctor<SRC_DIM, TGT_DIM>::AffineFunctor(Runtime* runtime,
   : LegateProjectionFunctor(runtime), transform_(create_transform(dims, weights))
 {
   for (int32_t dim = 0; dim < TGT_DIM; ++dim) offsets_[dim] = offsets[dim];
+
+  // mapping to a different dimension
+  if (SRC_DIM != TGT_DIM) set_collective();
```
magnatelee, line 128: I assume `set_collective` is monotonic, so you can just return after calling it.

src/core/runtime/projection.cc
```diff
@@ -123,6 +123,23 @@ AffineFunctor<SRC_DIM, TGT_DIM>::AffineFunctor(Runtime* runtime,
...
+
+  // find if there is `-1` in the dimensions
+  std::set<int32_t> unique;
+  for (int32_t dim = 0; dim < SRC_DIM; ++dim) {
+    // std::cout <<"IRINA DEBUG dims = "<<dims[dim]<< " weights = "<< weights[dim]<<", offsets =
+    // "<<offsets[dim]<<std::endl;
+    if (dims[dim] == -1) { set_collective(); }
```
magnatelee, line 135: same here. you don't need to continue the check if you already set the flag.

src/core/mapping/base_mapper.cc
ipdemes: @magnatelee :did you remove 214-216, 220-222  intentionally? Without it reduction and broadcasted instances are not marked as collective.

magnatelee: I forgot a condition in the if statement. Here's the right one:
```
for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
    auto& req = task.regions[idx];
    if (req.handle_type == LEGION_SINGULAR_PROJECTION || find_legate_projection_functor(req.projection)->is_collective())
      output.check_collective_regions.insert(idx);
  }
```
Note that your original code is incorrect because of two reasons: 1) it mis-categorizes a read-only requirement with an identity projection as a broadcasting requirement, 2) it misses a reduction requirement on a region, because that requirement has 0 for its `projection`.

src/core/runtime/projection.h
```diff
@@ -38,10 +38,15 @@ class LegateProjectionFunctor : public Legion::ProjectionFunctor {
...
   virtual unsigned get_depth(void) const { return 0; }
+  virtual bool is_collective() { return _is_collective; }
+  virtual void set_collective() { _is_collective = true; }
 
  public:
   virtual Legion::DomainPoint project_point(const Legion::DomainPoint& point,
                                             const Legion::Domain& launch_domain) const = 0;
+
+ private:
+  bool _is_collective = false;
```
magnatelee, line 49: a nitpick: suffixing, not prefixing, member variables with `_` has been the convention.

src/core/mapping/base_mapper.cc
```diff
@@ -963,8 +973,8 @@ void BaseMapper::legate_select_sources(const MapperContext ctx,
     } else
       band_ranking.push_back(std::pair<PhysicalInstance, uint32_t>(instance, finder->second));
   }
-  assert(!band_ranking.empty());
-  // Easy case of only one instance
+  // assert(!band_ranking.empty());
```
magnatelee, line 976: remove these bogus changes

ipdemes: the ` assert(!band_ranking.empty());` doesn't work anymore when all instances are collective. Do you think I need to modify logic for legate_select_sources` to parse thought collectiveViews as well?

src/core/runtime/projection.h
```diff
@@ -38,10 +38,15 @@ class LegateProjectionFunctor : public Legion::ProjectionFunctor {
   virtual bool is_functional(void) const { return true; }
   virtual bool is_exclusive(void) const { return true; }
   virtual unsigned get_depth(void) const { return 0; }
+  virtual bool is_collective() { return is_collective_; }
```
magnatelee, line 41: this could be a `const` method

magnatelee:
Looks like this PR contains changes from #511, so I hold off on reviewing this until #511 is merged.

src/core/mapping/base_mapper.cc
```diff
@@ -208,6 +208,16 @@ void BaseMapper::select_task_options(const MapperContext ctx,
...
+  LegateProjectionFunctor* key_functor = nullptr;
+  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
+    auto& req   = task.regions[idx];
+    key_functor = find_legate_projection_functor(req.projection);
+    if ((req.handle_type == LEGION_SINGULAR_PROJECTION) ||
+        (key_functor != nullptr && key_functor->is_collective())) {
+      output.check_collective_regions.insert(idx);
+    }
+  }
+
```
magnatelee, line 220:
```suggestion
  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
    auto& req   = task.regions[idx];
    if ((req.handle_type == LEGION_SINGULAR_PROJECTION) ||
        (find_legate_projection_functor(req.projection)->is_collective())) {
      output.check_collective_regions.insert(idx);
    }
  }

```

magnatelee: fyi, `find_legate_projection_functor` can't return a `nullptr` anymore.

magnatelee (approved):

src/core/mapping/base_mapper.cc
```diff
@@ -208,6 +208,14 @@ void BaseMapper::select_task_options(const MapperContext ctx,
                                      const LegionTask& task,
                                      TaskOptions& output)
 {
+  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
```
magnatelee, line 211: Let's wrap this loop with a C macro, say `LEGATE_NO_COLLECTIVE_INSTANCE`, so we can build the code with the `control_replication` branch as well. in the worst case where we don't do perf testing with the `collective` branch, we will need to create packages without collective instances.

ipdemes: sounds good

manopapad:
@ipdemes I think `legate_select_sources` now needs to consider `input.collective_views` as well. According to @lightsighter , `PhysicalInstance`s that are part of a collective view will appear only in `collective_views`, and will not appear in `source_instances`.

PR #468, opened 2022-11-04
Refactor CPU collective communicator

eddy16112:
Support both MPI and local thread communicator even if the code is compiled with GASNet. 
Now, it is easier to swap communicator by either creating a `MPINetwork` or `LocalNetwork` object when initialize the communicator.


src/core/comm/coll.h
```diff
@@ -97,6 +92,150 @@ struct Coll_Comm {
 
 typedef Coll_Comm* CollComm;
 
+class BackendNetwork {
+ public:
+  BackendNetwork();
```
manopapad, line 97: Do we need to declare a virtual destructor also? I know that would normally be necessary to get proper destruction of subclasses referenced through the static variable `BackendNetwork* backend_network`, but possibly virtual destructors are the default for abstract classes (?)

eddy16112: You are correct, we need a virtual destructor, otherwise, the destructor of derived classes will not be called.

manopapad (approved):
It was hard to track the changes in the backend network implementations, when mixed with all the file movement, so I only took a cursory look at those.

src/core/comm/local_comm.cc
```diff
@@ -34,7 +34,7 @@ extern Logger log_coll;
 
 LocalNetwork::LocalNetwork(int argc, char* argv[]) : BackendNetwork()
 {
-  log_coll.print("Enable LocalNetwork");
+  log_coll.info("Enable LocalNetwork");
```
magnatelee, line 37: use `debug` instead. this isn't an `info` worthy message

PR #469, opened 2022-11-07
Add option to have per-file, per-stage test overrides

bryevdv:
This PR adds a test config 
```
CUSTOM_FILES: list[CustomTest] = [
    CustomTest(
        "tests/integration/test_swapaxes.py", "cuda", ["--fbmem", "8192"]
    )
]
```
That can be used to have customized test invocations (per-stage) that override normally computed shard options. Customized tests (if they are present) are executed in serial after all the normal test stage invocations.

@manopapad can see example custom test in https://github.com/nv-legate/legate.core/pull/469/commits/68a35f385840b0d6523b8c62f6b2ee9c53305a67

### DIFF
--- a/legate/tester/__init__.py
+++ b/legate/tester/__init__.py
@@ -17,8 +17,10 @@
 """
 from __future__ import annotations
 
-from typing import Union
+from typing import TYPE_CHECKING, Union
 from typing_extensions import Literal, TypeAlias
+
+from ..util.types import ArgList
 
 #: Define the available feature types for tests
 FeatureType: TypeAlias = Union[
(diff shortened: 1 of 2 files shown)


manopapad (approved):
Looks good. Just a question, if I have a test that I want to run only with specific options, would I need to also include it in `SKIPPED_EXAMPLES`, so it doesn't get run with the default settings?

bryevdv:
> Looks good. Just a question, if I have a test that I want to run only with specific options, would I need to also include it in SKIPPED_EXAMPLES, so it doesn't get run with the default settings?

Yes that would be how to accomplish that with the current PR. I thought about excluding the custom tests from the normal runs, but then we would need to add some mechanism to explicitly add them back in case that was desired. But we already have the converse for skipping, so this seemed simplest.

PR #470, opened 2022-11-07
Use bind.sh for all rank_id handling

bryevdv:
This PR updates the driver to use the `LEGATE_RANK` env var that is now always set by `bind.sh`, in order to set the launcher `rank_id`. This affects the current `nvprof` and `nsys` options. Once the PR is tested on a real cluster and merged, I will make a follow-on PR to add `cprofile` option that uses the same.

### DIFF
--- a/bind.sh
+++ b/bind.sh
@@ -165,4 +165,15 @@ if [[ -n "${cpus+x}" || -n "${mems+x}" ]]; then
   fi
 fi
 
+# arguments may contain the substring %%LEGATE_GLOBAL_RANK%% which needs to be
+# be replaced with the actual computed rank for downstream processes to use
+count=0
+updated=()
+for arg in $@; do
+  updated+=("${arg/\%\%LEGATE_GLOBAL_RANK\%\%/$LEGATE_GLOBAL_RANK}")
+  count=$((count++))
+done
+
+set -- "${updated[@]}"
+
 exec "$@"
--- a/bind.sh
+++ b/bind.sh
@@ -69,27 +69,48 @@ do
 done
 
 case "$launcher" in
-  mpirun) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
-  jsrun ) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
-  srun  ) rank="${SLURM_LOCALID:-unknown}" ;;
-  auto  ) rank="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-unknown}}}" ;;
-  local ) rank="0" ;;
+  mpirun)
+    local_rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}"
+    global_rank="${OMPI_COMM_WORLD_RANK:-unknown}"
+    ;;
+  jsrun )
+    local_rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}"
+    gloabl_rank="${OMPI_COMM_WORLD_RANK:-unknown}"
+    ;;
+  srun  )
+    local_rank="${SLURM_LOCALID:-unknown}"
+    global_rank="${SLURM_PROCID:-unknown}"
+    ;;
+  auto  )
+    local_rank="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-unknown}}}"
+    global_rank="${OMPI_COMM_WORLD_RANK:-${PMI_RANK:-${MV2_COMM_WORLD_RANK:-${SLURM_PROCID:-unknown}}}}"
+    ;;
+  local )
+    local_rank="0"
+    global_rank="0"
+    ;;
   *)
     echo "Unexpected launcher value: $launcher" 1>&2
     help
     ;;
 esac
 
-if [[ "$rank" == "unknown" ]]; then
+if [[ "$local_rank" == "unknown" ]]; then
     echo "Error: Could not determine node-local rank" 1>&2
     exit 1
 fi
 
-export LEGATE_RANK="$rank"
+if [[ "$global_rank" == "unknown" ]]; then
+    echo "Error: Could not determine global rank" 1>&2
+    exit 1
+fi
+
+export LEGATE_LOCAL_RANK="$local_rank"
+export LEGATE_GLOBAL_RANK="$global_rank"
 
 if [ -n "${cpus+x}" ]; then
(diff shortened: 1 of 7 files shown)


legate/driver/launcher.py
```diff
@@ -83,6 +84,15 @@ def __init__(self, config: ConfigProtocol, system: System) -> None:
         self._config = config
         self._system = system
 
+        if config.multi_node.ranks == 1:
+            self.rank_id = "0"
```
bryevdv, line 88: @manopapad I guess maybe we could lose this first condition, `bind.sh` should be taking care of this too

legate/driver/launcher.py
```diff
@@ -51,6 +44,14 @@
 )
 
 
+RANK_ERR_MSG = """\
+Could not detect rank ID on multi-rank run with no --launcher provided. If you
+want Legate to use a launcher, e.g. mpirun, internally (recommended), then you
+need to specify which one to use by passing --launcher. Otherwise you need to
+invoke the legate script itself through a launcher.
```
bryevdv, line 51: Should the message be softened or changed? It now happens if `LEGATE_RANK` is not set, for any reason.

legate/driver/launcher.py
```diff
@@ -29,13 +29,6 @@
 
 __all__ = ("Launcher",)
 
-RANK_ENV_VARS = (
```
manopapad, line 32: So it looks like `LEGATE_RANK` is not set correctly in `bind.sh` currently. Note the difference between the envvars we try in that file, vs the ones we're trying here.

Consider a case where we're running on 3 nodes, with 2 ranks (process copies) per node. In that case there are 6 copies of `legate`, with global ranks (i.e. what would be stored in `OMPI_COMM_WORLD_RANK`) 0, 1, 2, 3, 4, 5 and local ranks (i.e. what would be stored in `OMPI_COMM_WORLD_LOCAL_RANK`) 0, 1, 0, 1, 0 1.

Resource binding (that is the main responsibility of `bind.sh`) cares about the local rank (e.g. each process with node local rank 0 will get the first half of the CPU cores on each node), while the naming of output files needs to differentiate between all processes on all nodes, so it wants the global rank.

Therefore, we need to get both the local and the global rank in `bind.sh`, use the local rank for all the binding happening in that file, and store the global rank in `LEGATE_RANK`, for use by the legate driver.

legate/driver/launcher.py
```diff
@@ -83,6 +84,15 @@ def __init__(self, config: ConfigProtocol, system: System) -> None:
         self._config = config
         self._system = system
 
+        if config.multi_node.ranks == 1:
+            self.rank_id = "0"
+
+        elif "LEGATE_RANK" in system.env:
+            self.rank_id = system.env["LEGATE_RANK"]
```
manopapad, line 91: The legate driver is the one that prepares the call to `bind.sh`. That script hasn't run yet, and therefore `LEGATE_RANK` is not available.

What we could do instead is use a placeholder on the synthesized command line, e.g. `%%RANK%%`, for the different wrappers' output filenames. For example, the command for an invocation wrapped with cprofile might be:

```
mpirun -N 2 ./bind.sh legion_python -m cProfile -o legate_%%RANK%%.py-prof prog.py
```

and an invocation for nvprof would be:

```
mpirun -N 2 ./bind.sh nvprof -o legate_%%RANK%%.nvvp legion_python prog.py
```

Then we would have `bind.sh` find the actual global rank on each invocation, and replace all instances of `%%RANK%%` with that.

bryevdv: :facepalm: sometimes you get an idea in your head, and it just obscures some important detail...

This seem ok, I don't love it exactly, but I'd rather manage this substitution in one place.  I wonder if just using the env var directly could work
```
-o "legate_${LEGATE_RANK}"
```
with so many layers it's not 100% obvious when the expansion would happen tho I'm guessing also too soon, but maybe there is a way to introduce one layer of quoting that gets "naturally" stripped off as the arg is passed down a level.

manopapad: If you just do a `subprocess` call, then it looks like the envvars don't get naturally stripped off:

```
~/Desktop> echo $HOME
/Users/mpapadakis
~/Desktop> cat a.py
import subprocess
subprocess.check_call(["./bind.sh", "ls", "$HOME"])
~/Desktop> cat bind.sh
#!/bin/bash
echo "gonna execute $@"
"$@"
~/Desktop> python a.py
gonna execute ls $HOME
ls: $HOME: No such file or directory
  File "a.py", line 2, in <module>
    subprocess.check_call(["./bind.sh", "ls", "$HOME"])
  File "/Users/mpapadakis/opt/miniconda3/envs/legate/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['./bind.sh', 'ls', '$HOME']' returned non-zero exit status 1.
```

This would probably work if we used `shell=True`, but that's generally frowned upon https://docs.python.org/3/library/subprocess.html#security-considerations.

legate/driver/launcher.py
bryevdv: 038228316d3f5e3824a556eff207108650e54934 should take care of this comment

legate/driver/launcher.py
bryevdv: This message is currently gone since the global rank is replaced by `bind.sh` in all cases (it will error if rank cannot be determined)

legate/driver/launcher.py
bryevdv: @manopapad see https://github.com/nv-legate/legate.core/pull/470/commits/8635b8b0c802a07184db0c98b9f996a2c6eb1578 if this looks good I will update the driver unit tests

bryevdv:
@manopapad as it is this breaks verbose driver printout at start:
```
        if self.config.info.verbose:
            # we only want to print verbose output on a "head" node
            if self.launcher.kind != "none" or self.launcher.rank_id == "0":
                print_verbose(self.system, self)
```
since `rank_id` is always `"%%LEGATE_GLOBAL_RANK%%"` now. 

We could restore `rank_id` to be computed rank id and define some new property for embedding the substitution string into command output. Any thoughts you have are welcome

bind.sh
```diff
@@ -133,15 +154,28 @@ fi
...
+count=0
+updated=()
+for arg in $@; do
+  updated+=("${arg/\%\%LEGATE_GLOBAL_RANK\%\%/$LEGATE_GLOBAL_RANK}")
+  count=$((count+1))
+done
+
+set -- "${updated[@]}"
+
+# echo "bind.sh: $@" 1>&2
```
bryevdv, line 179: I guess printing this unconditionally is not good, but this is the ultimate command that actually get run, I do think we need some mechanism to surface this for debugging, etc.

bryevdv:
superseded by #470

PR #471, opened 2022-11-08
Use bind.sh for all rank detection

bryevdv:
This PR updates the driver to do all rank detection in `bind.sh`. This affects the current `nvprof` and `nsys` options. Once the PR is tested on a real cluster and merged, I will make a follow-on PR to add `cprofile` option that uses the same. 

This PR also adds a driver option `--bind-detail` to the option (and corresponding `--debug` option to `bind.sh`) that will show exactly what invocation `bind.sh` ends up calling. It can be used with `--verbose` and tests, to see what ultimately gets executed, e.g. 

```
+LEGATE_TEST=1 CUNUMERIC_MIN_CPU_CHUNK=2000000000 CUNUMERIC_MIN_OMP_CHUNK=2000000000 CUNUMERIC_MIN_GPU_CHUNK=2000000000 legate /home/bryan/work/cunumeric/tests/integration/test_2d_reduction.py --cpus 1 --cpu-bind 3,9 -v --bind-detail
[PASS] (Eager) tests/integration/test_2d_reduction.py
   bind.sh: numactl --physcpubind 3,9 /home/bryan/work/legate.core/_skbuild/linux-x86_64-3.10/cmake-build/_deps/legion-build/bin/legion_python -ll:py 1 -lg:local 0 -ll:util 2 -ll:csize 4000 -level openmp=5 -lg:eager_alloc_percentage 50 /home/bryan/work/cunumeric/tests/integration/test_2d_reduction.py -v
```

### DIFF
--- a/bind.sh
+++ b/bind.sh
@@ -69,27 +69,48 @@ do
 done
 
 case "$launcher" in
-  mpirun) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
-  jsrun ) rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}" ;;
-  srun  ) rank="${SLURM_LOCALID:-unknown}" ;;
-  auto  ) rank="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-unknown}}}" ;;
-  local ) rank="0" ;;
+  mpirun)
+    local_rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}"
+    global_rank="${OMPI_COMM_WORLD_RANK:-unknown}"
+    ;;
+  jsrun )
+    local_rank="${OMPI_COMM_WORLD_LOCAL_RANK:-unknown}"
+    gloabl_rank="${OMPI_COMM_WORLD_RANK:-unknown}"
+    ;;
+  srun  )
+    local_rank="${SLURM_LOCALID:-unknown}"
+    global_rank="${SLURM_PROCID:-unknown}"
+    ;;
+  auto  )
+    local_rank="${SLURM_LOCALID:-${OMPI_COMM_WORLD_LOCAL_RANK:-${MV2_COMM_WORLD_LOCAL_RANK:-unknown}}}"
+    global_rank="${OMPI_COMM_WORLD_RANK:-${PMI_RANK:-${MV2_COMM_WORLD_RANK:-${SLURM_PROCID:-unknown}}}}"
+    ;;
+  local )
+    local_rank="0"
+    global_rank="0"
+    ;;
   *)
     echo "Unexpected launcher value: $launcher" 1>&2
     help
     ;;
 esac
 
-if [[ "$rank" == "unknown" ]]; then
+if [[ "$local_rank" == "unknown" ]]; then
     echo "Error: Could not determine node-local rank" 1>&2
     exit 1
 fi
 
-export LEGATE_RANK="$rank"
+if [[ "$global_rank" == "unknown" ]]; then
+    echo "Error: Could not determine global rank" 1>&2
+    exit 1
+fi
+
+export LEGATE_LOCAL_RANK="$local_rank"
+export LEGATE_GLOBAL_RANK="$global_rank"
 
 if [ -n "${cpus+x}" ]; then
   cpus=(${cpus//\// })
-  if [[ "$rank" -ge "${#cpus[@]}" ]]; then
+  if [[ "$local_rank" -ge "${#cpus[@]}" ]]; then
       echo "Error: Incomplete CPU binding specification" 1>&2
       exit 1
   fi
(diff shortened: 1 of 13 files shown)


bind.sh
```diff
@@ -133,15 +154,26 @@ fi
...
   else
       echo "Warning: numactl is not available, cannot bind to cores or memories" 1>&2
   fi
 fi
 
+# arguments may contain the substring %%LEGATE_GLOBAL_RANK%% which needs to be
+# be replaced with the actual computed rank for downstream processes to use
+count=0
+updated=()
+for arg in $@; do
```
manopapad, line 172:
```
for arg in "$@"; do
```

bind.sh
```diff
@@ -133,15 +154,26 @@ fi
...
       fi
       set -- numactl "$@"
   else
       echo "Warning: numactl is not available, cannot bind to cores or memories" 1>&2
   fi
 fi
 
+# arguments may contain the substring %%LEGATE_GLOBAL_RANK%% which needs to be
+# be replaced with the actual computed rank for downstream processes to use
+count=0
```
manopapad, line 170: What is this for?

legate/driver/launcher.py
```diff
@@ -380,7 +389,7 @@ class SRunLauncher(Launcher):
     def __init__(self, config: ConfigProtocol, system: System) -> None:
         super().__init__(config, system)
 
-        self.rank_id = "%q{SLURM_PROCID}"
+        self.rank_id = LEGATE_GLOBAL_RANK_SUBSTITUTION
```
manopapad, line 392: Is there a reason to keep the old `.rank_id` field at this point? It looks like we could simply use `LEGATE_GLOBAL_RANK_SUBSTITUTION` directly in `command.py`.

bind.sh
bryevdv: cruft :) first attempt tried to `set` individual args with slices

legate/driver/launcher.py
bryevdv: The thought occurred, I was just trying to keep changed minimal here. But it would allow deleting quite a bit of test complexity so I will proceed with this idea in a follow-on PR.

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/471/commits/b71a55698687ed76659682f3d104446af5d8dc64

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/471/commits/b71a55698687ed76659682f3d104446af5d8dc64

bind.sh
```diff
@@ -133,15 +156,29 @@ fi
...
+# be replaced with the actual computed rank for downstream processes to use
+updated=()
+for arg in "$@"; do
+  updated+=("${arg/\%\%LEGATE_GLOBAL_RANK\%\%/$LEGATE_GLOBAL_RANK}")
+done
+
+set -- "${updated[@]}"
+
+if [ "$debug" == "1" ]; then
+  echo "bind.sh: $@" 1>&2
```
manopapad, line 180: I'd suggest doing it like this https://github.com/nv-legate/quickstart/blob/branch-22.12/common.sh#L116-L117, so that spaces are quoted on the output. Also no need to print to stderr for this message.

manopapad (approved):
A minor comment, otherwise LGTM

bind.sh
bryevdv: https://github.com/nv-legate/legate.core/pull/471/commits/fa410818d407d9f14d03044109b1dd3791c16376

legate/driver/launcher.py
bryevdv: I went ahead and did this here in https://github.com/nv-legate/legate.core/pull/471/commits/718fea851dca78824136e837efa4559b7f5c4756

bind.sh
manopapad: Do we need to be printing this message to stderr (`1>&2`)? I would prefer to use stderr only for errors and warnings, and AFAICT this message is just informational.

legate/driver/launcher.py
```diff
@@ -50,9 +50,6 @@
     "NVIDIA_",
 )
 
-# this will be replaced by bind.sh with the actual computed rank at runtime
-LEGATE_GLOBAL_RANK_SUBSTITUTION = "%%LEGATE_GLOBAL_RANK%%"
-
 
 class Launcher:
```
manopapad, line 54: Also remove "rank_id" from docstring and class member type annotations.

bryevdv: https://github.com/nv-legate/legate.core/pull/471/commits/256ee056e3b15421ec12994629388ba0e7161cce

bind.sh
bryevdv: I guess I am confused by the standard to apply, previously you had wanted the standard help message on stderr as well. Can certainly change in the next PR for adding `--cprofile`

manopapad: The help message is printed right before we exit with a failing exit code, and the output is used as part of the error message whenever the user passes a wrong argument ("argument --foo not recognized, here's what we accept: ..."). So I think it makes sense to print that to stderr. You could make the argument that it should be printed to stdout if the user asks for `--help` explicitly, but I would rather live with this minor inconsistency than add extra cruft.

Conversely, the information message about binding is not pointing to any warning or error, so I'm just going according to the convention we've used elsewhere in the code (e.g. the launcher itself), where we use stdout for verbose informational messages.

I don't know if this convention is best-practice, so I wouldn't disagree if we decided to use stderr for all our informational messaging instead, but then IMHO we should do it consistently across the layers.

bryevdv: it's changed in https://github.com/nv-legate/legate.core/pull/475/commits/20703a649a9ce7f4ffbe502fb4f87f09ea860191

PR #482, opened 2022-11-18
Fixes for mypy>=0.990

magnatelee:


magnatelee:
@bryevdv I had to suppress the type error [here](https://github.com/nv-legate/legate.core/pull/482/files#diff-065d93f49b450e5f99ff0e389b455cf20f3f4298c4620ee12e3f1b62038cc698R53), as I couldn't find the right way to fix it. The mypy error that you see if you remove that comment is legit, because [this type variable](https://github.com/nv-legate/legate.core/blob/branch-22.12/legate/util/types.py#L99) isn't actually what we want; what `object_to_dataclass` really expects is a value of `Type[T]`, not `T` itself, that implements `DataclassProtocol`. Since mypy has no way of telling whether `T` implements `DataclassProtocol`, it refuses to unify it with `Target` in the test.

This could be done easily if mypy supported intersection types, but since it doesn't, I had to work around with multiple inheritance. Unfortunately, I haven't been successful, mostly because I can't inherit `Type[T]`. If you know how to fix it, I'm all ears!

bryevdv:
Hrm you may be right, we could pass dataclass instances, but we would have to give up `frozen=True`. I think just adding an ignore in the test is also file at least for now too, though.

bryevdv:
@magnatelee oh, the real code works because it inherits the mixin:
```
class LegatePaths(DataclassMixin):
```
the test could do that too.

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
           files: \.(cu|cuh|h|cc|inl)$
           types_or: []
     - repo: https://github.com/pre-commit/mirrors-mypy
-      rev: 'v0.982'
+      rev: 'v0.990'
```
bryevdv, line 21: FWIW I think 0.991 is also available now

bryevdv (approved):

.pre-commit-config.yaml
magnatelee: https://github.com/nv-legate/legate.core/pull/482/commits/42e4de630424285b43fcd3757c6bdfb35cee53ed

magnatelee:
> @magnatelee oh, the real code works (I think) because those dataclasses inherit the mixin I defined:
> the test could do that too.

Yup, that worked. https://github.com/nv-legate/legate.core/pull/482/commits/2d82d45222a1416c40a63aacd85eb4c8f7786689

PR #484, opened 2022-11-18
Don't load a Realm network module if running on 1 rank

manopapad:


bryevdv:
Could use a test, there are lots of examples to emulate in https://github.com/nv-legate/legate.core/blob/branch-22.12/tests/unit/legate/driver/test_command.py

eddy16112 (approved):
LGTM

manopapad:
@bryevdv is the unit test good?

bryevdv (approved):

PR #488, opened 2022-11-20
fix several reference cycle / leak related bugs

rohany:
See individual commit comments for details.

### DIFF
--- a/legate/core/partition.py
+++ b/legate/core/partition.py
@@ -38,9 +38,6 @@ if TYPE_CHECKING:
 RequirementType = Union[Type[Broadcast], Type[Partition]]
 
 
-part_mgr = runtime.partition_manager
-
-
 class PartitionBase(ABC):
     @abstractproperty
     def color_shape(self) -> Optional[Shape]:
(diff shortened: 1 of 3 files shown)


rohany:
cc @manopapad, turns out cycles we decided to ignore in the cycle checker that come from `__globals__` are actually important, see the first commit for details.

magnatelee (approved):

manopapad:
@rohany Should/can we then remove the filtering of `__globals__` from the cycle detector here https://github.com/nv-legate/legate.core/blob/branch-22.12/legate/core/cycle_detector.py#L27?

rohany:
I think so? I think those cycles should be "opt-in" to remove, rather than by default. When debugging this i was stuck for a while because the cycle detector didn't report that I had any cycles, so I spent time looking in the wrong places until i found that filtering in the cycle detector.

manopapad:
Are you seeing any spurious cycles if you remove these lines from the filter? https://github.com/nv-legate/legate.core/blob/branch-22.12/legate/core/cycle_detector.py#L27-L29

I think we added this filter based on your feedback, that you were seeing a lot of spurious cycles.

rohany:
I think that problem is that what we deemed as "spurious" is not actually spurious, as it was those cycles that lead to the leaks I debugged here. I only saw those cycles when I made that function return `False` always.

manopapad:
Changed the filter on https://github.com/nv-legate/legate.core/pull/489

PR #490, opened 2022-11-22
Throw an error if an allocation is not consumed

manopapad:


magnatelee:
Is this mainly to help developers catch mistakes? I don't think unconsumed inline allocations would cause any problem. Am I missing anything?

magnatelee:
>  If an allocation goes unconsumed, then decrement_inline_mapped_ref_count is never called, even if the Store gets garbage collected and the RegionField gets reused for another Store. If we then try to inline-map the new Store, the new inline mapping can conflict with the leftover one, producing an error like:

I feel we should just call `decrement_inline_mapped_ref_count` instead of raising an exception. Though highly unlikely, it's technically possible to catch the exception coming from `InlineMappedAllocation.__del__` and move forward as if nothing happened. What do you think?

legate/core/allocation.py
```diff
@@ -40,6 +40,10 @@ def __init__(
         self._strides = strides
         self._consumed = False
 
+    def __del__(self) -> None:
+        if not self._consumed:
+            self._region_field.decrement_inline_mapped_ref_count()
```
magnatelee, line 45: don't we need to pass `unordered=True`?

magnatelee (approved):

PR #494, opened 2022-11-23
Add type for LegateDataInterface

bryevdv:
This PR adds types for `__legate_data_interface__` that can also be added to downstream projects.

Not sure about the names, happy to change based on any suggestions.

### DIFF
--- a/legate/core/legate.py
+++ b/legate/core/legate.py
@@ -23,7 +23,7 @@ from .resource import ResourceConfig
 
 if TYPE_CHECKING:
     from .store import Store
-    from .types import LegateDataInterface
+    from .types import LegateDataInterface, LegateDataInterfaceItem
 
 
 class Array:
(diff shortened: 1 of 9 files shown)


legate/core/legate.py
```diff
@@ -127,30 +128,30 @@ def __init__(self, schema: pyarrow.Schema, columns: list[Array]) -> None:
...
         'version' (required) : int
             An integer showing the version number of this implementation of
             the interface (i.e. 1 for this version)
-        'data' (required) : OrderedDict[Field,Array]
-            An ordered dictionary mapping 'Field' objects that represent the
-            names and types of the field data to 'Array' objects containing
+
+        'data' (required) : dict[Field, Array]
+            An dictionary mapping ``pyarrow.Field`` objects that represent the
+            names and types of the field data to ``Array`` objects containing
```
bryevdv, line 148: `dict` is ordered for all versions of Python we support.

legate/core/runtime.py
```diff
@@ -1208,7 +1208,7 @@ def destroy(self) -> None:
         self.index_spaces = {}
         # Explicitly release the reference to the partition manager so that
         # it may be collected, releasing references to Futures and FutureMaps.
-        self._partition_manager = None  # type: ignore
+        del self._partition_manager
```
bryevdv, line 1211: I think this captures the intention, without muddying the types.

legate/timing/timing.py
```diff
@@ -53,17 +55,15 @@ def measure_nanoseconds(self) -> Future:
...
         self.dtype = dtype
         self.value: Union[int, float, None] = None
 
     @property
-    def __legate_data_interface__(self) -> dict[str, Any]:
-        result: dict[str, Any] = {"version": 1, "data": dict()}
-        result["data"]["Legate Timestamp"] = self
-        return result
+    def __legate_data_interface__(self) -> LegateDataInterfaceItem:
+        return {"version": 1, "data": dict()}
```
manopapad, line 66: @magnatelee Do you still want this to expose the legate data interface? If so, we will need to wrap its Future in a Store, e.g. something like`"data": {pa.Field(...), self.convert_to_store()}`. I don't think this is worth the trouble.

magnatelee: I agree. let's just remove `__legate_data_interface__`.

legate/core/legate.py
manopapad: @magnatelee @lightsighter FYI this changes the specification of the legate data interface, to align with all existing (technically wrong) uses of dict rather than OrderedDict.

legate/timing/timing.py
bryevdv: https://github.com/nv-legate/legate.core/pull/494/commits/e880767d5d01fa04a541baba3f223c9f14154241

manopapad (approved):

bryevdv:
Has to move to `legate.py` to fix a circular import I didn't realize until I actually tried running tests. If possible, `legate.py` might be nice to split up at some point in the future.

PR #496, opened 2022-11-23
Check for cycles involving Futures after runtime shutdown

manopapad:


manopapad:
@rohany can you check whether this change would catch https://github.com/nv-legate/legate.core/pull/496 without reporting useless cycles?

rohany:
Yeah, this should work. A maybe problem is that many `RegionField` leaks reported here are not actually leaks on program shutdown either -- for example, in the `legate-sparse` branch of the core, the `PartitionManager` holds partitions, which hold references to stores. Anything pointed to by the core `Runtime` will be thought of as a cycle before the runtime is destroyed, which leads to some spurious cycles too.

manopapad:
And are these cycles through the `PartitionManager` not impeding the collection of those `RegionField`s?

rohany:
That's a different question I think. These "cycles" are somewhat intentional. One example is the `PartitionManager` has a reference to the `Runtime`, and the `Runtime` has a reference to the `PartitionManager`, so that's a cycle and nothing the partition manager points to can be collected. That means that all the `RegionField`s I take images from etc and store in the `PartitionManager` won't be collected unless explicitly removed, which is a different problem. Cycles with this sort of thing as the root (runtime -> *Manager -> runtime) are somewhat spurious because the runtime deletes these things at shutdown.

manopapad:
One reason for the cycle checker's existence is to detect "transient" cycles that keep RegionFields alive, because they disallow RegionField reuse, and thus instances accumulate at runtime, potentially resulting in OOM. Even if such cycles will be collected when the runtime is destroyed, the accumulation of instances will still happen.

In pure cuNumeric programs, any cycle that is created at any time during program execution that involves `RegionField`s is dangerous. Are the cycles that legate.sparse creates inhibiting `RegionField` reuse, that would otherwise be possible? If yes, then I would consider them "true" cycles, in the sense that they will exacerbate memory problems in larger executions, even if they won't cause shutdown failures.

In any case, it might be for the best to separate the cycle checking phases into RegionField-only (happening right after the user script is done) and Future/FutureMap-only (happening after runtime shutdown). I won't be able to test the second kind, of course, since we're fixing Future leaks as we find them.

rohany:
This discussion makes sense. For the sake of limiting the output generated by this command, can users opt into which one they want to see, instead of it always printing both?

manopapad:
@rohany I separated the future leak option into a separate option `-lg:future-leak-check`. Can you check that it finds the shutdown-causing cycles in your case, without reporting the `RegionField` cycles?

rohany:
> @rohany I separated the future leak option into a separate option -lg:future-leak-check. Can you check that it finds the shutdown-causing cycles in your case, without reporting the RegionField cycles?

LGTM. I don't have the code anymore that has the leak (that version is being used for some debugging other things), but this is how I did it when I was debugging the original problem.

rohany (approved):

PR #500, opened 2022-11-28
Add machinery for legate.core base tests

bryevdv:
This PR added the plain python `pytest` unit tests as well as a Mypy check to legate CI


bryevdv:
This was a bit of a slog, but everything is finally green cc @manopapad @marcinz

manopapad (approved):
Looks great! Our testing infrastructure is shaping up!

PR #503, opened 2022-11-30
Don't access stream pools unless we're on GPUs

### LINKED ISSUE #501: `ReturnValues` should not access GPU streams when not on TOC proc
This code: https://github.com/nv-legate/legate.core/blob/branch-22.12/src/core/task/return.cc#L212 mistakenly tries to access CUDA streams when not tied to a GPU task.

magnatelee:
Fixes #501

### DIFF
--- a/src/core/task/return.cc
+++ b/src/core/task/return.cc
@@ -208,17 +208,19 @@ void ReturnValues::legion_serialize(void* buffer) const
   //
   // the size of value i is computed by offsets[i] - (i == 0 ? 0 : offsets[i-1])
 
-#ifdef LEGATE_USE_CUDA
-  auto stream = cuda::StreamPool::get_stream_pool().get_stream();
-#endif
-
+  // Special case with a single scalar
   if (return_values_.size() == 1) {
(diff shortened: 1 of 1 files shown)


src/core/task/return.cc
```diff
@@ -233,15 +235,25 @@ void ReturnValues::legion_serialize(void* buffer) const
...
+      uint32_t size = ret.size();
+      if (ret.is_device_value())
+        CHECK_CUDA(cudaMemcpyAsync(ptr, ret.ptr(), size, cudaMemcpyDeviceToHost, stream));
+      else
+        memcpy(ptr, ret.ptr(), size);
+      ptr += size;
+    }
+  }
 #endif
+  else {
```
rohany, line 251: I think this dangling `else` will raise an error for you if `LEGATE_USE_CUDA` is not defined. Did you mean the macro else?

magnatelee: https://github.com/nv-legate/legate.core/pull/503/commits/8484263a0c3458291d701611805aa39b625a8686

rohany (approved):

PR #508, opened 2022-12-01
Don't nocr for single node

bryevdv:
In the interest of making future example in docs as clean and unconfusing as possible, this PR removes the "nocr fixup" for single node console usage, where it is not needed, so that users are not confronted by a WARNING that does not apply and that they may not understand.

### DIFF
--- a/legate/driver/config.py
+++ b/legate/driver/config.py
@@ -206,7 +206,7 @@ class Config:
         return not any(opt.endswith(".py") for opt in self.user_opts)
 
     def _fixup_nocr(self, args: Namespace) -> None:
-        if self.console and not args.not_control_replicable:
+        if self.console and not args.not_control_replicable and args.nodes > 1:
             print(warn("Disabling control replication for interactive run"))
             args.not_control_replicable = True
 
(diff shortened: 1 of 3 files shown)


legate/driver/config.py
```diff
@@ -206,7 +206,7 @@ def console(self) -> bool:
         return not any(opt.endswith(".py") for opt in self.user_opts)
 
     def _fixup_nocr(self, args: Namespace) -> None:
-        if self.console and not args.not_control_replicable:
+        if self.console and not args.not_control_replicable and args.nodes > 1:
```
manopapad, line 209: I think this needs to check that ranks > 1 (i.e. even single-node multi-rank runs should trigger this warning, and disable control replication).

bryevdv: sure, this is a "fixup" that happens before `config.multi_node` is initialized from args (because it might alter `not_control_replicable` before that point) so we will have to duplicate the ranks computation but that seems ok.

bryevdv: https://github.com/nv-legate/legate.core/pull/508/commits/20c2505d4cda24e0808b5d1836469c3692f1ae00

manopapad (approved):

PR #510, opened 2022-12-06
Fix for 509

magnatelee:


legate/core/solver.py
```diff
@@ -350,21 +350,21 @@ def compute_launch_shape(
         partitions: dict[PartSym, PartitionBase],
         all_outputs: set[Store],
         unbound_ndim: Optional[int],
-    ) -> Optional[Shape]:
+    ) -> Tuple[Optional[Shape], bool]:
```
rohany, line 353: I would add a comment to this now explaining what the boolean return value is.

legate/core/solver.py
```diff
@@ -476,45 +541,28 @@ def cost(unknown: PartSym) -> tuple[int, bool]:
...
+                constraints,
+                must_be_even,
+            )
 
-            for to_align in cls:
-                if to_align in partitions:
+            launch_shape, done = self.compute_launch_shape(
+                result, all_outputs, unbound_ndim
+            )
+            if can_retry and not done:
```
rohany, line 562: This is a bit of tricky logic that could use a comment.

rohany (approved):
Logic looks good to me, just some requests for comments.

legate/core/solver.py
magnatelee: https://github.com/nv-legate/legate.core/pull/510/commits/595965ec41013f28447c8294b7f8811ff327c763

legate/core/solver.py
magnatelee: https://github.com/nv-legate/legate.core/pull/510/commits/595965ec41013f28447c8294b7f8811ff327c763

rohany (approved):

PR #511, opened 2022-12-08
Implementin logic for reuse of reduction instances 

ipdemes:
- created  ReductionInstanceManager to store information about already created reduction instances
- added logic to the mapper for reuse of reduction instances
- changed coherence for reduction region requirements: it should be ATOMIC to avoid potential thread conflicts in the case several index points of the same IndexTask are using the same reduction instance

### DIFF
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -686,7 +686,7 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
   auto& fields = layout_constraints.field_constraint.field_set;
 
   // We need to hold the instance manager lock as we're about to try to find an instance
-  AutoLock lock(ctx, local_instances->manager_lock());
+  AutoLock reduction_lock(ctx, reduction_instances->manager_lock());
 
   // This whole process has to appear atomic
   runtime->disable_reentrant(ctx);
@@ -745,6 +745,7 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
     return true;
   }
 
+  AutoLock lock(ctx, local_instances->manager_lock());
   // See if we already have it in our local instances
   if (fields.size() == 1 && regions.size() == 1 &&
       local_instances->find_instance(
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -693,18 +693,21 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
 
   // If we're making a reduction instance:
   if (redop != 0) {
-    // See if we already have it in our local instances
-    if (fields.size() == 1 && regions.size() == 1 &&
-        reduction_instances->find_instance(
-          redop, regions.front(), fields.front(), target_memory, result, policy)) {
+    // reuse reductions only for GPU tasks:
+    if (target_proc.kind() == Processor::TOC_PROC) {
+      // See if we already have it in our local instances
+      if (fields.size() == 1 && regions.size() == 1 &&
+          reduction_instances->find_instance(
+            redop, regions.front(), fields.front(), target_memory, result, policy)) {
 #ifdef DEBUG_LEGATE
-      logger.debug() << "Operation " << mappable.get_unique_id()
-                     << ": reused cached reduction instance " << result << " for "
-                     << regions.front();
+        logger.debug() << "Operation " << mappable.get_unique_id()
+                       << ": reused cached reduction instance " << result << " for "
+                       << regions.front();
 #endif
-      runtime->enable_reentrant(ctx);
-      // Needs acquire to keep the runtime happy
-      return true;
+        runtime->enable_reentrant(ctx);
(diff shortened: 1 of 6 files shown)


magnatelee:
> changed coherence for reduction region requirements: it should be ATOMIC to avoid potential thread conflicts in the case several index points of the same IndexTask are using the same reduction instance

sharing the same instance of a region across different point tasks within the same index task doesn't sound like a good idea. isn't that going to serialize their execution?

ipdemes:
@magnatelee : I have made reduction being reused only for GPU tasks. Therefore we can remove "ATOMIC" coherence.

src/core/mapping/base_mapper.cc
```diff
@@ -682,9 +683,34 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
   // Generate layout constraints from the store mapping
   LayoutConstraintSet layout_constraints;
   mapping.populate_layout_constraints(layout_constraints);
+  auto& fields = layout_constraints.field_constraint.field_set;
+
+  // We need to hold the instance manager lock as we're about to try to find an instance
+  AutoLock lock(ctx, local_instances->manager_lock());
```
magnatelee, line 689: Are you sure this is the right manager you want to lock? in the following you use `reduction_instances` if you're creating a reduction instance.

src/core/mapping/instance_manager.h
```diff
@@ -54,10 +54,10 @@ std::ostream& operator<<(std::ostream& os, const RegionGroup& region_group);
 
 struct InstanceSet {
  public:
+  using RegionGroupP = std::shared_ptr<RegionGroup>;
```
magnatelee, line 57: what's the reason for this bogus change?

typings/legion_cffi/lib.pyi
```diff
@@ -33,6 +33,7 @@ LEGION_TYPE_TOTAL: int
 LEGION_READ_ONLY: int
 LEGION_READ_WRITE: int
 LEGION_EXCLUSIVE: int
+LEGION_SIMULTANEOUS: int
```
magnatelee, line 36: I don't think we need this change.

src/core/mapping/instance_manager.h
```diff
@@ -95,15 +95,52 @@ struct InstanceSet {
...
+  struct ReductionInstanceSpec {
+    ReductionInstanceSpec() {}
+    ReductionInstanceSpec(const ReductionOpID& op,
+                          const Instance& inst,
+                          const InstanceMappingPolicy& po)
+      : redop(op), instance(inst), policy(po)
+    {
+    }
+
+    ReductionOpID redop = 0;
```
magnatelee, line 115:
```suggestion
    ReductionOpID redop{0};
```

src/core/mapping/instance_manager.h
```diff
@@ -162,7 +207,31 @@ class InstanceManager {
 
  private:
   std::map<FieldMemInfo, InstanceSet> instance_sets_{};
-  Legion::Mapping::LocalLock manager_lock_{};
+};
+
+class ReductionInstanceManager : public BaseInstanceManager {
```
magnatelee, line 212: Don't inherit `BaseInstanceManager`, as `find`/`record_instance` in that class don't make any sense to reduction instances.

src/core/mapping/instance_manager.h
ipdemes: No real reason, leftover from a change I reverted. I will undo this.

src/core/mapping/instance_manager.h
ipdemes: `BaseInstanceManager` doesn't have these methods, it is just a placeholder for the common data and functions between InstanceManager and ReductionInstanceManager

typings/legion_cffi/lib.pyi
ipdemes: Right.

src/core/mapping/base_mapper.cc
ipdemes: you a right, fixing this

src/core/mapping/base_mapper.cc
```diff
@@ -682,9 +683,34 @@ bool BaseMapper::map_legate_store(const MapperContext ctx,
   // Generate layout constraints from the store mapping
   LayoutConstraintSet layout_constraints;
   mapping.populate_layout_constraints(layout_constraints);
+  auto& fields = layout_constraints.field_constraint.field_set;
+
+  // We need to hold the instance manager lock as we're about to try to find an instance
+  AutoLock reduction_lock(ctx, reduction_instances->manager_lock());
```
magnatelee, line 689: can't this be moved inside the if statement at line 695? we shouldn't hold locks unnecessarily.

ipdemes: sure, makes sense

magnatelee (approved):

PR #516, opened 2022-12-16
Add support for building Legion with the UCX backend

SeyedMir:


manopapad:
@trxcllnt is it possible to add ucx detection and building to the legate.core build (if the user selects `--network ucx`)? Here is an example of how we're doing an external ucx build currently:

```
git clone https://github.com/openucx/ucx
cd ucx
git checkout f8cd013d5f9 # need a recent (unreleased) version; top-of-tree would also work
./autogen.sh
./contrib/configure-release --enable-mt --with-cuda=/usr/local/cuda --with-java=no
make -j install
```

But it looks like ucx also has a cmake build flow.

SeyedMir:
@manopapad where do we stand with this PR? Are we all good or do you want to add UCX installation in here too?

PR #521, opened 2022-12-22
More changes to map futures to zero-copy memory

magnatelee:
This PR is a companion to #518 to make sure futures are mapped to zero-copy unless requested otherwise.

### DIFF
--- a/src/core/data/store.cc
+++ b/src/core/data/store.cc
@@ -136,13 +136,14 @@ FutureWrapper::FutureWrapper(
 #ifdef DEBUG_LEGATE
     assert(!initialize || future_.get_untyped_size() == field_size);
 #endif
-    auto proc = Processor::get_executing_processor();
+    auto proc     = Processor::get_executing_processor();
+    auto mem_kind = find_memory_kind_for_executing_processor(
 #ifdef LEGATE_NO_FUTURES_ON_FB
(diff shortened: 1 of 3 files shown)


src/core/mapping/base_mapper.cc
```diff
@@ -477,16 +477,14 @@ void BaseMapper::map_task(const MapperContext ctx,
   auto default_option            = options.front();
   auto generate_default_mappings = [&](auto& stores, bool exact) {
     for (auto& store : stores) {
+      auto mapping = StoreMapping::default_mapping(store, default_option, exact);
       if (store.is_future()) {
-        auto option  = default_option == StoreTarget::FBMEM ? StoreTarget::ZCMEM : default_option;
```
magnatelee, line 481: Note the change in #518 wasn't actually necessary, as the target will be overriden in the following `map_futures` function anyway.

src/core/data/store.cc
```diff
@@ -136,13 +136,14 @@ FutureWrapper::FutureWrapper(
 #ifdef DEBUG_LEGATE
     assert(!initialize || future_.get_untyped_size() == field_size);
 #endif
-    auto proc = Processor::get_executing_processor();
+    auto proc     = Processor::get_executing_processor();
```
manopapad, line 139: No longer need this line

manopapad (approved):

PR #523, opened 2023-01-03
Update env gen script so OS type works for mac

m3vaz:
If os is specified as `osx` nothing is generated, and if os is specified as `darwin` then input validation throws an error.

Solve this by replacing input ostype as darwin.


m3vaz:
This could also be updated so that everything is osx instead of darwin, that should also work.

bryevdv (approved):

manopapad (approved):

PR #525, opened 2023-01-05
Remove realm pylib check

bryevdv:
This PR removes the check on `REALM_PYTHON_LIB` per:

> realm no longer hardcodes a full path, so someone needs to make sure that libpython.so can be found at runtime through normal .so loading procedures

### DIFF
--- a/legate/driver/launcher.py
+++ b/legate/driver/launcher.py
@@ -15,12 +15,10 @@
 from __future__ import annotations
 
 import os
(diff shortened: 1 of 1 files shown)


manopapad (approved):

manopapad:
More context about the inciting change on the Legion side: https://github.com/StanfordLegion/legion/issues/1363

PR #526, opened 2023-01-09
Don't check for collective behavior when we have WRITE privilege

manopapad:
Otherwise a program like this:

```
M1 = cn.ones((30,30))
M2 = cn.ones((30,30))
M3 = M1.dot(M2)
```

produces this warning:

```
Ignoring request by mapper cunumeric on Node 0 to check for collective usage for region requirement 2 of task cunumeric::MatMulTask (UID 159) because region requirement has writing privileges
```


ipdemes (approved):

ipdemes:
Thank you for fixing this warning!

PR #527, opened 2023-01-11
legate/core/_legion: add default new argument to dep part functions

rohany:
Recent Legion changes added an argument to these functions, we so we need to pass something through.

Signed-off-by: Rohan Yadav <rohany@alumni.cmu.edu>


rohany:
@manopapad @magnatelee can I get a review on this? I'd like to merge it because otherwise all of my legate.sparse CI pipelines are failing.

magnatelee (approved):

PR #530, opened 2023-01-12
Make an image with spy logging, for debugging a cunumeric CI failure

manopapad:


manopapad:
This is no longer necessary

PR #531, opened 2023-01-13
Turn off use of collectives, to check if cuNumeric CI passes

manopapad:
CC @ipdemes, following your suggestion of building an image w/o collective behavior, to see if CI passes


lightsighter:
Just to be clear, this will only be conclusive in one direction. If it fails without the collectives then we know they aren't to blame. However, if it doesn't fail, that doesn't necessarily imply that the collectives are to blame because you're likely going to perturb the timing quite by not using them and therefore you might just be observing a false-negative.

manopapad:
This is no longer necessary

PR #532, opened 2023-01-18
adding logic for collective instances to the legate_select_sources

ipdemes:


src/core/mapping/base_mapper.cc
```diff
@@ -941,34 +976,37 @@ void BaseMapper::legate_select_sources(const MapperContext ctx,
...
+                                     all_local,
+                                     source_memories,
+                                     band_ranking,
+                                     affinity,
+                                     destination_memory,
+                                     machine);
+    }
+  }  // if (!all_local)
+
+  assert(!band_ranking.empty());
```
manopapad, line 1009: Probably want to do this only if DEBUG_LEGATE

src/core/mapping/base_mapper.cc
```diff
@@ -941,34 +976,37 @@ void BaseMapper::legate_select_sources(const MapperContext ctx,
...
-      uint32_t memory_bandwidth = 0;
-      if (!affinity.empty()) {
-        assert(affinity.size() == 1);
-        memory_bandwidth = affinity[0].bandwidth;
-      }
-      source_memories[location] = memory_bandwidth;
-      band_ranking.push_back(std::pair<PhysicalInstance, uint32_t>(instance, memory_bandwidth));
-    } else
-      band_ranking.push_back(std::pair<PhysicalInstance, uint32_t>(instance, finder->second));
+    if (!all_local)  // Skip any remote instances once we're local
```
manopapad, line 979: I don't think you want to do this check here. Otherwise once you hit the first local instance, you will ignore all the other local instances that come after, even if they happen to have better bandwidth.

manopapad: And I think the same is true about the two `!all_local` checks below.

ipdemes: yes, you are right. I will push the fix shortly.

src/core/mapping/base_mapper.cc
```diff
@@ -941,34 +979,36 @@ void BaseMapper::legate_select_sources(const MapperContext ctx,
...
+                                 affinity,
+                                 destination_memory,
+                                 machine);
   }
-  // If there aren't any sources (for example if there are some collective views
-  // to choose from, not yet in this branch), just return nothing and let the
-  // runtime pick something for us.
-  if (band_ranking.empty()) { return; }
+
+  if (!all_local)  // no need to go through the collective instanes if there
```
manopapad, line 992: I feel like we'd want to go through the collective views, even if we found a local stand-alone instance `I`. There might be another instance in a collective view that's actually closer to the target than `I`.

lightsighter: It also might be more expensive to figure out which one you want. If you find a local instance in the same memory as the target, you aren't going to do better than that digging through a collective instance.

src/core/mapping/base_mapper.cc
```diff
@@ -921,12 +921,47 @@ void BaseMapper::select_task_sources(const MapperContext ctx,
...
+                                  const Legion::Machine& machine)
+{
+  Memory location = instance.get_location();
+  if (location.address_space() == local_node) {
+    if (!all_local) {
+      source_memories.clear();
+      band_ranking.clear();
+      all_local = true;
+    }
+  }
```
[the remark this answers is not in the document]
ipdemes: I think I have added it with my last commit

src/core/mapping/base_mapper.cc
manopapad: Note that "local" in this code refers to "same address space", not necessarily "directly on the target memory", so there is a (small) possibility that we are trying to source an instance on GPU 0, there is a standalone instance `I` on GPU 4 on the same node but a different NUMA domain, and a collective view contains an instance `J` that's on GPU 1, which is on the same NUMA domain as GPU 0 (or even directly on GPU 0!).

I don't like having an arbitrary fork in the logic in this case, where a whole class of possibly same-address-space instances are ignored, but only if they're part of a collective view. Unless we're trying hard to avoid a call to `find_instances_nearest_memory`, because it's expected to be prohibitively expensive (e.g. if it triggers remote queries) -- @lightsighter ?

A safe optimization that we could do instead in the `all_local` case is to only consider instances within the `CollectiveView` that are on the local address space. There doesn't seem to be a way to filter based on that characteristic on `CollectiveView`, so either we can add that API call, or find all the local memories and iterate over those using `find_instances_in_memory`. Ideally, however, `CollectiveView` would just provide an iterator over contained instances. This would be more practical in the general case, because then a mapper is free to prioritize however it wants, including on things other than memory channel latency/bandwidth, e.g. instance layout.

src/core/mapping/base_mapper.cc
```diff
@@ -921,12 +921,50 @@ void BaseMapper::select_task_sources(const MapperContext ctx,
...
+void add_instance_to_band_ranking(const PhysicalInstance& instance,
+                                  const Legion::AddressSpace& local_node,
+                                  bool& all_local,
+                                  std::map<Memory, uint32_t /*bandwidth*/>& source_memories,
+                                  std::vector<std::pair<PhysicalInstance, uint32_t>>& band_ranking,
+                                  std::vector<MemoryMemoryAffinity>& affinity,
+                                  const Memory& destination_memory,
+                                  const Legion::Machine& machine)
+{
+  Memory location = instance.get_location();
```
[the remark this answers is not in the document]
ipdemes: @manopapad : just to make sure I understand this correctly, in this case (`location==destination_memory`), we will add this instance to the `ranking` and exit ? Do we care about other instances or we want to find the one with the largest bandwidth in the `select_task_sources` ?

manopapad:
So after a chat with @lightsighter I realized a fundamental issue with our logic here. We are assuming that we can skip some of the instances (e.g. in the case where we find at least one local instance, we completely ignore all remote instances). However, nowadays we are *not* guaranteed that all candidate source instances will completely cover the target instance. If the first instance we return to the runtime is not enough to completely cover the target instance, then the runtime will try the next one, and so on. If there are no more instances left in the output, the runtime will make an arbitrary decision.

Consider, for example, the case where `I1` is a local instance covering the first half the target instance, and `I2`, `I3` are remote instances covering the second half. In the current model, we would be returning only `I1`, and essentially letting the runtime choose between `I2` and `I3`.

We should instead always sort and return the full list of candidate source instances. I propose we:

* completely remove the logic about `all_local`
* for each collective view, we pick the instance that is closest to the target memory, and add it to the list of source instances
* we sort the (augmented) list of source instances and return it

I believe that is a good baseline, that will return consistent rankings. We could optimize further potentially, but this is fine for now.

src/core/mapping/base_mapper.cc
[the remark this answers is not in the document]
manopapad: My suggestion was actually wrong, read below.

src/core/mapping/base_mapper.cc
manopapad: This conversation was missing some context, see my updated comment below.

ipdemes:
@manopapad : just to make sure I understand this correctly:  for the second bullet you are not suggesting to modify `input.source_instances` but rather adding these collective instances to the list of instances we loop through in our `legate_select_sources` logic( e.g temporary vector)
@lightsighter would it be sufficient to take the first instance from the vector of instances returned by `find_instances_nearest_memory` to ` pick the instance that is closest to the target memory` (e.g does the `find_instances_nearest_memory` return list of instances sorted by bandwidth)?

manopapad:
> for the second bullet you are not suggesting to modify input.source_instances but rather adding these collective instances to the list of instances we loop through in our legate_select_sources logic (e.g temporary vector)

Yup, that's my suggestion.

manopapad:
Also @rohany brought up another point against the "all_local" heuristic. Imagine we are running multi-rank, are trying to source a device instance, and there is a source instance on a system memory on the local process, and another source instance on a device memory on another process on the same node. The old heuristic would pick the "local" host instance, even though most likely the device instance on the other process would be faster to pull.

src/core/mapping/base_mapper.cc
```diff
@@ -928,54 +928,69 @@ void BaseMapper::select_task_sources(const MapperContext ctx,
...
 void BaseMapper::legate_select_sources(const MapperContext ctx,
                                        const PhysicalInstance& target,
                                        const std::vector<PhysicalInstance>& sources,
+                                       const std::vector<CollectiveView>& collective_sources,
                                        std::deque<PhysicalInstance>& ranking)
 {
   std::map<Memory, uint32_t /*bandwidth*/> source_memories;
   // For right now we'll rank instances by the bandwidth of the memory
   // they are in to the destination, we'll only rank sources from the
   // local node if there are any
```
manopapad, line 971:
```suggestion
  // they are in to the destination.
```

src/core/mapping/base_mapper.cc
```diff
@@ -928,54 +928,68 @@ void BaseMapper::select_task_sources(const MapperContext ctx,
...
+                                  std::map<Memory, uint32_t /*bandwidth*/>& source_memories,
+                                  std::vector<std::pair<PhysicalInstance, uint32_t>>& band_ranking,
+                                  std::vector<MemoryMemoryAffinity>& affinity,
+                                  const Memory& destination_memory,
+                                  const Legion::Machine& machine)
+{
+  Memory location = instance.get_location();
+  auto finder     = source_memories.find(location);
+  if (finder == source_memories.end()) {
+    affinity.clear();
```
manopapad, line 946: I think we can just declare `affinity` as a temporary here, instead of passing it and clearing it.

manopapad (approved):
One minor comment, otherwise looks good

PR #533, opened 2023-01-18
Don't turn on Legate debug checks on debug-rel builds

manopapad:


rohany:
i think you need the same PR also for cunumeric

trxcllnt (approved):

PR #535, opened 2023-01-19
Build rust `legion_prof`

trxcllnt:


marcinz:
To have this build with conda, I think it will require that we add rust as a dependency in the conda package spec. Could you try if adding the `rust` package in the host section fixes the build error?

PR #538, opened 2023-01-19
Change the default CPU architecture to haswell.

marcinz:
Change the default architecture to haswell. Currently, we build our conda packages with haswell to avoid compatibility issues. Changing the default architecture to haswell will make default builds comparable to the conda packages and it will allow developers to share sccache with CI builds.

### DIFF
--- a/install.py
+++ b/install.py
@@ -574,7 +574,7 @@ def driver():
         "--march",
         dest="march",
         required=False,
-        default="native",
+        default="haswell",
         help="Specify the target CPU architecture.",
     )
     parser.add_argument(
--- a/install.py
+++ b/install.py
@@ -574,7 +574,7 @@ def driver():
         "--march",
         dest="march",
         required=False,
-        default="haswell",
+        default=("haswell" if platform.machine() == "x86_64" else "native"),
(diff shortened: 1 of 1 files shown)


manopapad:
This should be the default only when building on an x86 machine.

Please add a comment for why we're doing this in the script itself.

This should also be changed on the cunumeric side.

marcinz:
@manopapad Do we have a check whether we are building on an x86 machine? If we do not, maybe this is not a good idea.

manopapad:
I think it could just be
```
default=("haswell" if platform.machine() == "x86_64" else "native"),
```
but make @bryevdv @m3vaz have a better suggestion?

bryevdv:
Seems reasonable to me

install.py
```diff
@@ -574,7 +574,7 @@ def driver():
         "--march",
         dest="march",
         required=False,
-        default="native",
+        default="haswell",
```
trxcllnt, line 577:
```suggestion
        default=("haswell" if platform.machine() == "x86_64" else "native"),
```

manopapad (approved):

PR #539, opened 2023-01-20
Default python interpreter support for Legate

eddy16112:
This PR allows people to run legate application using default python interpreter.


legate/core/corelib.py
```diff
@@ -59,6 +59,7 @@ def get_resource_configuration(self) -> ResourceConfig:
         return config
 
     def destroy(self) -> None:
+        print("destroy corelib")
```
manopapad, line 62: Leftover debugging code

eddy16112: Yes, I plan to remove it before the merge.

legate/driver/command.py
```diff
@@ -395,6 +401,28 @@ def cmd_user_opts(
     cmd_log_levels,
     cmd_log_file,
     cmd_eager_alloc,
+    # User script
+    cmd_user_script,
+    # Append user flags so they can override whatever we provided
+    cmd_user_opts,
+)
+
+CMD_PARTS_CANONICAL = (
```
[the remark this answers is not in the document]
bryevdv: It also looks like `cmd_cpus` through `cmd_eager_alloc` alloc are identical in both lists. I'd also suggest factoring that part of the sequence out and then assembling
```python
CMD_PARTS_LEGION = (
    cmd_bind,
    cmd_rlwrap,
    cmd_gdb,
    cmd_cuda_gdb,
    cmd_nvprof,
    cmd_nsys,
    # Add memcheck right before the binary
    cmd_memcheck,
    # Now we're ready to build the actual command to run
    cmd_legion,
    # This has to go before script name
    cmd_nocr,
    cmd_module,
    cmd_processor,
    cmd_kthreads,
) + _CMD_PARTS_SHARED + (
    # User script
    cmd_user_script,
    # Append user flags so they can override whatever we provided
    cmd_user_opts,
)

CMD_PARTS_CANONICAL = (
    # User script
    cmd_user_script,
) + _CMD_PARTS_SHARED + (
    # Append user flags so they can override whatever we provided
    cmd_user_opts,
)
That way if a new shared step is added in the future it only needs to be added in one place. 
```

legate/driver/main.py
```diff
@@ -60,3 +62,46 @@ def main(argv: list[str]) -> int:
...
+
+    try:
+        driver = CanonicalDriver(config, system)
+    except Exception as e:
+        msg = "Could not initialize canonical driver, path config and exception follow:"  # noqa
+        print(error(msg))
+        print_verbose(system)
+        raise e
+
+    return driver.cmd, driver.env
```
bryevdv, line 107: I'm not sure this function belongs here and I am not sure it should be called a "main" since main function are typically responsible for ultimately running something as an "entry point", and this function does not do that. Maybe `prepare_legion_main_args` or similar would be more descriptive?

bryevdv: A possible factorization:

```python
def prepare_driver(driver_type):
    from ..util.system import System
    from ..util.ui import error
    from . import Config
    from .driver import print_verbose

    try:
        config = Config(argv)
    except Exception as e:
        print(error("Could not configure canonical driver:\n"))
        raise e

    try:
        system = System()
    except Exception as e:
        print(error("Could not determine System settings: \n"))
        raise e

    try:
        driver = driver_type(config, system)
    except Exception as e:
        msg = "Could not initialize driver, path config and exception follow:"  # noqa
        print(error(msg))
        print_verbose(system)
        raise e

    return driver
````

This function could be called by `main` (passing in a plain `Driver`) and also used up in `__init__` (passing in a `CanonicalDriver`)

This is also another case we might want to rename `Driver` to something more specific like `LegateDriver` now that we are adding another specialized kind of driver alongside.

File: legate/driver/command.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (eddy16112) [reply legate/driver/command.py:418]: Good point, I have fixed it by using your code.

File: legate/driver/config.py [hunk 4]
```diff
@@ -203,7 +212,7 @@ def __init__(self, argv: ArgList) -> None:
     @cached_property
     def console(self) -> bool:
         """Whether we are starting Legate as an interactive console."""
-        return not any(opt.endswith(".py") for opt in self.user_opts)
+        return self.user_script == ""
```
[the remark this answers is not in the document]
eddy16112: Yes, I have set the `user_script` to None by default.

legate/driver/command.py
```diff
@@ -24,7 +24,7 @@
     from .config import ConfigProtocol
     from .launcher import Launcher
 
-__all__ = ("CMD_PARTS",)
+__all__ = ("_CMD_PARTS_SHARED", "CMD_PARTS_LEGION", "CMD_PARTS_CANONICAL")
```
bryevdv, line 27: No need to "export" private names
```suggestion
__all__ = ("CMD_PARTS_LEGION", "CMD_PARTS_CANONICAL")
```

legate/driver/main.py
eddy16112: I have renamed the `Driver` to `LegateDriver`, `main` to `legate_main`.  I also unified `legate_main` with `canonical_main` by using the `prepare_driver`, so now both `legate_main` and `canonical_main` calls `prepare_driver` to create a `LegateDriver` or `CanonicalDriver`.

legate/driver/main.py
```diff
@@ -79,29 +89,5 @@ def canonical_main(argv: list[str]) -> Tuple[Tuple[str, ...], Dict[str, str]]:
...
-
-    try:
-        driver = CanonicalDriver(config, system)
-    except Exception as e:
-        msg = "Could not initialize canonical driver, path config and exception follow:"  # noqa
-        print(error(msg))
-        print_verbose(system)
-        raise e
-
+    driver = prepare_driver(argv, CanonicalDriver)
```
bryevdv, line 92: I would suggest getting rid of `canonical_main` altogether (it doesn't really do what a "main" function usually does) Instead. Just have the code up in `legate/core/__init__.py` call  `prepare_driver(argv, CanonicalDriver)` itself directly.

legate/driver/driver.py
```diff
@@ -65,7 +65,14 @@ def cmd(self) -> Command:
...
         system = self.system
 
-        parts = (part(config, system, launcher) for part in CMD_PARTS)
+        if config.user_script is None:
+            # run with console, so remove the cmd_user_script
+            new_cmd_parts = tuple(
+                x for x in CMD_PARTS_LEGION if x != CMD_PARTS_LEGION[-2]
+            )
+        else:
+            new_cmd_parts = CMD_PARTS_LEGION
```
bryevdv, line 74: This conditional is no longer necessary, correct? `cmd_user_script` is a no-op if `config.user_script` is None, so it can just aways be included unconditionally, I think.

If that is not the case, then then condition needs to be made more robust:
```
x for x in CMD_PARTS_LEGION if x != cmd_user_script
```
Relying on exactly `CMD_PARTS_LEGION[-2]` to be `cmd_user_script` is fragile in case the order or contents of `CMD_PARTS_LEGION` has to change in the future.

legate/driver/main.py
eddy16112: I see, but isn't it cleaner to wrap `prepare_driver` into `legate_main` and `canonical_main` and then we can call `canonical_main` directly inside `legate/core/__init__.py`?

bryevdv: I don't personally think it is cleaner, since calling a function like this "main" seems confusing to me (usually a main function starts a blocking loop event loop, etc and runs for the duration of execution). It also seems superfluous since `canonical_main` really doesn't do anything else at all apart from call `prepare_driver`.

eddy16112: OK, do you think if we should remove `legate_main`? The `legate_main` only calls `prepare_driver` as well.

bryevdv: I would personally keep it, because it is responsible for calling `driver.run()` on the result of `prepare_driver` (which is a blocking API call to "start" the application, and that is actually want a "main" is typically responsible for)

legate/driver/driver.py
eddy16112: You are right, the if else is no longer needed, and I have removed it.

legate/driver/config.py
```diff
@@ -184,7 +185,15 @@ def __init__(self, argv: ArgList) -> None:
...
-        self.user_opts = tuple(extra)
+        self.user_script = ""
+        for opt in tuple(extra):
+            if opt.endswith(".py"):
+                self.user_script = opt
+                break
+
+        self.user_opts = tuple(
+            x for x in tuple(extra) if x != self.user_script
+        )
```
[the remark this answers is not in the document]
eddy16112: I have modified the code to take care of this case, https://github.com/eddy16112/legate.core/blob/canonical_python/legate/driver/config.py#L190-L197

legate/driver/main.py
```diff
@@ -17,32 +17,26 @@
...
 
-__all__ = ("main",)
+from typing import Type, Union
 
+from . import CanonicalDriver, LegateDriver
 
-def main(argv: list[str]) -> int:
-    """A main function for the Legate driver that can be used programmatically
-    or by entry-points.
+__all__ = "legate_main"
```
bryevdv, line 24: value for all needs to be a sequence
```suggestion
__all__ = ("legate_main",)
```

bryevdv (commented):
Looking really good, only a couple of final comments

legate/core/__init__.py
```diff
@@ -17,9 +17,40 @@
...
+        "python",
+    ] + sys.argv
+    driver = prepare_driver(sys_argv, CanonicalDriver)
+    legate_argv = driver.cmd
+    legate_env = driver.env
+    # sys_argv = sys.argv[0:]
+    print(legate_argv)
+    # print(legate_env)
+    for key, value in legate_env.items():
+        if (
```
manopapad, line 42: Can we reuse the list of prefixes from here? https://github.com/nv-legate/legate.core/blob/branch-22.12/legate/driver/launcher.py#L37

bryevdv: If so there is also already

https://github.com/nv-legate/legate.core/blob/025a66ba7a2134c5f1652b99ef28969d6b7107aa/legate/driver/launcher.py#L146-L151

legate/driver/command.py
```diff
@@ -395,6 +386,43 @@ def cmd_user_opts(
...
+        cmd_nsys,
+        # Add memcheck right before the binary
+        cmd_memcheck,
+        # Now we're ready to build the actual command to run
+        cmd_legion,
+        # This has to go before script name
+        cmd_nocr,
+        cmd_module,
+        cmd_processor,
+        cmd_kthreads,
```
manopapad, line 407: Is the force-kthreads option not compatible with canonical python mode?

legate/driver/command.py
```diff
@@ -395,6 +386,43 @@ def cmd_user_opts(
     cmd_log_levels,
     cmd_log_file,
     cmd_eager_alloc,
-    # Append user flags so they can override whatever we provided
-    cmd_user_opts,
+)
+
+CMD_PARTS_LEGION = (
```
manopapad, line 391: We should document somehow which options are only available under "legion_python" mode. Possibly these options can be listed under a separate header on the `legate --help` output.

bryevdv: Unless I'm mistaken, all of the options are available via `legate` and only some are not available, when using "plain" python. Are you saying `legate --help` should be responsible for pointing out options that are not available when _not_ using `legate`? I guess I don't think `legate --help` needs to be responsible for non-`legate` usage.

manopapad:
We should document this new capability on the README. Also how it "consumes" arguments, its shortcomings, how it works with launchers like `mpirun`, and when it's preferable vs `legion_python` mode.

Doesn't necessarily need to happen on this PR, but we definitely need to have documentation before we advertise this feature.

manopapad:
I expect this new feature will make it even more apparent how our argument parsing logic clashes with common python practice.

Take these two programs:

```
import legate.core # happens before argument parsing
import argparse
parser = argparse.ArgumentParser()
parser.add_argument("filename")
args = parser.parse_args()
```

and

```
import argparse
parser = argparse.ArgumentParser()
parser.add_argument("filename")
args = parser.parse_args()
import legate.core # happens after argument parsing
```

IIUC, running `python prog.py --help` will produce a different result for the two programs; the first will print the help message for legate, whereas the second will print the help message for the program.

Previously this behavior might have been more acceptable, since we were using a custom executable. But now we are telling users they can just use the normal python interpreter, where this behavior will be quite surprising. We may want to reconsider enforcing a strict separation of legate vs user script arguments:

```
python/legate <legate-args> prog.py <prog-args>
```

legate/driver/command.py
manopapad: That's fine, but in that case the documentation for common options will need to be duplicated somewhere. I was thinking that we could just tell the users in the README that when they use a legate library some options will be consumed, and they should run `legate --help` to see more.

legate/driver/command.py
eddy16112: Good catch! It should be fixed by https://github.com/nv-legate/legate.core/pull/539/commits/3f89270fb8c5471737ccda8d618f81f107088471

legate/core/__init__.py
eddy16112: I just tried `is_launcher_var`, and it returns:
```
PYTHONPATH /scratch2/wwu/legate.coll/legion/bindings/python/build/lib:/scratch2/wwu/legate.coll/legion/jupyter_notebook:/scratch2/wwu/legate.coll/legate
LD_LIBRARY_PATH /scratch2/wwu/legate.coll/_skbuild/linux-x86_64-3.9/cmake-build/_deps/legion-build/lib:/scratch2/wwu/legate.coll/build/lib:/usr/local/openmpi-4.1.0/lib:/scratch2/wwu/anaconda3/envs/legate-test/lib:/scratch2/wwu/legate.coll/_skbuild/linux-x86_64-3.9/cmake-install/lib:/usr/local/cuda-11.1/lib64:/usr/local/openmpi-4.1.0/lib
PATH /usr/local/openmpi-4.1.0/bin:/usr/local/openmpi-4.1.0/bin:/scratch2/wwu/anaconda3/envs/legate-test/bin:/scratch2/wwu/anaconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/opt/puppetlabs/bin:/usr/local/cuda-11.1/bin:/usr/local/slurm-20.11.4/bin
CONDA_EXE /scratch2/wwu/anaconda3/bin/conda
CONDA_PREFIX /scratch2/wwu/anaconda3/envs/legate-test
MANPATH ::/opt/puppetlabs/puppet/share/man:/usr/local/slurm-20.11.4/share/man
CONDA_PROMPT_MODIFIER (legate-test) 
CONDA_SHLVL 2
CONDA_PYTHON_EXE /scratch2/wwu/anaconda3/bin/python
CONDA_DEFAULT_ENV legate-test
MODULEPATH /etc/environment-modules/modules:/usr/share/modules/versions:/usr/share/modules/$MODULE_VERSION/modulefiles:/usr/share/modules/modulefiles:/usr/local/modules
CONDA_PREFIX_1 /scratch2/wwu/anaconda3
PMIX_DSTORE_21_BASE_PATH /tmp/ompi.g0004.1052/pid.860378/pmix_dstor_ds21_860378
PMIX_DSTORE_ESH_BASE_PATH /tmp/ompi.g0004.1052/pid.860378/pmix_dstor_ds12_860378
PYTHONDONTWRITEBYTECODE 1
NCCL_LAUNCH_MODE PARALLEL
GASNET_MPI_THREAD MPI_THREAD_MULTIPLE
LEGATE_NEED_CUDA 1
LEGATE_NEED_NETWORK 1
LEGATE_MAX_DIM 4
LEGATE_MAX_FIELDS 256
REALM_BACKTRACE 1
```
I am wondering if we need so many `CONDA` envs?

manopapad:
This is not working for me on an editable install on my local Mac. It looks like the liblegion_canonical_python library is not found:

<details>
<summary>test program</summary>

```
~/legate.core> cat a.py
import legate.core
```
</details>

<details>
<summary>output</summary>

```
~/legate.core> python a.py --verbose
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_cffi.py", line 31, in <module>
    lib.legion_runtime_has_context()
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 912, in __getattr__
    make_accessor(name)
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 908, in make_accessor
    accessors[name](name)
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 838, in accessor_function
    value = backendlib.load_function(BType, name)
AttributeError: function/symbol 'legion_runtime_has_context' not found in library '<None>': dlsym(RTLD_DEFAULT, legion_runtime_has_context): symbol not found

During handling of the above exception, another exception occurred:

  File "a.py", line 1, in <module>
    import legate.core
  File "/Users/mpapadakis/legate.core/legate/core/__init__.py", line 20, in <module>
    from legion_cffi import is_legion_python, ffi, lib as legion
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_cffi.py", line 35, in <module>
    from legion_canonical_cffi import ffi, lib
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_canonical_cffi.py", line 33, in <module>
    lib = ffi.dlopen(libname)
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 150, in dlopen
    lib, function_cache = _make_ffi_library(self, name, flags)
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 832, in _make_ffi_library
    backendlib = _load_backend_lib(backend, libname, flags)
  File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 827, in _load_backend_lib
    raise OSError(msg)
OSError: cannot load library '/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib': dlopen(/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib, 2): Library not loaded: @rpath/liblegion.1.dylib
  Referenced from: /Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib
  Reason: image not found.  Additionally, ctypes.util.find_library() did not manage to locate a library called '/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib'
```
</details>

<details>
<summary>library RPATH information</summary>

```
~/legate.core/_skbuild/macosx--x86_64-3.8> otool -l ./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib | grep -A 2 RPATH
          cmd LC_RPATH
      cmdsize 64
         path /Users/mpapadakis/opt/miniconda3/envs/test/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 112
         path /Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-build/_deps/legion-build/lib (offset 12)
~/legate.core/_skbuild/macosx--x86_64-3.8> otool -L ./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib
./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib:
	@rpath/liblegion_canonical_python.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/liblegion.1.dylib (compatibility version 1.0.0, current version 0.0.0)
	@rpath/librealm.1.dylib (compatibility version 1.0.0, current version 0.0.0)
	@rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
~/legate.core/_skbuild/macosx--x86_64-3.8> otool -l ./cmake-install/lib/liblegion_canonical_python.dylib | grep -A 2 RPATH
          cmd LC_RPATH
      cmdsize 64
         path /Users/mpapadakis/opt/miniconda3/envs/test/lib (offset 12)
~/legate.core/_skbuild/macosx--x86_64-3.8> otool -L ./cmake-install/lib/liblegion_canonical_python.dylib
./cmake-install/lib/liblegion_canonical_python.dylib:
	@rpath/liblegion_canonical_python.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/liblegion.1.dylib (compatibility version 1.0.0, current version 0.0.0)
	@rpath/librealm.1.dylib (compatibility version 1.0.0, current version 0.0.0)
	@rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
</details>
```

legate/core/__init__.py
manopapad: These are variables set by conda. We are not setting them, we're just forwarding them.

Looking at this again, I don't think it's necessary to do this filtering at all. IIUC the driver is only appending some custom variables to the environment, it's not removing anything that's already there. Therefore, it should be enough to do `os.environ = legate_env`.

eddy16112:
> This is not working for me on an editable install on my local Mac. It looks like the liblegion_canonical_python library is not found:
> 
> test program
> ```
> ~/legate.core> cat a.py
> import legate.core
> ```
> 
> output
> ```
> ~/legate.core> python a.py --verbose
> Traceback (most recent call last):
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_cffi.py", line 31, in <module>
>     lib.legion_runtime_has_context()
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 912, in __getattr__
>     make_accessor(name)
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 908, in make_accessor
>     accessors[name](name)
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 838, in accessor_function
>     value = backendlib.load_function(BType, name)
> AttributeError: function/symbol 'legion_runtime_has_context' not found in library '<None>': dlsym(RTLD_DEFAULT, legion_runtime_has_context): symbol not found
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "a.py", line 1, in <module>
>     import legate.core
>   File "/Users/mpapadakis/legate.core/legate/core/__init__.py", line 20, in <module>
>     from legion_cffi import is_legion_python, ffi, lib as legion
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_cffi.py", line 35, in <module>
>     from legion_canonical_cffi import ffi, lib
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/legion_canonical_cffi.py", line 33, in <module>
>     lib = ffi.dlopen(libname)
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 150, in dlopen
>     lib, function_cache = _make_ffi_library(self, name, flags)
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 832, in _make_ffi_library
>     backendlib = _load_backend_lib(backend, libname, flags)
>   File "/Users/mpapadakis/opt/miniconda3/envs/test/lib/python3.8/site-packages/cffi/api.py", line 827, in _load_backend_lib
>     raise OSError(msg)
> OSError: cannot load library '/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib': dlopen(/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib, 2): Library not loaded: @rpath/liblegion.1.dylib
>   Referenced from: /Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib
>   Reason: image not found.  Additionally, ctypes.util.find_library() did not manage to locate a library called '/Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-install/lib/liblegion_canonical_python.dylib'
> ```
> 
> library RPATH information
> ```
> ~/legate.core/_skbuild/macosx--x86_64-3.8> otool -l ./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib | grep -A 2 RPATH
>           cmd LC_RPATH
>       cmdsize 64
>          path /Users/mpapadakis/opt/miniconda3/envs/test/lib (offset 12)
> --
>           cmd LC_RPATH
>       cmdsize 112
>          path /Users/mpapadakis/legate.core/_skbuild/macosx--x86_64-3.8/cmake-build/_deps/legion-build/lib (offset 12)
> ~/legate.core/_skbuild/macosx--x86_64-3.8> otool -L ./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib
> ./cmake-build/_deps/legion-build/lib/liblegion_canonical_python.dylib:
> 	@rpath/liblegion_canonical_python.dylib (compatibility version 0.0.0, current version 0.0.0)
> 	@rpath/liblegion.1.dylib (compatibility version 1.0.0, current version 0.0.0)
> 	@rpath/librealm.1.dylib (compatibility version 1.0.0, current version 0.0.0)
> 	@rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
> 	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
> ~/legate.core/_skbuild/macosx--x86_64-3.8> otool -l ./cmake-install/lib/liblegion_canonical_python.dylib | grep -A 2 RPATH
>           cmd LC_RPATH
>       cmdsize 64
>          path /Users/mpapadakis/opt/miniconda3/envs/test/lib (offset 12)
> ~/legate.core/_skbuild/macosx--x86_64-3.8> otool -L ./cmake-install/lib/liblegion_canonical_python.dylib
> ./cmake-install/lib/liblegion_canonical_python.dylib:
> 	@rpath/liblegion_canonical_python.dylib (compatibility version 0.0.0, current version 0.0.0)
> 	@rpath/liblegion.1.dylib (compatibility version 1.0.0, current version 0.0.0)
> 	@rpath/librealm.1.dylib (compatibility version 1.0.0, current version 0.0.0)
> 	@rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
> 	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
> </details>
> ```

You will need to export the following variables:
```
export PYTHONPATH=/scratch2/wwu/legate.coll/legion/bindings/python/build/lib:$PYTHONPATH
export LD_LIBRARY_PATH=/scratch2/wwu/legate.coll/_skbuild/linux-x86_64-3.9/cmake-install/lib:$LD_LIBRARY_PATH
```

legate/driver/command.py
bryevdv: Can we make a separate issue to discuss/resolve before 23.03? I do feel like saying "here are options you can't use when are not running legate" is confusing in and of itself, so I'd like to think on it a bit more.

bryevdv:
> We may want to reconsider enforcing a strict separation of legate vs user script arguments:

```
python <legate-args> prog.py <prog-args>
```
is a non-started, since `python` itself will try to consume any args up front:
```
(base) ❯ python --nsys foo.py a b c
unknown option --nsys
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
```
And TBH I don't think the legate case is much better, I think you'd need to resort to something very explicit, e.g.
```
legate <legate-args> -- prog.py <prog-args>
```

bryevdv:
> You will need to export the following variables 

There was a whole thread about this recently wasn't there? (A change on the realm side caused this) @manopapad what was the actual resolution here?

eddy16112:
> > You will need to export the following variables
> 
> There was a whole thread about this recently wasn't there? (A change on the realm side caused this) @manopapad what was the actual resolution here?

It is a different issue. That realm issue is regarding loading `libpython.so`. For this one, if we do a non-editable install, the `lib legion_canonical_python.so` along with `liblegion.so` and `librealm.so` will be installed to the condo env, so we do not need to export the `LD_LIBRARY_PATH`.

BTW, the `liblegion_canonical_python.so` is installed into https://gitlab.com/StanfordLegion/legion/-/blob/master/bindings/python/CMakeLists.txt#L55
https://gitlab.com/StanfordLegion/legion/-/blob/master/bindings/python/setup.py#L30

legate/driver/command.py
eddy16112: I am thinking about doing something like this:
```
$ legate --help
...
  --gdb                 run Legate inside gdb (default: False), not supported with default python interpreter. 
...
```
We add a sentence to flags that are not supported by canonical python mode.

bryevdv: I would be OK with that.

eddy16112:
> Doesn't necessarily need to happen on this PR, but we definitely need to have documentation before we advertise this feature.

I have created a new PR for updating the docs, https://github.com/nv-legate/legate.core/pull/541. It is still a draft.

legate/driver/command.py
eddy16112: Just want to clarify how should we name the canonical python mode? plain python, canonical python or default python? In the current readme, we call it "default python" https://github.com/nv-legate/legate.core/blob/branch-22.12/README.md?plain=1#L249-L250

bryevdv: Would defer to @manopapad but I will register my opinion that I don't think the phrase "canonical python" will carry an immediate, obvious meaning for typical python users.

manopapad: I agree with @bryevdv, "standard python interpreter" is probably the best

legate/driver/command.py
```diff
@@ -360,27 +367,23 @@ def cmd_eager_alloc(
...
-    cmd_nvprof,
-    cmd_nsys,
-    # Add memcheck right before the binary
-    cmd_memcheck,
-    # Now we're ready to build the actual command to run
-    cmd_legion,
+_CMD_PARTS_SHARED = (
     # This has to go before script name
     cmd_nocr,
     cmd_module,
```
manopapad, line 385: I think these two options should still be legion_python-only. `--nocr` is consumed only by `legion_python_main`, not `legion_canonical_python_main`. `cmd_module` sets `-m`, and it's too late to consume that at this point.

eddy16112: Actually, we do have `--nocr` in canonical python, https://gitlab.com/StanfordLegion/legion/-/blob/master/bindings/python/canonical_python.cc#L35-39

manopapad: OK, IIUC `cmd_module` is still not suitable for the canonical case (and correspondingly the `--module` and `--cprofile` flags are `legion_python`-only).

bryevdv: Yes if someone is using standard python, those would be `-m <module>` and `-m cProfile <module>` respectively

eddy16112: Yes, you are right. I have removed `--module` and `--cprofile`. I have also enabled `--verbose`  and  `--dry-run` for canonical python, please check this commit https://github.com/nv-legate/legate.core/pull/539/commits/92d8f3fb14b6c36c6da0c67c5561e00af25a843e

legate/driver/command.py
eddy16112: I have added `[legate-only, not supported with standard Python invocation]` to `legate --help` https://github.com/nv-legate/legate.core/pull/539/commits/92d8f3fb14b6c36c6da0c67c5561e00af25a843e

manopapad:
> You will need to export the following variables [to make canonical python work in an editable installation]:
> export PYTHONPATH=/scratch2/wwu/legate.coll/legion/bindings/python/build/lib:$PYTHONPATH
> export LD_LIBRARY_PATH=/scratch2/wwu/legate.coll/_skbuild/linux-x86_64-3.9/cmake-install/lib:$LD_LIBRARY_PATH

Just setting `LD_LIBRARY_PATH` worked for me. However, I'd prefer if if this worked automatically. With this Legion PR https://gitlab.com/StanfordLegion/legion/-/merge_requests/643 it works out of the box.

eddy16112:
I have implemented a prototype of loading args from env variables. 
```
diff --git a/legate/driver/main.py b/legate/driver/main.py
index bb02c7c..9a1783c 100644
--- a/legate/driver/main.py
+++ b/legate/driver/main.py
@@ -17,12 +17,18 @@
 """
 from __future__ import annotations
 
+import os
 from typing import Type, Union
 
 from . import CanonicalDriver, LegateDriver
 
 __all__ = ("legate_main",)
 
+CONFIG_ENV_CMD = {
+    "LEGATE_CONFIG_GPUS": "--gpus",
+    "LEGATE_CONFIG_CPUS": "--cpus",
+}
+
 
 def prepare_driver(
     argv: list[str],
@@ -33,8 +39,16 @@ def prepare_driver(
     from . import Config
     from .driver import print_verbose
 
+    new_argv = argv.copy()
+    for key, value in os.environ.items():
+        if key in CONFIG_ENV_CMD:
+            if CONFIG_ENV_CMD[key] not in new_argv:
+                new_argv.append(CONFIG_ENV_CMD[key])
+                new_argv.append(value)
+    print(new_argv)
+
     try:
-        config = Config(argv)
+        config = Config(new_argv)
     except Exception as e:
         print(error("Could not configure driver:\n"))
         raise e
```
Currently, the philosophy is that people can set args using env variables like `LEGATE_CONFIG_GPUS`, but if `--gpus` is passed through command line, we will take the value from command line instead of env.

manopapad:
> I have implemented a prototype of loading args from env variables.

I suggest we tackle this on a separate PR, after we get this one merged. There's a couple of options we want to consider (e.g. is there a single envvar vs multiple, do we support configuration files etc.)

bryevdv:
I agree with @manopapad let's tackle this separately, once there ialso an idea what the legion init will look like. 

Very roughly speculating, I think we will want to prefer a single optional env var that points to a config file, and then if that env var is present, the file contents are used to drive programmatic config APIs (or those APIs could be invoked directly)

eddy16112:
> > I have implemented a prototype of loading args from env variables.
> 
> I suggest we tackle this on a separate PR, after we get this one merged. There's a couple of options we want to consider (e.g. is there a single envvar vs multiple, do we support configuration files etc.)

Sounds good to me.

eddy16112:
> Just noting that things currently seems to work pretty well on an editable install on my workstation. I can even run most of the tests with plain `pytest` What's the shortlist of things needed before this can be merged?

I just need to clean up the print, and then after the branch 22.12 is closed and released, we can merge this PR.

bryevdv:
@eddy16112 do we need to wait for the release? This PR is already re-targeted to `branch-23.03` that was created a few days ago.

eddy16112:
> @eddy16112 do we need to wait for the release? This PR is already re-targeted to `branch-23.03` that was created a few days ago.

I thought we should wait for the release to avoid any conflicts, looks like as long as the PR is re-targeted, it is OK. @manopapad @bryevdv I have cleaned up prints, let me know if you have any other comments, otherwise, the PR is ready.

bryevdv:
> I thought we should wait for the release to avoid any conflicts,

I'd defer to @manopapad on this, but IMO if there are conflicts, there are conflicts either way, so it doesn't really matter the order :)

legate/core/__init__.py
```diff
@@ -17,9 +17,33 @@
 from ..rc import check_legion
 from ..util.args import parse_library_command_args
 
-check_legion()
```
manopapad, line 20: We can probably remove `legate/rc.py` altogether now.

eddy16112: removed.

legate/core/__init__.py
```diff
@@ -14,12 +14,33 @@
...
+    ] + sys.argv
+    driver = prepare_driver(sys_argv, CanonicalDriver)
+    if driver.dry_run:
+        sys.exit(0)
+    legate_argv = driver.cmd
+    legate_env = driver.env
+    # set env
+    for key, value in legate_env.items():
+        if driver.launcher.is_launcher_var(key):
+            os.environ[key] = value
```
manopapad, line 41: Did `os.environ = legate_env` not work?

legate/driver/driver.py
```diff
@@ -83,12 +83,13 @@ def custom_env_vars(self) -> set[str]:
...
-    def run(self) -> int:
-        """Run the Legate process.
+    @property
+    def dry_run(self) -> bool:
+        """Check verbose and dry run.
 
         Returns
         -------
-            int : process return code
+            bool : wether dry run
```
manopapad, line 92: Fix documentation message

legate/core/__init__.py
eddy16112: No, we can set set `os.environ` like this, its type is `<class 'os._Environ'>` even though it behaves like a dict. I can not find a way to convert a dict into a `os._Environ`.

legate/driver/args.py
```diff
@@ -188,7 +188,8 @@ def print_help(self, file: Optional[IO[str]] = None) -> None:
     dest="cprofile",
     action="store_true",
     required=False,
-    help="profile Python execution with the cprofile module",
+    help="profile Python execution with the cprofile module, "
+    "[legate-only, not supported with standard Python invocation]",
 )
```
manopapad, line 194: All of the arguments in the "Hardware binding" group in this file should also be marked as "legate-only".

legate/core/__init__.py
manopapad: You might be able to do `os.environ.update(legate_env)`.

legate/driver/args.py
eddy16112: good catch, fixed now. Thank you.

legate/driver/driver.py
eddy16112: fixed

legate/core/__init__.py
eddy16112: Yeah, it works. BTW, how do you find the API? I tried in docs.python.org, but did not find it.

manopapad: The answer I would like to give is that I went to the documentation for `os.environ`:

https://docs.python.org/3/library/os.html#os.environ

and I saw that it's documented as a "mapping":

https://docs.python.org/3/glossary.html#term-mapping

and I also know that it's mutable, so I looked at the documentation for the `MutableMapping` collection:

https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

and I saw that it adds the `update` method.

The truth is that I opened a python shell and typed `os.environ.` and then tab-tab'ed to see method completions, and picked what looked appropriate.

legate/driver/driver.py
```diff
@@ -83,12 +83,13 @@ def custom_env_vars(self) -> set[str]:
...
-    def run(self) -> int:
-        """Run the Legate process.
+    @property
+    def dry_run(self) -> bool:
+        """Check verbose and dry run.
 
         Returns
         -------
-            int : process return code
+            bool : wether dry run is enabled
```
manopapad, line 92:
```suggestion
            bool : whether dry run is enabled
```

manopapad (approved):
LGTM. Anything else @bryevdv?

bryevdv:
LGTM!

eddy16112:
@manopapad @bryevdv Thank you for the review. I am going to merge the PR now.

PR #549, opened 2023-01-30
Report version info in --verbose outupt

bryevdv:
This PR adds a `Versions:` section to the `-legate -verbose` output, that reports `legate` and (if found) `cunumeric` version information. Future work will add `legion` version information. 

This PR uses the recent work to make `cunumeric` top-level importable for purposes of scraping metadata from the package.

image

### DIFF
--- a/install.py
+++ b/install.py
@@ -413,7 +413,7 @@ def install(
     cmake_flags = cmd_env.get("CMAKE_ARGS", "").split(" ")
 
     if debug or verbose:
-        cmake_flags += ["--log-level=%s" % ("DEBUG" if debug else "VERBOSE")]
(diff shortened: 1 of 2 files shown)


bryevdv:
Not sure whats up  w/ pre-commit job, or for that matter, how to restart it in case that would help, since the "Details" link goes to a service site with a login. @marcinz any ideas?

marcinz:
I just tried it, and it shows me the output of the job without being logged in. I will send it to you on Slack.

bryevdv:
Right I see the output, just not any way to explicitly re-run things. I guess I could push new commits

manopapad (approved):

bryevdv:
Just noting that there is also an open MR against `legion` to add `legion_top.__version__` so we will also be able to report detailed `legion` version information as well, once that is merged.

PR #556, opened 2023-02-02
Add support for LEGATE_CONFIG

bryevdv:
This PR adds initial support for a `LEGATE_CONFIG` environment variable to pass options to the legate runtime (either with the `legate` driver or with standard python invocation). This PR also explicitly changes arg parsing to consider options *before* a `script.py` to be "legate config" and options *after* a `script.py` to be "user options".

Some specific invocations and their results are given below for comparison

#### 
```bash
LEGATE_CONFIG="--fbmem 2048 --gpus 1 --gpu-bind 1 --verbose --color" python cholesky.py

#  cholesky.py -lg:local 0 -ll:cpu 4 -ll:gpu 1 -cuda:skipbusy -ll:util 2 -ll:csize 4000 -ll:fsize 2048 -ll:zsize 32 -ll:networks none -level openmp=5,gpu=5 -lg:eager_alloc_percentage 50 -ucx:tls_host '^dc,ud'
```

#### with command line args

See "Gotchas" below

#### with env var
```bash
LEGATE_CONFIG="--fbmem 2048 --gpus 1 --gpu-bind 1 --verbose --color" legate cholesky.py

# bind.sh --launcher local --gpus 1 -- legion_python -ll:py 1 -lg:local 0 -ll:cpu 4 -ll:gpu 1 -cuda:skipbusy -ll:util 2 -ll:csize 4000 -ll:fsize 2048 -ll:zsize 32 -ll:networks none -level openmp=5,gpu=5 -lg:eager_alloc_percentage 50 -ucx:tls_host '^dc,ud' cholesky.py
```

#### with command line args
```bash
legate --fbmem 2048 --gpus 1 --gpu-bind 1 --verbose --color cholesky.py
 
# bind.sh --launcher local --gpus 1 -- legion_python -ll:py 1 -lg:local 0 -ll:cpu 4 -ll:gpu 1 -cuda:skipbusy -ll:util 2 -ll:csize 4000 -ll:fsize 2048 -ll:zsize 32 -ll:networks none -level openmp=5,gpu=5 -lg:eager_alloc_percentage 50 -ucx:tls_host '^dc,ud' cholesky.py
```

## Gotchas

The splitting on user script results in this behvior, which is just to be aware of:
```bash
legate --fbmem 2048 --gpus 1 --gpu-bind 1 cholesky.py  # legate sees --fbmem, etc

legate cholesky.py --fbmem 2048 --gpus 1 --gpu-bind 1  # legate does NOT see --fbmem, etc
```
This is definitely cleaner from a "stay out of the user script's way" POV but will need to be communicated widely. 

Another gotcha from splitting on user script. This does *not* work:
```
python cholesky.py --fbmem 2048 --gpus 1 --gpu-bind 1 --verbose --color
```
since all of the args are considered user args. I think this might be worth some consideration. 

## Tests

Note that all existing tests pass with these changes, since the implementation works by inserting values into "argv" early on. Appropriate tests to add would be to ensure that the presence   `LEGATE_CONFIG`  results in the expected modification of the `argv` that is passed to `Config`. I will add tests once the behavior is all agreed on. 

## Future work

Ultimately, it might be nice to have all CLLR configuration options pass via env vars. This would eliminate all possible contention with argument parsing that a user `script.py` would do and also would provide a clearer contract between the different library layers (each looks at its own env var, rather than all trying to gobble up `sys.argv` on their own with no coordination). 

This will require a fair somewhat more work to change the current machinery that builds up a command line to change to building up different env vars instead.

### DIFF
--- a/legate/driver/config.py
+++ b/legate/driver/config.py
@@ -176,8 +176,18 @@ class Config:
     """
 
     def __init__(self, argv: ArgList) -> None:
-        self.argv = argv
+        user_script = next((x for x in argv if x.endswith(".py")), None)
+        if user_script:
+            user_script_index = argv.index(user_script)
+            legate_argv = argv[:user_script_index]
+            user_argv = argv[user_script_index:][1:]
+        else:
+            legate_argv = argv
+            user_argv = []
 
+        self.argv = legate_argv
+
+        # extra might contain legion command line options for now
         args, extra = parser.parse_known_args(self.argv[1:])
 
         colors.ENABLED = args.color
@@ -185,12 +195,10 @@ class Config:
         # only saving this for help with testing
         self._args = args
 
-        self.user_script = next((x for x in extra if x.endswith(".py")), None)
+        self.user_script = user_script
 
-        user_opts = list(extra)
-        if self.user_script in user_opts:
-            user_opts.remove(self.user_script)
-        self.user_opts = tuple(user_opts)
+        # extra here might be legion or realm command line options
+        self.user_opts = tuple(user_argv) + tuple(extra)
 
         # these may modify the args, so apply before dataclass conversions
         self._fixup_nocr(args)
--- a/legate/core/__init__.py
+++ b/legate/core/__init__.py
@@ -14,8 +14,6 @@
 #
 from __future__ import annotations
 
-from ..util.args import parse_library_command_args
-
 from legion_cffi import is_legion_python, ffi, lib as legion
 
 if is_legion_python == False:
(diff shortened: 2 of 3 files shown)


bryevdv:
I went ahead and pushed the small change to the test runner to make sure it puts the test script after the legate args.

legate/tester/stages/test_stage.py
```diff
@@ -252,12 +252,12 @@ def run(
...
         cov_args = self.cov_args(config)
 
-        cmd = [str(config.legate_path)] + cov_args + [str(test_path)]
+        cmd = [str(config.legate_path)] + cov_args
 
         stage_args = self.args + self.shard_args(shard, config)
         file_args = self.file_args(test_file, config)
 
-        cmd += stage_args + file_args + config.extra_args
+        cmd += stage_args + [str(test_path)] + file_args + config.extra_args
```
manopapad, line 260: Do we need to be careful about what we put in file/custom args now? Since the legate args are only read before the script name?

legate/driver/config.py
```diff
@@ -176,21 +176,29 @@ class Config:
...
 
-        self.user_script = next((x for x in extra if x.endswith(".py")), None)
+        self.user_script = user_script
 
-        user_opts = list(extra)
-        if self.user_script in user_opts:
-            user_opts.remove(self.user_script)
-        self.user_opts = tuple(user_opts)
+        # extra here might be legion, realm, or cunumeric command line options
+        self.user_opts = tuple(user_argv) + tuple(extra)
```
manopapad, line 201:
```suggestion
        self.user_opts = tuple(extra) + tuple(user_argv)
```

Since the `extra` came before `user_argv` in the original command line.

legate/tester/stages/test_stage.py
bryevdv: My only previous experience and expectation for  `file_args` was as args that the example script itself would be parsing e.g. 

https://github.com/nv-legate/cunumeric/blob/12233e8021be8d490e1b87097037be2396e8ba3b/examples/cholesky.py#L35-L54

If we need to supply per-file CLLR args then yes we would need to explicitly provide those in another way (or else hold off on the before/after splitting)

robinw0928: @manopapad Currently, the below issue happened. The WAR is to rename /opt/conda/envs/legate/bin/coverage as coverage.py.

```
+LEGATE_TEST=1 legate /opt/conda/envs/legate/bin/coverage run -a --branch /opt/legate/cunumeric/tests/integration/test_take_along_axis.py -cunumeric:test --fbmem 4096 --gpus 1 --gpu-bind 0 -v
[FAIL] (GPU) tests/integration/test_take_along_axis.py (exit: 1)
   ERROR: usage: --capture=sys [options] [file_or_dir] [file_or_dir] [...]
   --capture=sys: error: unrecognized arguments: -a --branch
     inifile: /opt/legate/cunumeric/pyproject.toml
     rootdir: /opt/legate/cunumeric (edited)
```

bryevdv:
New PR to supersede this one coming shortly.

PR #559, opened 2023-02-03
Mappers should skip collective views with no suitable instance

magnatelee:
This PR fixes an intermittent segfault in `legate_select_sources`.

### DIFF
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -980,8 +980,10 @@ void BaseMapper::legate_select_sources(const MapperContext ctx,
   for (uint32_t idx = 0; idx < collective_sources.size(); idx++) {
     std::vector<PhysicalInstance> col_instances;
     collective_sources[idx].find_instances_nearest_memory(destination_memory, col_instances);
-    // if we can't find any nearest instance in this collective view, we move on to the next one
-    if (col_instances.empty()) continue;
+#ifdef DEBUG_LEGATE
+    // there must exist at least one instance in the collective view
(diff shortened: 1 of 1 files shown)


manopapad:
@lightsighter Shouldn't we expect the `find_instances_nearest_memory` API call to always return something (as long as a collective view has at least one thing in it)?

lightsighter:
> Shouldn't we expect the find_instances_nearest_memory API call to always return something (as long as a collective view has at least one thing in it)?

@manopapad It should, if you have a counter example then I would like to see it.

manopapad:
It sounds like @ipdemes is seeing such a case? @aschaffer was also seeing this happen while testing the quantile feature on 2 nodes.

lightsighter:
I would like to examine at least one of those use cases to see what is happening.

magnatelee:
I no longer see the issue that this PR meant to fix since I pulled the collective branch. I changed the code to assert collective sources being non-empty.

ipdemes:
@magnatelee : should we close it then? I am planning to check if TorchSWE still sees the error with the most recent legion's collective branch as well

magnatelee:
> @magnatelee : should we close it then? I am planning to check if TorchSWE still sees the error with the most recent legion's collective branch as well

I'm fine with closing it, but this PR would replace a segfault with a more informative assertion failure, so we could merge it for the sake of quality-of-life.

ipdemes (approved):

ipdemes:
@magnatelee : this makes sense.

PR #562, opened 2023-02-08
Args and Config

bryevdv:
This PR adds support for `LEGATE_CONFIG` as a mechanism for supplying legate or legion configuration options. Other points:

* The `legate` driver now recognizes a script and its arguments more generally by using all of the "remaining" args, rather than assuming a `.py` file extension somewhere.
* The `legate` driver still supports passing ***Legate*** (only) command line options before a script to run. The driver is an application, and it is reasonable for an application to process its own arguments since it can control when that processing happens. 

Broadly speaking, these two usages are what is now possible:

```
LEGATE_CONFIG="legate or legion opts" legate <legate opts only> <script> <script or legion opts>

LEGATE_CONFIG="legate or legion opts" python <script> <script opts>
```

### Legion command line config args

Technically, this will also "work":
```
python script.py -ll:cpu 2
```
The reason to discourage this (why it was not listed above) is that we have no control whether `script.py` attempts its own arg-parsing before or after `legate.core` is imported. If the script tries to parse args first, then it could abort due to "unkown" args

### Coverage 

The relative position of `cov_bin` was moved, to ensure legate args came before it. This seems the correct thing to do. The following also seems to work for me:
```
test.py --use=cpus,eager,cuda --cpus 2  --gpus 1 --debug ---cov-bin /opt/envs/dev310/bin/coverage
```
and results in an invocation like:
```
LEGATE_TEST=1 REALM_SYNTHETIC_CORE_MAP= legate --cpus 2 --cpu-bind 3,4,5,9,10,11 /opt/envs/dev310/bin/coverage run -a --branch /home/bryan/work/cunumeric/tests/integration/test_0d_store.py -cunumeric:test

```
@robinw0928 can you confirm whether this meets your expectations?

### Cunumeric args

The handling of `-cunumerc:test` was a but tricky. It cannot be included with the legate-only args (as it was previously) so it has been moved after the script in a fairly clunky way for now. @manopapad has proposed to remove all cunumeric command line processing and replace with env vars of some sort, and I *strongly* agree with this.  There are just too many combinations of import orders, command invocations, invocation methods, etc to contend with, and we can't even exert control over all of them if we wanted to try. The complexity added by having downstream libraries try to rummage through `sys.argv`, in addition to the upstream ones, is basically unmanageable. The good news is that there are only a handful of cunumeric args at present. 

### Legion env var config args

I have left out any conversions around `LEGION_DEFAULT_ARGS` and `REALM_DEFAULT_ARGS` from this PR. I actually think the current PR goes far enough to get things into a better place before a beta release. I would definitely like to include these env vars in documentation I intend to write, but I would like to stop there for now, and not introduce more changes and complexity to the driver before the beta release.

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -102,7 +102,7 @@ parser = _LegateArgumentParser(
 parser.add_argument(
     "command",
     nargs=REMAINDER,
-    help="A python script to run, plus any argumentsfor the script. "
+    help="A python script to run, plus any arguments for the script. "
     "Any arguments after the script will be passed to the script, i.e. "
     "NOT used as arguments to legate itself.",
 )
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -16,7 +16,7 @@
 #
 from __future__ import annotations
 
-from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
+from argparse import REMAINDER, ArgumentDefaultsHelpFormatter, ArgumentParser
 from typing import IO, Optional
 
 from ..util.shared_args import (
@@ -99,6 +99,13 @@ parser = _LegateArgumentParser(
     formatter_class=ArgumentDefaultsHelpFormatter,
 )
 
+parser.add_argument(
+    "command",
+    nargs=REMAINDER,
+    help="A python script to run, plus any argumentsfor the script. "
+    "Any arguments after the script will be passed to the script, i.e. "
+    "NOT used as arguments to legate itself.",
+)
 
 multi_node = parser.add_argument_group("Multi-node configuration")
 multi_node.add_argument(NODES.name, **NODES.kwargs)
--- a/legate/core/__init__.py
+++ b/legate/core/__init__.py
@@ -14,8 +14,6 @@
 #
 from __future__ import annotations
 
-from ..util.args import parse_library_command_args
-
 from legion_cffi import is_legion_python, ffi, lib as legion
 
 if is_legion_python == False:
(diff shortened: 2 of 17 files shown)


legate/driver/args.py
```diff
@@ -99,6 +99,13 @@ def print_help(self, file: Optional[IO[str]] = None) -> None:
     formatter_class=ArgumentDefaultsHelpFormatter,
 )
 
+parser.add_argument(
+    "command",
+    nargs=REMAINDER,
+    help="A python script to run, plus any argumentsfor the script. "
```
manopapad, line 105:
```suggestion
    help="A python script to run, plus any arguments for the script. "
```

manopapad (approved):
Everything I wanted for Christmas! I didn't see any handling of `LEGATE_CONFIG` on this PR specifically, I assume that's coming on a later one?

bryevdv:
@manopapad no I just missed pushing the last commit somehow. See https://github.com/nv-legate/legate.core/pull/562/commits/4fc43ee694cf9006741e756100e2dccf94a1bae8

manopapad (approved):
Tested it on local machine, works like a charm

bryevdv:
OK I will wait to see if @robinw0928 has a chance to check it out, but merge it tomorrow in any case.

robinw0928:
```
./test.py --use=cuda --cov-bin /home/robinw/miniconda3/envs/legate/bin/coverage --files tests/integration/test_take_along_axis.py --debug --verbose
```
I have verified that this command PASS on my local Linux.
The coverage part meets my expectation.

robinw0928 (approved):

PR #563, opened 2023-02-09
API reference

magnatelee:
This PR contains initial changes to add an API reference. This PR is intended to add doc strings only to the user facing APIs; internal classes and functions will be documented in a follow-on PR.

~~A question to folks who know about Doxygen better (maybe @trxcllnt @jjwilke?): what's the flag I should set to get qualified header paths in summary pages? This is what I'm seeing now:~~ Resolved!


jjwilke:
```
diff --git a/legate_core_cpp.cmake b/legate_core_cpp.cmake
index e22a4a3..8ca9f85 100644
--- a/legate_core_cpp.cmake
+++ b/legate_core_cpp.cmake
@@ -331,6 +331,7 @@ if (legate_core_BUILD_DOCS)
     set(DOXYGEN_EXTENSION_MAPPING cu=C++ cuh=C++)
     set(DOXYGEN_HIDE_UNDOC_MEMBERS YES)
     set(DOXYGEN_HIDE_UNDOC_CLASSES YES)
+    set(DOXYGEN_STRIP_FROM_INC_PATH ${CMAKE_SOURCE_DIR})
     doxygen_add_docs("doxygen_legate" ALL
       ${legate_core_DOC_SOURCES}
       COMMENT "Custom command for building Doxygen docs."
```

Or `${CMAKE_SOURCE_DIR}/src`, etc, depending on what you want displayed.

magnatelee:
> ```
> diff --git a/legate_core_cpp.cmake b/legate_core_cpp.cmake
> index e22a4a3..8ca9f85 100644
> --- a/legate_core_cpp.cmake
> +++ b/legate_core_cpp.cmake
> @@ -331,6 +331,7 @@ if (legate_core_BUILD_DOCS)
>      set(DOXYGEN_EXTENSION_MAPPING cu=C++ cuh=C++)
>      set(DOXYGEN_HIDE_UNDOC_MEMBERS YES)
>      set(DOXYGEN_HIDE_UNDOC_CLASSES YES)
> +    set(DOXYGEN_STRIP_FROM_INC_PATH ${CMAKE_SOURCE_DIR})
>      doxygen_add_docs("doxygen_legate" ALL
>        ${legate_core_DOC_SOURCES}
>        COMMENT "Custom command for building Doxygen docs."
> ```
> 
> 
>     
>   
> 
> Or `${CMAKE_SOURCE_DIR}/src`, etc, depending on what you want displayed.
> 
> 
> The default behavior if the `STRIP_` variable is empty is to only show the header name. If you give it an initial path, it puts the full path minus the stripped part.

That worked. Thanks!

magnatelee:
@bryevdv @jjwilke I haven't found a good way to link the doxygen doc in the sphinx doc. In case you have a good idea, I'm all ears.

bryevdv:
I think you'll just need to use an external link to wherever the doxygen docs end up located.

legate/core/operation.py
```diff
@@ -220,6 +270,23 @@ def get_name(self) -> str:
...
+        """
+        Creates a partition symbol for the store
+
+        Parameters
+        ----------
+        store : Store
+            Store to associate the partition symbol with
+        disjoint : bool, optional
+            ``True`` (by default) means the partition must be disjoint
+        complete : bool, optional
```
jjwilke, line 282: Question rather than comment on the PR - but when would you ever want an incomplete partition? I get that disjoint = false might save you communication if you replicate some points, but seems weird to me that you would be okay with a partition that doesn't cover every point.  

Unless I'm misunderstanding complete.... and if I am, maybe we could clarify what complete means : )

legate/core/operation.py
```diff
@@ -253,16 +320,53 @@ def __init__(
...
+        ----------
+        side_effect : bool
+            A new boolean value indicating whether the task has side effects
+        """
         self._side_effect = side_effect
 
     @property
     def concurrent(self) -> bool:
+        """
+        Inidicates whether the task needs a concurrent task launch.
```
jjwilke, line 348: nit: Indicates

legate/core/operation.py
```diff
@@ -470,9 +611,38 @@ def get_requirement(
         tag = self.get_tag(strategy, part_symb)
         return req, tag, store_part
 
+
+class AutoTask(AutoOperation, Task):
```
jjwilke, line 615: class documentation for AutoTask?

legate/core/operation.py
```diff
@@ -470,9 +611,38 @@ def get_requirement(
...
+        )
+        self._reusable_stores: list[Tuple[Store, PartSym]] = []
+        self._reuse_map: dict[int, Store] = {}
+
     def add_input(
         self, store: Store, partition: Optional[PartSym] = None
     ) -> None:
+        """
+        Adds a store as input to the task
+
```
jjwilke, line 637: Do we need the docstring in the subclass? @bryevdv I think Sphinx automatically adds the docstring if the parent class provides it? I'm not sure what the best way is to document abstract methods so that we don't have to repeat identical docstrings in each subclass.

legate/core/store.py
```diff
@@ -915,6 +953,14 @@ def shape(self) -> Shape:
 
     @property
     def ndim(self) -> int:
+        """
+        Returns the dimension of the store.
```
jjwilke, line 957: nit: Returns the number of dimensions in the store

legate/core/store.py
```diff
@@ -915,6 +953,14 @@ def shape(self) -> Shape:
...
+
+        Returns
+        -------
+        int
+          The store's dimension
```
jjwilke, line 962: nit: The number of dimensions

legate/core/store.py
```diff
@@ -923,12 +969,25 @@ def ndim(self) -> int:
 
     @property
     def size(self) -> int:
+        """
+        Returns the size of the store.
```
jjwilke, line 973: Byte size or shape volume? Might want to clarify.

legate/core/store.py
```diff
@@ -1038,6 +1131,46 @@ def __repr__(self) -> str:
...
     def promote(self, extra_dim: int, dim_size: int = 1) -> Store:
+        """
+        Adds an extra dimension to the store. Value of ``extra_dim`` decides
+        where a new dimension should be added, and each dimension `i`, where
+        `i` >= ``extra_dim``, is mapped to dimension `i+1` in a returned store.
+        A returned store provides a view to the input store where the values
+        are broadcasted along the new dimension.
+
+        For example, for a 1D store ``A`` contains ``[1, 2, 3]``,
+        ``A.promote(0, 2)`` yields:
```
jjwilke, line 1142: might want to say "yields an array equivalent to"

Not specific to this PR, but the semantics here are confusing to me.  If I promote an array, I have a view which means modifying the promoted store would modify the original values.  But `A[0][0] = 1; A[0][1] = 5` are both views onto the same value?

legate/core/store.py
```diff
@@ -1038,6 +1131,46 @@ def __repr__(self) -> str:
...
+
+        Returns
+        -------
+        Store
+            A new store with an extra dimension
+
+        Raises
+        ------
+        ValueError
+            If ``extra_dim`` is negative or greater than the store's dimension
```
jjwilke, line 1172: extra_dim can be negative, yeah? 

```
ValueError
            If ``extra_dim`` or `-extra_dim` is greater than the store's dimension
```

legate/core/store.py
```diff
@@ -1060,6 +1193,35 @@ def promote(self, extra_dim: int, dim_size: int = 1) -> Store:
...
+
+        Returns
+        -------
+        Store
+            A new store with one fewer dimension
+
+        Raises
+        ------
+        ValueError
+            If any of the following happens: 1) ``dim`` is negative; 2) ``dim``
```
jjwilke, line 1221: same comment as above. `dim` can be negative, yeah?

legate/core/store.py
```diff
@@ -1097,6 +1259,68 @@ def project(self, dim: int, index: int) -> Store:
...
+
+            [[1, 2],
+             [4, 5],
+             [7, 8]]
+
+        Finally, chained slicing calls
+
+        ::
+
+            A.slice(0, slice(1, None)).A.slice(1, slice(None, 2))
```
jjwilke, line 1293: Should this be:

```
A.slice(0, slice(1, None)).slice(1, slice(None, 2))
```

? without the extra `A` in the middle?

jjwilke (changes requested):
Mostly nits and typos.  Thanks so much for all the huge effort on this. Will be massively helpful to users - and even helped clarify a few things for me reading the PR.

docs/legate/core/source/api/operation.rst
```diff
@@ -0,0 +1,117 @@
...
+Manually Parallelized Tasks
+---------------------------
+
+In some occassions, tasks are unnatural or even impossible to write in the
+auto-parallelized style. For those occassions, Legate provides explicit control
+on how tasks are parallelized via ``ManualTask``. Each manual task requires the
+caller to provide a `launch domain` that determines the degree of parallelism
+and also names task instances initiaed by the task. Direct store arguments to a
+manual task are assumed to be replicated across task instances, and it's the
+developer's responsibility to partition stores. Mapping between points in the
```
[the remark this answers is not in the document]
jjwilke: Follow up - I would delay the hover references to another PR. Not at all necessary here.

legate/core/operation.py
jjwilke: Not necessary for this PR. I would worry about this later.

docs/legate/core/source/api/store.rst
```diff
@@ -0,0 +1,123 @@
...
+
+
+StorePartition
+==============
+
+A ``StorePartition`` is an object that represents a partitioned state of a
+store. A store partition is a name of a collection of `sub-stores`, each of
+which contains to a subset of elements in the store. Sub-stores in a store
+partition are uniquely identified by their `colors`, and a set of all colors
+of a given store partition is called a `color space`.
```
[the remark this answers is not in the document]
jjwilke: As above. Just a feature request for now. I wouldn't bother on this PR.

src/core/runtime/context.h
```diff
@@ -37,9 +42,22 @@ class LegateMapper;
 class Store;
 class Scalar;
 
+/**
+ * @ingroup runtime
+ * @brief POD for library configuration.
+ */
 struct ResourceConfig {
+  /**
+   * @brief Maximum number of tasks
```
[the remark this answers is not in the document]
magnatelee: > maximum number of tasks?

"Task" is such an overloaded term in Legate... in this context, a number of tasks means a number of distinct task implementations. the runtime needs to know how many tasks there will be in a library to assign a non-overlapping range of ids.

> that a LibraryContext can create at any given time?

that a library of the library context can register to the runtime at any given time, including registration time and execution time.

src/core/runtime/context.h
```diff
@@ -37,9 +42,22 @@ class LegateMapper;
...
+   */
   int64_t max_tasks{1000000};
+  /**
+   * @brief Maximum number of mappers
```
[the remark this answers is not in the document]
magnatelee: > that a LibraryContext can create at any given time?

that a LibraryContext can register at any given time. that said, we won't expose dynamic mapper registration to client libraries.

src/core/runtime/context.h
```diff
@@ -37,9 +42,22 @@ class LegateMapper;
...
+   */
   int64_t max_mappers{1};
+  /**
+   * @brief Maximum number of reduction operators
```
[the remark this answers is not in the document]
magnatelee: > Maximum number of... custom reduction ops? 

yes

> So the built-in ops are still valid even if this is zero?

yes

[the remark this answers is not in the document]
magnatelee: fyi, the current interface for custom reduction ops is not clean and needs some refactoring, hopefully after I unify the type system.

legate/core/operation.py
magnatelee: Partitions in legate and legion are not mathematical partitions, but a collection of subsets of a set. so the union of those subsets may not cover the original set. the runtime can and does perform a few optimizations when it knows that the partition used in updating a store is complete, so if a partition that you said is complete was actually incomplete, you would see correctness issues. lying in the other direction is actually safe, but will cause unnecessary runtime overhead. 

Incomplete partitions are useful when you know you only need to make accesses to a subset of data points, such as stencil accesses. A stencil code often creates two partitions of the same grid data, one that is complete and disjoint, used for updating the grid, and one that only selects elements used in stencil accesses. Doing this helps minimize the amount of data movement.

legate/core/operation.py
magnatelee: > Do we need the docstring in the subclass? 

No we don't in general, but look at their signatures. The methods are actually different and that was intentional, as the methods admit different types of arguments.

legate/core/store.py
magnatelee: > might want to say "yields an array equivalent to"

Funny that I actually wrote like that first and shortened it, as I thought the reader wouldn't notice the subtlety. let me put that back.

> If I promote an array, I have a view which means modifying the promoted store would modify the original values. But A[0][0] = 1; A[0][1] = 5 are both views onto the same value?

aliased updates have undefined semantics. what value the original store will get depends on the evaluation order. that's something I should clarify in a follow-on programming model doc. (we have a similar issue with indirect copies.)

docs/legate/core/source/api/operation.rst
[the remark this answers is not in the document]
magnatelee: Note that we can't use `hoverxref` unless we host our API reference on Read the Docs: https://sphinx-hoverxref.readthedocs.io/en/latest/installation.html. I instead put a link to the page where the definitions can be found. please take a look.
image

docs/legate/core/source/api/operation.rst
```diff
@@ -0,0 +1,117 @@
...
+launch domain and those in the color space of a store partition is identity by
```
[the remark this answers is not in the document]
magnatelee: it is technically an identity mapping. legate inserts a delinearizing functor for a mismatch is a temporary hack and should be gone once we revise the projection functor expression language to support linearization and delinearization.

legate/core/context.py
```diff
@@ -226,6 +252,25 @@ def get_sharding_id(self, shard_id: int) -> int:
     def get_tunable(
         self, tunable_id: int, dtype: DataType, mapper_id: int = 0
     ) -> npt.NDArray[Any]:
+        """
+        Queries a tunable parameter to the mapper.
+
```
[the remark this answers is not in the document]
magnatelee: I'll defer the description on tunable parameters to a follow-on programming model doc. there are two sides of tunable parameters clients need to fill in and it may not be obvious without enough context why we separate these.

docs/legate/core/source/api/operation.rst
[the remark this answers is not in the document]
jjwilke: LGTM

legate/core/operation.py
jjwilke: Stencils, duh. Thank you.  LGTM.

legate/core/operation.py
jjwilke: Ah, okay.

jjwilke (approved):
LGTM!

PR #565, opened 2023-02-10
Update Build.md to add the missing dependency, rust

natsukium:
Rust was added to the build tools in PR #535, but BUILD.md has not been updated yet.

Signed-off-by: OTABI Tomoya <tomoya.otabi@gmail.com>


manopapad:
Good catch, thank you!

PR #567, opened 2023-02-13
Move RC command line args to env var / settings

bryevdv:
This PR removes Legate *runtime* command line arguments (i.e. the ones parsed by the *library*&mdash;not the driver) and replaces them with settings / env vars to accomplish the same. The table below summarizes:

| arg  (removed)| setting | env var|
|-----|----|---|
| `-legate:consensus` | `settings.consensus` | `LEGATE_CONSENSUS`|
| `-legate:cycle-check` | `settings.cycle_check` | `LEGATE_CYCLE_CHECK`|
| `-legate:future-leak-check` | `settings.future_leak_check` | `LEGATE_FUTURE_LEAK_CHECK`|

Once this PR is merged I will make a similar PR to consolidate all the varied manual and explicit env var and command line arg handling in `cunumeric` using the `Settings` class. After that I will also remove the legate util functions for parsing command line args at runtime altogether. The eventual goal to to be able to auto-document all settings and env vars in the ref guides.


legate/util/settings.py
```diff
@@ -0,0 +1,307 @@
+# Copyright 2021-2022 NVIDIA Corporation
```
manopapad, line 1: 2023

legate/util/settings.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def convert_bool(value: bool | str) -> bool:
+    """Convert a string to True or False.
+
+    If a boolean is passed in, it is returned as-is. Otherwise the function
+    maps the following strings, ignoring case:
+
+    * "yes", "1", "on" -> True
+    * "no", "0", "off" -> False
```
manopapad, line 109: Also "true" and "false"

legate/util/settings.py
```diff
@@ -0,0 +1,307 @@
...
+
+    Raises:
+        ValueError
+
+    """
+    if isinstance(value, bool):
+        return value
+
+    val = value.lower()
+    if val in ["yes", "1", "on", "true", "True"]:
```
manopapad, line 126: No need to check against "True", since you called `lower()`

legate/util/settings.py
```diff
@@ -0,0 +1,307 @@
...
+
+    The following methods are searched in order for the setting:
+
+    4. immediately supplied values
+    3. previously user-set values (e.g. set from command line)
+    2. environment variable
+    1. local defaults
+    0. global defaults
+
+    If a value cannot be determined, a ValueError is raised.
```
manopapad, line 170:
```suggestion
    If a value cannot be determined, a RuntimeError is raised.
```

bryevdv: c5c0796

legate/util/settings.py
bryevdv: c5c0796

legate/util/settings.py
bryevdv: c5c0796

legate/util/settings.py
bryevdv: c5c0796 (and other files)

legate/util/settings.py
```diff
@@ -0,0 +1,307 @@
...
+        return self._help
+
+    @property
+    def convert_type(self) -> str:
+        if self._convert is convert_str:
+            return "String"
+        if self._convert is convert_bool:
+            return "Bool"
+        if self._convert is convert_str_seq:
+            return "List[String]"
```
manopapad, line 302: It looks like `convert_str_seq` returns `tuple[str, ...]`.

legate/settings.py
```diff
@@ -0,0 +1,61 @@
...
+        """,
+    )
+
+    future_leak_check: PrioritizedSetting[bool] = PrioritizedSetting(
+        "future_leak_check",
+        "LEGATE_FUTURE_LEAK_CHECK",
+        default=False,
+        convert=convert_bool,
+        help="""
+        Whether to check for reference cycles keeping Future/FutureMap objects.
```
manopapad, line 56: Some help text missing:

Whether to check for reference cycles keeping Future/FutureMap objects alive after Legate runtime exit (developer option). Such leaks can result in Legion runtime shutdown hangs.

manopapad (approved):

legate/util/settings.py
bryevdv: f83dfba

legate/settings.py
bryevdv: f83dfba

PR #571, opened 2023-02-15
Build helper functions for legate projects, legate-hello example

jjwilke:
With this set of helper functions, the CMakeLists.txt for a downstream project like cunumeric would now look like:

src/CMakeLists.txt:

```cmake
add_library(hello
  hello_world.cc
  hello_world.h)
target_link_libraries(hello PRIVATE legate::core)
....
```

CMakeLists.txt:

```cmake
cmake_minimum_required(VERSION 3.24.0 FATAL_ERROR)

project(hello VERSION 1.0 LANGUAGES C CXX)

find_package(legate_core REQUIRED)
legate_add_cpp_subdirectory(src hello EXPORT hello-export)
legate_add_cffi(${CMAKE_SOURCE_DIR}/src/hello_world.h)
legate_default_python_install(hello EXPORT hello-export)
```


jjwilke:
Hello world example being developed here:

https://github.com/jjwilke/legate-hello-world

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,184 @@
...
+        target = "@target"
+        if exists(join(libdir, f"lib{target}{so_ext}")):
+            return libdir
+        return None
+
+    return (
+        find_libcunumeric(join(cn_path, "build", "lib")) or
+        find_libcunumeric(join(dirname(dirname(dirname(cn_path))), "lib")) or
+        find_libcunumeric(join(dirname(dirname(sys.executable)), "lib")) or
+        ""
```
trxcllnt, line 97: typos?

```suggestion
    return (
        find_lib(join(cn_path, "build", "lib")) or
        find_lib(join(dirname(dirname(dirname(cn_path))), "lib")) or
        find_lib(join(dirname(dirname(sys.executable)), "lib")) or
        ""
```

jjwilke: Oops, yep, thank you.

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,184 @@
+macro(legate_include_rapids)
+  if (NOT _LEGATE_HAS_RAPIDS)
+    if(NOT EXISTS ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
+      file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-22.08/RAPIDS.cmake
```
trxcllnt, line 4: Should probably update to (at least) `branch-22.10`, which added the ability for the user [to control](https://github.com/rapidsai/rapids-cmake/blob/branch-22.10/RAPIDS.cmake#L21-L24) the _actual_ version of `rapids-cmake` that is downloaded externally:

```shell
cmake -S . -B build -Drapids-cmake-version=23.04
```


```suggestion
      file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-23.02/RAPIDS.cmake
```

trxcllnt:
@robertmaynard I'd also appreciate your thoughts on this PR.

cmake/legate_helper_functions.cmake
robertmaynard: Also place the file in `LEGATE_RAPIDS.cmake` so that you don't get clobbered by any other dep / consumer that uses rapids and writes to `RAPIDS.cmake`

trxcllnt:
@jjwilke can we also add the `LANGUAGES ${ENABLED_LANGUAGES}` option to the `rapids_export()` calls in `legate_core_{cpp,python}.cmake`?

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,384 @@
...
+    set(_LEGATE_HAS_RAPIDS ON)
+  endif()
+endmacro()
+
+function(legate_default_cpp_install target)
+  set(options)
+  set(one_value_args EXPORT)
+  set(multi_value_args)
+  cmake_parse_arguments(
+    OPT
```
robertmaynard, line 22: I would use something like `LEGATE_OPT` instead of `OPT`. I have seen callers code activate bad code paths because they are using variables with the same name and you get side-effects.

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,384 @@
...
+
+  legate_include_rapids()
+
+  rapids_cmake_install_lib_dir(lib_dir)
+
+  install(TARGETS ${target}
+          DESTINATION ${lib_dir}
+	  EXPORT ${OPT_EXPORT})
+
+  set(code_string "set(legate_core_DIR ${legate_core_DIR})")
```
robertmaynard, line 41: Do you want to generate relocatable packages? This will bind this package to where-ever legate core is on the local machine.

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,384 @@
...
+    BUILD ${target}
+    EXPORT_SET ${OPT_EXPORT}
+    GLOBAL_TARGETS ${target}
+    NAMESPACE legate::
+    FINAL_CODE_BLOCK code_string
+  )
+endfunction()
+
+function(legate_add_cffi header target)
+execute_process(
```
robertmaynard, line 62: Maybe want to error out if `CMAKE_C_COMPILER` is not set. That way users that have only enabled CXX or  CUDA don't get weird messages about unable to execute command `-E -P`

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,384 @@
...
+  if (SKBUILD)
+    if (NOT DEFINED ${target}_ROOT)
+      set(${target}_ROOT ${CMAKE_SOURCE_DIR}/build)
+    endif()
+    rapids_find_package(${target} CONFIG
+      GLOBAL_TARGETS legate::${target}
+      BUILD_EXPORT_SET ${OPT_EXPORT}
+      INSTALL_EXPORT_SET ${OPT_EXPORT})
+    if (NOT ${target}_FOUND)
+      add_subdirectory(${dir} ${CMAKE_SOURCE_DIR}/build)
```
robertmaynard, line 184: I would use something like `${CMAKE_SOURCE_DIR}/legate_${${target}}_build` instead of `${CMAKE_SOURCE_DIR}/build`.

The `build` directory could already be used by a target named that, or a user might want to call `legate_add_cpp_subdirectory` multiple time and this wouldn't work

trxcllnt: @robertmaynard that's actually the point of this line, since this is only executed when running via scikit-build

1. If the user configures C++ independently, they can do it into `./build`
2. If the user lets scikit-build configure the C++ lib as a side-effect of building the Python lib, the C++ libs still end up in `./build` (where normally they'd be under something like `_skbuild/<arch>/cmake-build`)

This stability helps with the runtime library loading we do when dev'ing locally w/ `pip install --editable .`

cmake/legate_helper_functions.cmake
jjwilke: Good call, thank you

cmake/legate_helper_functions.cmake
jjwilke: I split the difference since it's a valid point that this only supports a single source directory. I put everything in:
`${CMAKE_SOURCE_DIR}/build/legate_${target}` - which I think is okay? @robertmaynard 

This still makes the libraries easy to find.

cmake/legate_helper_functions.cmake
jjwilke: Yeah, I was trying to be too clever in the downstream app. I removed this.

cmake/legate_helper_functions.cmake
trxcllnt: Maybe we should make it configurable, or do something like `build/${cmake_build_type}`? So that way at least it's `build/Debug` and `build/Release`?

robertmaynard: When using `add_subdirectory(${dir} ${CMAKE_SOURCE_DIR}/build)` you are stating that CMake should parse the `CMakeLists.txt` in `${dir}` and place any build information in `${CMAKE_SOURCE_DIR}/build`. This has the following issues:

1. You are placing build artifacts in the soure tree. The path should start with `CMAKE_BUILD_DIR`.
2. The second argument must be unique across a CMake project. So `${CMAKE_BUILD_DIR}/legate_${target}` would be sufficient. So the following logic doesn't work currently:
```cmake
legate_add_cpp_subdirectory(legate_plugin1 hello EXPORT hello-export)
legate_add_cpp_subdirectory(src hello EXPORT hello-export)
```

Since `legate_add_cpp_subdirectory` is executed by the users project we have no idea what source tree layout they have and shouldn't try to write to `CMAKE_SOURCE_DIR`.

jjwilke: Thanks all for the suggestions. Silly mistakes on my part. I believe the latest version correctly puts things into `CMAKE_BINARY_DIR` into unique folders.

PTAL. @robertmaynard @trxcllnt

cmake/legate_helper_functions.cmake
```diff
@@ -0,0 +1,395 @@
...
+  legate_include_rapids()
+
+  rapids_find_package(legate_core CONFIG
+          GLOBAL_TARGETS legate::core
+          BUILD_EXPORT_SET ${LEGATE_OPT_EXPORT}
+          INSTALL_EXPORT_SET ${LEGATE_OPT_EXPORT})
+
+  if (SKBUILD)
+    if (NOT DEFINED ${target}_ROOT)
+      set(${target}_ROOT ${CMAKE_SOURCE_DIR}/build)
```
robertmaynard, line 188:
```suggestion
      set(${target}_ROOT ${CMAKE_BINARY_DIR}/legate_${target})
```
This needs to be the same as the add_subdirectory call

cmake/legate_helper_functions.cmake
robertmaynard: Updated version looks good to me

cmake/legate_helper_functions.cmake
jjwilke: This is a a shortcut so people's `pip install` will "just work" if they follow a standard directory naming scheme. When this runs, `CMAKE_BINARY_DIR` will be `_skbuild` rather than the C++ build directory.

jjwilke: It's admittedly hacky...

manopapad (approved):

jjwilke:
@trxcllnt @manopapad Sorry everybody... PTAL. With the recent breakage due to PR #581 I decided the hello-world examples should really be maintained in the legate-core repo with build testing in CI to make sure things don't get broken.

The current setup abuses `find_package` to "find" Legate in its own build tree when `legate_core_EXAMPLE_BUILD_TESTS` is `ON`.

tests/examples/hello/src/iota.cc
```diff
@@ -0,0 +1,28 @@
+#include "hello_world.h"
+#include "legate_library.h"
+
+namespace hello {
+
+class IotaTask : public Task<IotaTask, IOTA> {
+public:
+  static void cpu_variant(legate::TaskContext &context) {
+    auto &output = context.outputs()[0];
```
manopapad, line 9: IMHO it would be better for didactic purposes if we spelled out the types here.

tests/examples/hello/src/iota.cc
```diff
@@ -0,0 +1,28 @@
...
+    auto output_shape = output.shape<1>();
```
manopapad, line 10: I suggest adding assertions here (and maybe explain in a comment), to show that the dimensions and type of a Store are dynamic, and what we "receive" on the task side (in C++) depends on what the calling code (in python) passes in. Here the rest of the task can only work on 1d float arrays, so we check that the calling code indeed passed one such array.

```
assert(output.code() == FLOAT_LT);
assert(output.dim() == 1);
```

tests/examples/hello/src/iota.cc
```diff
@@ -0,0 +1,28 @@
...
+    auto out = output.write_accessor<float, 1>();
+    auto n = output_shape.volume();
+    for (size_t i = 0; i < n; ++i) {
+      out[i] = i + 1;
+    }
```
manopapad, line 15: Will this work on multiple processors/nodes? Each point task may receive a subset of the operation's "global" domain, and the Accessor class works with "global" indices, as opposed to local ones. I.e. I expect the code would need to look something like this.

```
for (int64_t i = output_shape.lo; i <= output_shape.hi; ++i) {
  out[i] = i + 1;
}
```

But maybe I'm missing something.

tests/examples/hello/src/iota.cc
```diff
@@ -0,0 +1,28 @@
+#include "hello_world.h"
+#include "legate_library.h"
+
```
manopapad, line 3: I suggest adding a logger (that all tasks in the hello example use), which log information that would help the reader understand what's going on:

```
log.info() << "Task iota handling points " << output_shape.lo << " to " << output_shape.hi;
```

tests/examples/hello/src/square.cc
```diff
@@ -0,0 +1,33 @@
...
+public:
+  static void cpu_variant(legate::TaskContext &context) {
+    auto &output = context.outputs()[0];
+    auto output_shape = output.shape<1>();
+    auto out = output.write_accessor<float, 1>();
+
+    auto &input = context.inputs()[0];
+    auto input_shape = input.shape<1>(); // should be a 1-Dim array
+    auto in = input.read_accessor<float, 1>();
+
```
manopapad, line 16: I suggest adding an assert here, which verifies that `add_alignment` worked as expected; i.e. every point task will received aligned spans over the two arrays.
```
assert(input_shape == output_shape);
```

tests/examples/hello/src/sum.cc
```diff
@@ -0,0 +1,34 @@
...
+    auto in = input.read_accessor<float, 1>();
+
+    auto n = input_shape.volume();
+    float total = 0;
+    for (size_t i = 0; i < n; ++i) {
+      total += in[i];
+    }
+
+    using Reduce = Legion::SumReduction<float>;
+    auto sum = context.reductions()[0].reduce_accessor<Reduce, true, 1>();
```
manopapad, line 20: Suggest adding some comments to explain what's going on.

"The task launch as a whole will return a single value (`Store` of size 1) to the caller. However, each point task gets a separate `Store` of the same size as the result, to reduce into. This "local accumulator" will be initialized by the runtime, and all we need to do is call `.reduce()` to add our local contribution. After all point tasks return, the runtime will make sure to combine all their buffers into the single final result."

manopapad, line 20: Adding type & size checks here will make it clear this is a singleton `Store` we're reducing into:
```
Store& output = context.reductions()[0];
assert(output.code() == FLOAT_LT);
assert(output.dim() == 1);
assert(output.shape() == Rect<1>(0,0));
```

tests/examples/hello/src/iota.cc
manopapad: It would be good to add some comments explaining this "local" vs "global" indexing.

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+
+class HelloOpCode(IntEnum):
+    HELLO_WORLD = user_lib.cffi.HELLO_WORLD
+    SUM = user_lib.cffi.SUM
+    SQUARE = user_lib.cffi.SQUARE
+    IOTA = user_lib.cffi.IOTA
+
+
+def print_hello(message: str) -> None:
+    """Create a Legate to task to print a message
```
manopapad, line 21:
```suggestion
    """Create a Legate task to print a message
```

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+        message (str): The message to print
+    """
+    task = user_context.create_auto_task(HelloOpCode.HELLO_WORLD)
+    task.add_scalar_arg(message, types.string)
+    task.execute()
+
+
+def print_hellos(message: str, n: int) -> None:
+    """Create a Legate to task launch to print a message
+       in n replicas of the task
```
manopapad, line 33:
```suggestion
    """Create a Legate task launch to print a message n times,
       using n replicas of the task
```

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Args:
+        message (str): The message to print
+        n (int): The number of times to print
+    """
+    launch_domain = Rect(lo=[0], hi=[n], exclusive=True)
```
manopapad, line 39:
```suggestion
    launch_domain = Rect(lo=[0], hi=[n])
```
`exclusive` is default I think

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+    """Extracts a Python scalar value from a Legate store
+       encapsulating a single scalar
+
+    Args:
+        input (Store): The Legate store encapsulating a scalar
+
+    Returns:
+        float: A Python scalar
+    """
+    buf = input.storage.get_buffer(np.float32().itemsize)
```
manopapad, line 76: I would note that this operation will block, waiting for the asynchronous task that will produce this value.

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+    result = np.frombuffer(buf, dtype=np.float32, count=1)
+    return float(result[0])
+
+
+def zero() -> Store:
+    """Create a Legates store representing a single zero scalar
```
manopapad, line 82:
```suggestion
    """Create a Legate store representing a single zero scalar
```

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+    """Sums a 1-D array into a single scalar
+
+    Args:
+        input (Any): A Legate store or any object implementing
+                     the Legate data interface.
+
+    Returns:
+        Store: A Legate store encapsulating the array sum
+    """
+    input = _get_legate_store(input)
```
manopapad, line 131: mypy might not like this, in which case it will want you to name the "sanitized" variable something different.

tests/examples/hello/hello/hello.py
```diff
@@ -0,0 +1,167 @@
...
+from enum import IntEnum
+from typing import Any
+
+import numpy as np
+
+import legate.core.types as types
+from legate.core import Rect, Store, get_legate_runtime
+
+from .library import user_context, user_lib
+
```
manopapad, line 11: Should we also include the "driver" code, which does the actual calls to print_hello, iota etc.?

tests/examples/hello/src/iota.cc
jjwilke: Done - except for the accessor. That's messy enough that I kept it `auto`

tests/examples/hello/src/iota.cc
jjwilke: 👍

tests/examples/hello/src/iota.cc
jjwilke: 👍

tests/examples/hello/hello/hello.py
jjwilke: Yep, I missed that on the first commit. Added.

tests/examples/hello/hello/hello.py
jjwilke: mypy in general chokes on the examples for a variety of reasons. I'm just excluding them for now. But I also renamed this `store_input` for this example.

examples/hello/README.md
```diff
@@ -0,0 +1,207 @@
...
+Here we illustrate a minimal example to get a Legate library up and running.
+The example here shows how to get started with the minimum amount of boilerplate.
+For advanced use cases, the boilerplate generated can be customized as needed.
+In general, a Legate application will need to implement three pieces.
+
+1. Build system
+1. C++ tasks
+1. Python library
+
+Please refer to the README in the [Legate repo](https://github.com/nv-legate/legate.core/blob/branch-23.03/README.md)
```
manopapad, line 12:
```suggestion
Please refer to the README in the [Legate repo](https://github.com/nv-legate/legate.core/blob/HEAD/README.md)
```

jjwilke:
@manopapad One last glance before merging? Sorry, you've done a lot of reading already : )

Helper scripts were added with documentation pointing to them.

manopapad (approved):

PR #572, opened 2023-02-15
Updates for cmd opts -> env vars

bryevdv:
This PR closes the loop with #567 and https://github.com/nv-legate/cunumeric/pull/799 

* Updates tester to set `CUNUMERC_TEST=1` instead of passing `-cunumeic:test`
* Removes argparse feature for help with library command line args (and also removes janky conditional imports)

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -17,7 +17,6 @@
 from __future__ import annotations
 
 from argparse import REMAINDER, ArgumentDefaultsHelpFormatter, ArgumentParser
-from typing import IO, Optional
 
 from ..util.shared_args import (
     CPUS,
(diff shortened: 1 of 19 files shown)


legate/driver/args.py
```diff
@@ -41,58 +40,7 @@
 __all__ = ("parser",)
```
manopapad, line 42: Should we also remove `parse_library_command_args` from `util/args.py`?

bryevdv: ah yes we certainly could

bryevdv: [7d2e5b4](https://github.com/nv-legate/legate.core/pull/572/commits/7d2e5b410481c9e421d4f364653601d7daa482f8)

manopapad (approved):

legate/tester/stages/util.py
```diff
@@ -25,10 +25,18 @@
...
 
-CUNUMERIC_TEST_ARG = "-cunumeric:test"
-
 UNPIN_ENV = {"REALM_SYNTHETIC_CORE_MAP": ""}
 
+CUNUMERIC_TEST_ENV = {"CUNUMERIC_TEST": "1"}
+
+# Raise min chunk sizes for deferred codepaths to force eager execution
+EAGER_ENV = {
+    "CUNUMERIC_MIN_CPU_CHUNK": "2000000000",
```
manopapad, line 34: Let's add `"CUNUMERIC_TEST": "0"` to this. The eager tests need to run with `CUNUMERIC_TEST` set to false, which is the default, but now it will be easier for users to unwittingly overwrite this setting by randomly having `CUNUMERIC_TEST` set in their environment, so better to make it explicit.

bryevdv: [16598d5](https://github.com/nv-legate/legate.core/pull/572/commits/16598d58a9f1f8ff6ce28003e6a85dd7ef2d0660)

PR #576, opened 2023-02-16
Add `--numamem` to the tester

magnatelee:
This PR adds `--numamem` to the tester so we can run tests with numa memory as before. (See the test driver failure: https://github.com/nv-legate/cunumeric/actions/runs/4191696755/jobs/7268552774)

### DIFF
--- a/legate/tester/__init__.py
+++ b/legate/tester/__init__.py
@@ -46,6 +46,9 @@ DEFAULT_OMPS_PER_NODE = 1
 #: Value to use if --ompthreads is not specified.
 DEFAULT_OMPTHREADS = 4
 
+#: Value to use if --numamem is not specified.
+DEFAULT_NUMAMEM = 0
+
 #: Default values to apply to normalize the testing environment.
 DEFAULT_PROCESS_ENV = {
     "LEGATE_TEST": "1",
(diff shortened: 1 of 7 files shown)


bryevdv:
Could use unit tests here:

* https://github.com/nv-legate/legate.core/blob/branch-23.03/tests/unit/legate/tester/test___init__.py
* https://github.com/nv-legate/legate.core/blob/branch-23.03/tests/unit/legate/tester/test_args.py

Otherwise LGTM

bryevdv (approved):

magnatelee:
> Could use unit tests here:
> 
> * https://github.com/nv-legate/legate.core/blob/branch-23.03/tests/unit/legate/tester/test___init__.py
> * https://github.com/nv-legate/legate.core/blob/branch-23.03/tests/unit/legate/tester/test_args.py

https://github.com/nv-legate/legate.core/pull/576/commits/3ba01acd5e94b17cc240b040aa8de7cc30b9e698

bryevdv:
@magnatelee is `--numamem` a `legate` arg? Then this is not actually necessary. But now legate args must go before the script name.

bryevdv:
@magnatelee FYI I think this could also be handled now with `LEGATE_CONFIG`

PR #577, opened 2023-02-16
Skip python 3.9.7 to work around python issue45121

manopapad:


m3vaz:
LGTM

m3vaz:
I would also want to try a conda build to confirm after this is merged. I know Conda does some special things when python/numpy are concerned

PR #578, opened 2023-02-17
Revert "NVTX: Use RangePush and Domain (#293)"

magnatelee:
One side-effect of #293 is that NVTX ranges are now associated with physical thread ids in Nsight. This is undesirable at the moment, as each GPU processor is currently rendered as multiple threads in Nsight (an issue that @manopapad and I are still investigating) and thus the ranges now spread across those threads. Even worse is that since each thread only has ~20 boxes to show, rendering is suppressed for them, as I guess they are seemingly uninteresting. Until we fix this behavior, we will fall back to the old code for NVTX ranges, which at least shows all the ranges on a single line.

cc @evanramos-nvidia @manopapad

### DIFF
--- a/src/core/runtime/runtime.cc
+++ b/src/core/runtime/runtime.cc
@@ -45,11 +45,6 @@ static const char* const core_library_name = "legate.core";
 
 /*static*/ bool Core::has_socket_mem = false;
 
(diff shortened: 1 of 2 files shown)


magnatelee:
Merging this PR without the CI passing, as CI will be out for another week.

PR #588, opened 2023-02-24
Copy warnings on stderr even if logging to file

manopapad:


bryevdv (approved):

bryevdv:
LGTM test here may need update as well 

https://github.com/nv-legate/legate.core/blob/branch-23.03/tests/unit/legate/driver/test_command.py#L1166

magnatelee (approved):

PR #590, opened 2023-02-27
Runtime network detection and UCX options

bryevdv:
This PR adds runtime detection of the Realm network (thanks to @trxcllnt for pointing me in a simpler direction than I was originally looking), exposed as a list `install_info.networks` 

Additionally, the options for `bgwork` and `bgworkpin` have been factored into their own `cmd_bgwork`

@manopapad before I add unit tests, one slight difference to clarify: previously, `-ll:bgwork` was *not* added if `utility == 1`. In the current PR, `cmd_bgwork` adds it unconditionally (even if `utility == 1`) Is this correct, or should `cmd_bgwork` retain that same condition?

### DIFF
--- a/legate/driver/command.py
+++ b/legate/driver/command.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from .. import install_info
 from ..util.ui import warn
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


legate/driver/command.py
```diff
@@ -258,26 +259,37 @@ def cmd_openmp(
...
-        return ()
-
-    opts: CommandPart = ("-ll:util", str(utility))
+    opts: CommandPart = ()
 
     # If we are running multi-rank then make the number of active
     # message handler threads equal to our number of utility
     # processors in order to prevent head-of-line blocking
     if ranks > 1:
         opts += ("-ll:bgwork", str(utility))
```
manopapad, line 274: The default for `-ll:bgwork` is 2, so let's just always allocate at least 2.
```suggestion
        opts += ("-ll:bgwork", max(str(utility), 2))
```

manopapad: @SeyedMir this heuristic, which sets `-ll:bgwork` equal to `-ll:util`, was (presumably) appropriate for gasnet. Is there a better default for this setting when using the UCX backend?

SeyedMir: bgwok 2 has shown the best performance with the UCX backend in my tests. I did not test with higher `-ll:util` values.

PR #592, opened 2023-02-27
Fix ucx:tls_host default value

SeyedMir:
"^dc,ud" will NOT work properly in UCX until 1.14.1 is released. So, for now we need to do it by listing all transports we except dc and ud.


SeyedMir:
@manopapad The UCX fix required for "^dc,ud" value to work properly for UCX_TLS will NOT be part of UCX 1.14.0. It will be added to 1.14.1.

manopapad (approved):

PR #593, opened 2023-02-28
Pin Legion to specific commit sha by default

trxcllnt:
This PR pins the Legion repo URL and git revision that's built if Legion is not found via `find_package()`.

It explicitly ignores values stored in the `CMakeCache.txt`, so subsequent rebuilds that don't set the `--legion-{url,branch}` args revert to the defaults:

```bash
$ install.py --editable --legion-branch "feature_branch"

# Reverts to building the default pinned Legion sha, not "feature_branch"
$ install.py --editable
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ endif()
 # - Download and initialize RAPIDS CMake helpers -----------------------------
 
 if(NOT EXISTS ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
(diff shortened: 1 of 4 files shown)


marcinz:
Is it possible to move the SHA for Legion to `versions.json`? I think we discussed it at some point, but I wonder how hard it is.

magnatelee:
> Is it possible to move the SHA for Legion to versions.json?

+1

trxcllnt:
Yes, that's possible.

magnatelee (approved):

PR #599, opened 2023-03-01
Add CTK stubs dir to implicit link directories

trxcllnt:
This PR adds the CTK stubs path to implicit link directories in legate's package config, so the stubs path doesn't end up in library consumers' RPATHs.

### DIFF
--- a/cmake/thirdparty/get_legion.cmake
+++ b/cmake/thirdparty/get_legion.cmake
@@ -86,37 +86,40 @@ function(find_or_configure_legion)
       endif()
 
       # Get the `stubs/libcuda.so` path so we can set CMAKE_LIBRARY_PATH for FindCUDA.cmake
-
-      # Prefer users' CUDA_PATH envvar (if set)
-      set(_cuda_stubs "$ENV{CUDA_PATH}")
-      if(NOT _cuda_stubs)
-        if(DEFINED ENV{CUDA_LIB_PATH})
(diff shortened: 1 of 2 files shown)


magnatelee (approved):

cmake/thirdparty/get_legion.cmake
```diff
@@ -131,6 +134,8 @@ function(find_or_configure_legion)
       list(APPEND CMAKE_C_IMPLICIT_LINK_DIRECTORIES "${_cuda_stubs}")
       list(APPEND CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES "${_cuda_stubs}")
       list(APPEND CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES "${_cuda_stubs}")
+      set(legate_core_cuda_stubs_path "${_cuda_stubs}" PARENT_SCOPE)
+      set(legate_core_cuda_stubs_path "${_cuda_stubs}" CACHE STRING "" FORCE)
```
jjwilke, line 138: Sorry - this is probably obvious - but why do we set a regular variable in the parent scope and a cache variable?

trxcllnt: I'm being defensive in case a reconfigure of the build dir causes this code path not to be. The package config file will still be written, and I don't want to drop the implicit include lines.

PR #600, opened 2023-03-01
add support for --version

bryevdv:


legate/driver/args.py
```diff
@@ -381,3 +382,9 @@
     required=False,
     help="Whether to use color terminal output (if colorama is installed)",
 )
+
+other.add_argument(
+    "--version",
+    action="version",
```
JohnZed, line 388: Nice, I did not know that action=version was built in to argparse! Makes sense..

bryevdv: https://docs.python.org/3/library/argparse.html#action

>'version' - This expects a version= keyword argument in the [add_argument()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument) call, and prints version information and exits when invoked:

It might be (somewhat) recent? I know I've had to write the little bit of code to do this manually in the past.

manopapad (approved):

PR #603, opened 2023-03-02
Fix a bug in the new registration callback API

magnatelee:
The new registration code was using the same callback to perform all Legate registration callbacks, which accidentally got them deduplicated. This PR fixes the bug by using a function template that will map each Legate callback to a unique Legion callback.

### DIFF
--- a/src/core/runtime/runtime.h
+++ b/src/core/runtime/runtime.h
@@ -52,9 +52,10 @@ struct Core {
    * @brief Performs a registration callback. Libraries must perform
    * registration of tasks and other components through this function.
    *
-   * @param callback Registration callback to perform
+   * @tparam CALLBACK Registration callback to perform
    */
-  static void perform_registration(RegistrationCallback callback);
+  template <RegistrationCallback CALLBACK>
+  static void perform_registration();
 
  public:
   // Configuration settings
@@ -66,3 +67,5 @@ struct Core {
 };
 
 }  // namespace legate
+
+#include "core/runtime/runtime.inl"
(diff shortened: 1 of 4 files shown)


src/core/runtime/runtime.h
```diff
@@ -52,9 +52,10 @@ struct Core {
    * @brief Performs a registration callback. Libraries must perform
    * registration of tasks and other components through this function.
    *
-   * @param callback Registration callback to perform
+   * @tparam CALLBACK Registration callback to perform
    */
-  static void perform_registration(RegistrationCallback callback);
```
jjwilke, line 57: Obviously we're going to update cunumeric ASAP, but as a matter of CI and deprecation best practice, I would probably keep the old callback and maybe mark it deprecated.  In this case, you can actually give the new and old callbacks the same name since C++ resolution allows functions with the same name where one is templated and the other is not.

jjwilke (changes requested):
One suggestion, but not a big deal.

Everything works in the other tests - so otherwise LGTM.

src/core/runtime/runtime.h
magnatelee: we've been using the branch name trick to match up PRs (maybe you didn't notice, but there's a companion PR for cuNumeric here: https://github.com/nv-legate/cunumeric/pull/827). and I think we continue doing something similar with the help from CI, instead of phasing out obsolete APIs like you suggested. Legate is still being actively developed, so I agree that we need a proper deprecation protocol, but I feel that'd just complicate the process for the time being.

jjwilke (approved):

src/core/runtime/runtime.h
jjwilke: Sounds good.

PR #604, opened 2023-03-02
More informative OOM message

manopapad:
I didn't mention the FieldManager LRU cache, as that will soon be irrelevant, after we start using Legion's new discard operation to eagerly delete instances.

Example output:

[0 - 70000a723000]    1.667014 {5}{cunumeric.mapper}: Mapper cunumeric on Node 0 failed to allocate 8000000 bytes on memory 1e00000000000000 (of kind SYSTEM_MEM: Visible to all processors on a node) for region requirement 0 of Task cunumeric::BinaryOpTask[a.py:24] (UID 22). This means Legate was unable to reserve ouf of its memory pool the full amount required for the above operation. Here are some things to try: (1) Make sure your code is not impeding the garbage collection of Legate-backed objects, e.g. by storing references in caches, or creating reference cycles. (2) Ask Legate to reserve more space on the above memory, using the appropriate --*mem legate flag. (3) Assign less memory to the eager pool, by reducing --eager-alloc-percentage. (4) If running on multiple nodes, increase how often distributed garbage collection runs, by reducing LEGATE_FIELD_REUSE_FREQ (default: 32, warning: may incur overhead). (5) Adapt your code to reduce temporary storage requirements, e.g. by breaking up larger operations into batches. (6) If the previous steps don't help, and you are confident Legate should be able to handle your code's working set, please open an issue on Legate's bug tracker.

### DIFF
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -764,7 +764,7 @@ bool BaseMapper::map_legate_store(const Legion::Mapping::MapperContext ctx,
       return false;
     }
     if (!can_fail)
-      report_failed_mapping(mappable, mapping.requirement_index(), target_memory, redop);
+      report_failed_mapping(mappable, mapping.requirement_index(), target_memory, redop, footprint);
     return true;
   }
 
@@ -863,7 +863,8 @@ bool BaseMapper::map_legate_store(const Legion::Mapping::MapperContext ctx,
   // If we make it here then we failed entirely
   if (!can_fail) {
     auto req_indices = mapping.requirement_indices();
-    for (auto req_idx : req_indices) report_failed_mapping(mappable, req_idx, target_memory, redop);
+    for (auto req_idx : req_indices)
+      report_failed_mapping(mappable, req_idx, target_memory, redop, footprint);
   }
   return true;
 }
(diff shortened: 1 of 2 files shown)


magnatelee (approved):

magnatelee:
I'm not a big fan of long single-line error messages, but I guess there's no other choice, as we can't avoid messages from different loggers interleaved with each other.

magnatelee:
we could potentially make this multi-line by padding each line at the beginning with the right amount of spaces so the messages are still lined up nicely; the log header format has been fixed for a while, so a hard-coded padding amount shouldn't be too bad.

manopapad:
I don't love adding the whitespace prefix, as that will look bad if the lines wrap. And there's some variable amounts in Realm's header (e.g. the amount of digits on the node id, or the logger name), that we would have to consider if we want to make this look just right.

I tried just starting subsequent lines from column 0, and I actually think it looks OK, since there's a clear distinction between the "header" part (which is the actual error message), and everything after it (which is the "explanatory" portion).

```
[0 - 700010442000]    1.698934 {5}{cunumeric.mapper}: Mapper cunumeric on Node 0 failed to allocate 8000000 bytes on memory 1e00000000000000 (of kind SYSTEM_MEM: Visible to all processors on a node) for region requirement 0 of Task cunumeric::BinaryOpTask[a.py:23] (UID 22).
This means Legate was unable to reserve ouf of its memory pool the full amount required for the above operation. Here are some things to try:
* Make sure your code is not impeding the garbage collection of Legate-backed objects, e.g. by storing references in caches, or creating reference cycles.
* Ask Legate to reserve more space on the above memory, using the appropriate --*mem legate flag.
* Assign less memory to the eager pool, by reducing --eager-alloc-percentage.
* If running on multiple nodes, increase how often distributed garbage collection runs, by reducing LEGATE_FIELD_REUSE_FREQ (default: 32, warning: may incur overhead).
* Adapt your code to reduce temporary storage requirements, e.g. by breaking up larger operations into batches.
* If the previous steps don't help, and you are confident Legate should be able to handle your code's working set, please open an issue on Legate's bug tracker.
```

PR #607, opened 2023-03-07
legate_helper_functions: fix missing `hello` substitutions

madsbk:
The current `legate_helper_functions.cmake` doesn't replace all occurrences of `hello`, fixed.

cc. @jjwilke


madsbk:
Alright

PR #612, opened 2023-03-07
Restrict gcc versions to match nvcc compatibility

bryevdv:
This PR adds gcc/gxx constraints when compilers are requested and CTK <= 11.5

In the current form, the gcc/gxx constraints are added to the "cuda" block in the environment file:

```
  # cuda
  - cudatoolkit=11.5
  - cutensor>=1.3.3
  - nccl
  - pynvml
  - gcc_linux-64<=11.2
  - gxx_linux-64<=11.2
```

The comment headings are only informational, and the constraint is derived from the cuda config, so this seems OK.

### DIFF
--- a/scripts/generate-conda-envs.py
+++ b/scripts/generate-conda-envs.py
@@ -17,6 +17,8 @@ from dataclasses import dataclass
 from textwrap import indent
 from typing import Literal, Protocol, Tuple
 
+from packaging.version import Version as V
+
 # --- Types -------------------------------------------------------------------
 
 Req = str
@@ -49,6 +51,7 @@ class SectionConfig(Protocol):
 @dataclass(frozen=True)
 class CUDAConfig(SectionConfig):
     ctk_version: str
+    compilers: bool
 
     header = "cuda"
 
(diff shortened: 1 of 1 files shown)


scripts/generate-conda-envs.py
```diff
@@ -57,13 +60,22 @@ def conda(self) -> Reqs:
...
-        return (
+        deps = (
             f"cudatoolkit={self.ctk_version}",  # runtime
             "cutensor>=1.3.3",  # runtime
             "nccl",  # runtime
             "pynvml",  # tests
         )
 
+        # gcc 11.3 is incompatible with nvcc <= 11.5.
+        if self.compilers and (V(self.ctk_version) <= V("11.5")):
```
manopapad, line 71: I think we also want to do this only on linux (which is presumably where gcc will be used as the backing for the c-compiler and cxx-compiler packages).

bryevdv: I keep forgetting we are nominally supporting non-linux platforms :)

bryevdv: @manopapad on OSX `self.ctk_version == "none"` so all this is skipped anyway. Do you still want an explicit check for linux? I guess that might future proof if windows is added.

manopapad: Yeah, let's add it for future-proofing.

bryevdv: https://github.com/nv-legate/legate.core/pull/612/commits/6f0d6d0a130ec908e23778ccb8582afb49934414

manopapad (approved):

PR #616, opened 2023-03-09
Avoid packaging dependency in conda env generate script

bryevdv:
This PR updates `generate-conda-envs.py ` to avoid needing any external dependencies.

### DIFF
--- a/scripts/generate-conda-envs.py
+++ b/scripts/generate-conda-envs.py
@@ -17,13 +17,16 @@ from dataclasses import dataclass
 from textwrap import indent
 from typing import Literal, Protocol, Tuple
 
-from packaging.version import Version as V
-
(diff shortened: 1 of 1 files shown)


scripts/generate-conda-envs.py
```diff
@@ -17,15 +17,18 @@
...
-
 # --- Types -------------------------------------------------------------------
 
 Req = str
 Reqs = Tuple[Req, ...]
 OSType = Literal["linux", "osx"]
 
 
+def V(version: str) -> tuple[int, ...]:
+    padded_version = (version.split(".") + ["0"])[:2]
```
bryevdv, line 28: I guess users can't actually pass "11" but just being extra safe

manopapad (approved):

PR #618, opened 2023-03-10
More robust nsys --sample flag with --nsys-extra

jjwilke:
The current nsys flag only checks for `-s` to deduplicate. The user might pass any of the following:

1. --sample <>
1. --sample=<>
1. -s <>
1. -s=<>

The user will get a cryptic error about --sample being passed twice unless they use `-s <>`

### DIFF
--- a/legate/driver/command.py
+++ b/legate/driver/command.py
@@ -127,7 +127,15 @@ def cmd_nsys(
 
     opts: CommandPart = ("nsys", "profile", "-t", targets, "-o", log_path)
     opts += tuple(extra)
-    if "-s" not in extra:
+
+    has_sample_flag = False
+    for option in extra:
+        flag = option.split("=")[0]
+        if flag == "-s" or flag == "--sample":
+            has_sample_flag = True
+            break
+
+    if not has_sample_flag:
         opts += ("-s", "none")
 
     return opts
--- a/legate/driver/command.py
+++ b/legate/driver/command.py
@@ -14,6 +14,7 @@
 #
 from __future__ import annotations
 
+import argparse
 from typing import TYPE_CHECKING
 
 from .. import install_info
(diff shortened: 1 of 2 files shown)


legate/driver/command.py
```diff
@@ -127,7 +127,15 @@ def cmd_nsys(
...
     opts: CommandPart = ("nsys", "profile", "-t", targets, "-o", log_path)
     opts += tuple(extra)
-    if "-s" not in extra:
+
+    has_sample_flag = False
+    for option in extra:
+        flag = option.split("=")[0]
+        if flag == "-s" or flag == "--sample":
+            has_sample_flag = True
+            break
```
bryevdv, line 136: This seems ok. This might also be OK

```python
has_sample_flag = any(x.startswith(("-s", "--sample")) for x in extra)
```
The "right" way to do this would be to define a separate `argparse` parser and call it manually with `extra` but that's probably excessive for this simple situation.

bryevdv (approved):

legate/driver/command.py
jjwilke: So I thought about that (not exactly that, but similar). The concern there was that there would be flags like "--sample-period" that would return true, but wouldn't actually be the `--sample` flag.  I guess `-s` is safe to check that way.

bryevdv: Definitely a valid concern, I would leave it as is then

bryevdv: This does seem amenable to a basic unit year in `test_command.py`

jjwilke:
@bryevdv PTAL. I added tests and made the checks more robust.

tests/unit/legate/driver/test_command.py
```diff
@@ -361,6 +362,28 @@ def test_multi_rank_no_launcher(
             "none",
         )
 
+    # spaces are necessary before the arguments to work around
+    # argparse inability to consume arguments with -- correctly
+    @pytest.mark.parametrize(
+        "nsys_extra",
+        (
+            " --sample=cpu",
+            " --backtrace=lbr -s cpu",
```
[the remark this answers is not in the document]
bryevdv: sorry I misinterpreted how this was being used, let me look into a fix for the spaces issue

tests/unit/legate/driver/test_command.py
```diff
@@ -361,6 +362,31 @@ def test_multi_rank_no_launcher(
             "none",
         )
 
+    # spaces are necessary before arguments to work around
+    # argparse inability to consume arguments with -- correctly
```
bryevdv, line 366: comment is out of date / irrelevant now

jjwilke: 👍

legate/driver/command.py
```diff
@@ -125,9 +126,20 @@ def cmd_nsys(
...
     extra = config.profiling.nsys_extra
 
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-s", "--sample")
+    parser.add_argument("-t", "--targets")
+    nsys_parsed_args, unparsed = parser.parse_known_args(extra)
+
+    if nsys_parsed_args.targets:
+        raise RuntimeError(
+            "please pass targets as arguments to --nsys"
```
manopapad, line 136:
```suggestion
            "please pass targets through --nsys-targets"
```

PR #621, opened 2023-03-14
remove generic create_task

bryevdv:
Follow up to https://github.com/nv-legate/cunumeric/pull/842 this PR removes the generic `create_task` API and leaves only explicit `create_auto_task` and `create_manual_task`

### DIFF
--- a/legate/core/context.py
+++ b/legate/core/context.py
@@ -370,8 +370,8 @@ class Context:
     def create_manual_task(
         self,
         task_id: int,
+        launch_domain: Rect,
         mapper_id: int = 0,
-        launch_domain: Optional[Rect] = None,
     ) -> ManualTask:
         """
         Creates a manual task.
--- a/legate/core/context.py
+++ b/legate/core/context.py
@@ -23,7 +23,6 @@ from typing import (
     Protocol,
     TypeVar,
     Union,
-    cast,
 )
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


legate/core/context.py
```diff
@@ -390,10 +387,6 @@ def create_task(
             Id of the mapper that should determine mapping policies for the
             task. Used only when the library has more than one mapper.
 
-        manual : bool
-            Indicates whether the task should be manually parallelized;
-            if ``True``, the task is parallelized manually by the caller.
-
         launch_domain : Rect, optional
             Launch domain of the task. Ignored if the task is automatically
             parallelized, mandatory otherwise.
```
manopapad, line 392:
```suggestion
            Launch domain of the task.
```

legate/core/context.py
```diff
@@ -368,16 +367,14 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
 
         return wrapper
 
-    def create_task(
+    def create_manual_task(
         self,
         task_id: int,
         mapper_id: int = 0,
-        manual: bool = False,
         launch_domain: Optional[Rect] = None,
```
manopapad, line 374: I think we can make this a mandatory argument now. Presumably we'd have to move it before `mapper_id`.

manopapad (approved):

legate/core/context.py
bryevdv: Very fortunately it is passed by kw everywhere it is currently used

bryevdv: https://github.com/nv-legate/legate.core/pull/621/commits/da4546fb882a56fbfe32eec4fe9596abec44d54b

manopapad (approved):

PR #625, opened 2023-03-15
Docs updates

bryevdv:
This PR makes some small fixes:

* manual `create_task` API needed removing (causing docs build failure)
* terminal copyright date was not correct
* rst file had unusable markdown-style link

Additionally the `conf.py` andindex was cleaned up and made to more closely match the changes that were already made to cunumeric. 

The front page now looks like this:

image

I do think the "Overview" would be improved if, instead of re-using the README.md, a dedicated `index.rst` was added. cc @magnatelee @manopapad

### DIFF
--- a/docs/legate/core/source/api/runtime.rst
+++ b/docs/legate/core/source/api/runtime.rst
@@ -50,7 +50,6 @@ runtime, the runtime gives back a context object unique to the library.
    :toctree: generated/
 
    context.Context.create_store
-   context.Context.create_task
    context.Context.create_manual_task
    context.Context.create_auto_task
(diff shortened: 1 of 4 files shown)


manopapad:
> I do think the "Overview" would be improved if, instead of re-using the README.md, a dedicated index.rst was added.

I agree, and more importantly I would say we should move a lot of the content out of the README into sub-sections of the sphinx docs. IMHO the README should only contain basic instructions for pulling from conda, doing a basic execution, and just point to the sphinx docs for the rest.

manopapad (approved):

PR #626, opened 2023-03-16
Use subsumption checks for instance policies

magnatelee:
The instance manager was checking equality between policies when deciding whether a given instance can be reused. That's overly strict in certain cases and was causing thrashing on the instance cache. This PR replaces the equality check with a subsumption check.


jjwilke (approved):
LGTM. Can confirm no more cache misses in the benchmark after the change.

PR #630, opened 2023-03-18
Start discarding instances for free fields

magnatelee:
This PR is to discard instances for free fields.


legate/core/runtime.py
```diff
@@ -327,6 +328,11 @@ def allocate_field(self) -> tuple[Region, int]:
     def free_field(
         self, region: Region, field_id: int, ordered: bool = False
```
manopapad, line 329: It looks like at this point this operation is always ordered (either because it's invoked after a consensus match, or because it's happening on a single-rank execution, in which case there is no other process to diverge from in terms of Python garbage collection). Otherwise it wouldn't be safe to emit a discard operation unconditionally. I suggest we remove the "ordered" argument from this method, and leave a comment that it's always implicitly ordered.

legate/core/_legion/operation.py
```diff
@@ -1517,3 +1517,50 @@ def launch(
...
+        """
+        A Discard operation will discard all valid instances for a given list
+        of region fields.
+
+        Parameters
+        ----------
+        region : Region
+            The logical region on which to relax restricted coherence
+        fields : int or FieldID or List[int] or List[FieldID]
+            The fields to perform the attach on
```
manopapad, line 1537:
```suggestion
        region : Region
            The logical region whose fields to discard
        fields : int or FieldID or List[int] or List[FieldID]
            The fields to discard
```

manopapad (approved):

legate/core/runtime.py
magnatelee: Unfortunately, we do need an unordered path for discard operations even on a single rank. The key role that unordered paths play is to avoid CFFI calls in Python destructors, as doing otherwise can cause hang.

legate/core/_legion/operation.py
magnatelee: https://github.com/nv-legate/legate.core/pull/630/commits/cc354e56046fbd694a1e5e4a93a81af6b4bbed53

legate/core/runtime.py
magnatelee: https://github.com/nv-legate/legate.core/pull/630/commits/cc354e56046fbd694a1e5e4a93a81af6b4bbed53

manopapad (approved):
OK, I confirmed that there is no race condition between the pending discard execution and the reuse of the field. Anything we might do with a reused field (e.g. launch a task that uses it, or attach to it) will need to go through a Legion operation, which will clear the pending deletions through the `@dispatch` wrapper before getting launched.

PR #633, opened 2023-03-22
IO example

magnatelee:


examples/io/setup.py
```diff
@@ -0,0 +1,59 @@
...
+        "Topic :: Scientific/Engineering",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+    ],
+    packages=find_packages(
+        where=".",
+        include=["legateio", "legateio.*"],
```
jjwilke, line 55: probably also want to list cunumeric as a dependency, even though the user has to install it themselves

examples/io/legateio/legateio.py
```diff
@@ -0,0 +1,408 @@
...
+        # Request a broadcasting for the input. Since this is the only store
+        # argument to the task, Legate will launch a single task from this
+        # task descriptor.
+        task.add_broadcast(self._store)
+        task.execute()
+
+    def to_uneven_tiles(self, path: str) -> None:
+        """
+        Dumps the IOArray into (potentially) uneven tiles
+
```
jjwilke, line 134: This does not run for me. It fails with:

```
Input shape: (5, 5)
[0 - 7f46b150f000]    0.850279 {3}{legateio}: Write to "test/.header"
[0 - 7f46b150f000]    0.850481 {3}{legateio}: Write a sub-array <0,0>..<4,4> to "test/0.0"
legion_python: /home/jwilke/Programs/legate-hello/legate.core/src/core/data/store.cc:300: legate::Domain legate::Store::domain() const: Assertion `!is_unbound_store_' failed.
Signal 6 received by node 0, process 1470359 (thread 7f46b150f000) - obtaining backtrace
Signal 6 received by process 1470359 (thread 7f46b150f000) at: stack trace: 20 frames
  [0] = /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7f46c05f5090]
  [1] = /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb) [0x7f46c05f500b]
  [2] = /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b) [0x7f46c05d4859]
  [3] = /lib/x86_64-linux-gnu/libc.so.6(+0x22729) [0x7f46c05d4729] 
```

examples/io/src/read_even_tiles.cc
```diff
@@ -0,0 +1,92 @@
...
+      in.read(reinterpret_cast<char *>(&extents[idx]), sizeof(legate::coord_t));
+
+    // Since the shape is already fixed on the Python side, the sub-store's
+    // extents should be the same as what's stored in the file
+    assert(shape.hi - shape.lo + legate::Point<DIM>::ONES() == extents);
+
+    logger.print() << "Read a sub-array of rect " << shape << " from " << path;
+
+    auto acc = output.write_accessor<VAL, DIM>();
+    for (legate::PointInRectIterator it(shape, false /*fortran_order*/);
```
jjwilke: This does not compile for me and needs an explicit `<DIM>`.  A better option might be a helper function that can properly do template argument deduction. Even better - we should fix this in Legion so that the shape can return an iter with a common begin/end idiom and this can be written with `for (auto point : shape.iter(false))`

examples/io/src/util.cc
```diff
@@ -0,0 +1,87 @@
...
+    for (int32_t idx = 0; idx < DIM; ++idx)
+      out.write(reinterpret_cast<const char *>(&extents[idx]),
+                sizeof(legate::coord_t));
+
+    if (empty)
+      return;
+    auto acc = store.read_accessor<VAL, DIM>();
+    // The iteration order here should be consistent with that in the reader
+    // task, otherwise the read data can be transposed.
+    for (legate::PointInRectIterator it(shape, false /*fortran_order*/);
```
jjwilke: This does not compile for me and needs an explicit `<DIM>`.  A better option might be a helper function that can properly do template argument deduction. Even better - we should fix this in Legion so that the shape can return an iter with a common begin/end idiom and this can be written with `for (auto point : shape.iter(false))`

examples/io/src/write_file.cc
```diff
@@ -0,0 +1,74 @@
...
+
+    // Store the type code and the number of elements in the array at the
+    // beginning of the file
+    std::ofstream out(filename,
+                      std::ios::binary | std::ios::out | std::ios::trunc);
+    out.write(reinterpret_cast<const char *>(&code), sizeof(int64_t));
+    out.write(reinterpret_cast<const char *>(&size), sizeof(size_t));
+
+    auto acc = input.read_accessor<VAL, 1>();
+    for (legate::PointInRectIterator it(shape); it.valid(); ++it) {
```
jjwilke, line 44: This does not compile for me and needs an explicit <DIM>.  A better option might be a helper function that can properly do template argument deduction. Even better - fix this in Legion so that the shape can return an iter with a common begin/end idiom and this can be written with `for (auto point : shape.iter(false))`

jjwilke (changes requested):
Some small fixes, otherwise LGTM.

Currently does not compile for me. After fixing compile, one of the tests does not run.

examples/io/setup.py
```diff
@@ -0,0 +1,59 @@
...
+    license="Apache 2.0",
+    classifiers=[
+        "Intended Audience :: Developers",
+        "Topic :: Database",
+        "Topic :: Scientific/Engineering",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
manopapad, line 51: Add python 3.11

examples/io/legateio/legateio.py
magnatelee: What compiler are you using? and what's the command for this issue? I can't reproduce it.

examples/io/setup.py
magnatelee: https://github.com/nv-legate/legate.core/pull/633/commits/eccdabbf323353492057eaaeb6f2cd99e7e43b9c

examples/io/src/read_even_tiles.cc
```diff
@@ -0,0 +1,87 @@
...
+
+#include "legate_library.h"
+#include "legateio.h"
+#include "util.h"
+
+namespace fs = std::filesystem;
+
+namespace legateio {
+
+namespace detail {
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/633/commits/964f05e88cfcd305a1db989b4c251a5a1f033028

examples/io/src/read_file.cc
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+#include <fstream>
+
+#include "legate_library.h"
+#include "legateio.h"
+
+namespace legateio {
+
+namespace detail {
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/633/commits/964f05e88cfcd305a1db989b4c251a5a1f033028

examples/io/src/write_file.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/633/commits/964f05e88cfcd305a1db989b4c251a5a1f033028

magnatelee:
> Even better - fix this in Legion so that the shape can return an iter with a common begin/end idiom and this can be written with for (auto point : shape.iter(false))

@jjwilke I agree that this better be fixed in Realm. (FYI, `PointInRectIterator` is a Realm API.)

Issue #632, opened 2023-03-21
Querying legate's built-in capabilities


manopapad:
As requested in #631. Suggestion is to have a `legate --info` or similar flag, that outputs the following compile-time configuration choices:

* compile mode (debug, debugRelease, release)
* OpenMP support
* CUDA support
* networks (gasnet1, gasnetex, mpi, ucx)
* gasnet conduit (mpi, ibv, ucx, ofi, ...) (if building with gasnet1/gasnetex)

PR #637, opened 2023-03-23
add --info command line option

bryevdv:
fixes: #632
 
This PR adds an `--info` command line option to the driver that prints out information about the build configuration:
```
docs310 ❯ legate --info
Legate build configuration:
  build_type : Release
  use_openmp : True
  use_cuda   : True
  networks   : None
  conduit    : 
```

Questions:

* Would machine-readable (e.g. JSON) output be preferable?
* Currently convert empty list to "None" for "networks" field, is this desired? Could print nothing, or [], or omit the field in this case.
* Prints nothing if there is no conduit, is this desired? Could omit the field, etc. 

cc @manopapad @trxcllnt

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 from argparse import REMAINDER, ArgumentDefaultsHelpFormatter, ArgumentParser
 
 from .. import __version__
+from ..util.args import InfoAction
 from ..util.shared_args import (
     CPUS,
     FBMEM,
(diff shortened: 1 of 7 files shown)


manopapad:
> Would machine-readable (e.g. JSON) output be preferable?

Not necessary at this point I would say.

> Currently convert empty list to "None" for "networks" field, is this desired? Could print nothing, or [], or omit the field in this case.
> Prints nothing if there is no conduit, is this desired? Could omit the field, etc.

I would suggest to always print out the field names, to make it obvious to the user that the corresponding configuration is missing. We can revisit if this configuration list ever grows to an unwieldy size. 

I would suggest being uniform between "networks" and "conduit", i.e. either

```
networks : None
conduit  : None
```

or

```
networks :
conduit  :
```

but I'm not particularly opinionated on this.

PR #639, opened 2023-03-24
Add per-library loggers at the python level

manopapad:
Useful for debugging multi-node issues, where we care to output information from all python ranks, not just the first one.

Requires https://gitlab.com/StanfordLegion/legion/-/merge_requests/719

@magnatelee Note that there already is a C++ logger in this repo, named "legate", which doesn't conflict with the python-level logger, which uses the library's name "legate.core". It would take some additional care if we wanted to merge the two (or make sure they are separate).


legate/core/_legion/util.py
```diff
@@ -416,3 +416,33 @@ def get_string(self) -> Optional[bytes]:
...
         return self.size
+
+
+class Logger:
+    def __init__(self, name: str) -> None:
+        self.handle = legion.legion_logger_create(name.encode("utf-8"))
+
+    def __del__(self) -> None:
+        legion.legion_logger_destroy(self.handle)
+        self.handle = None
```
bryevdv, line 427: There's no guarantee that a `__del__` will be called on interpreter exit, and if classes with`__del__` ever become part of an unreachable cycle, the presence of `__del__` makes the cycle uncollectible by the GC. I guess I'd ask:

* What are the consequences of not doing this at all?
* Could it be done in an `atexit` instead?

manopapad: > if classes with__del__ ever become part of an unreachable cycle, the presence of __del__ makes the cycle uncollectible by the GC

I think this no longer the case, as of python 3.4: https://docs.python.org/3/library/gc.html#gc.garbage

> There's no guarantee that a __del__ will be called on interpreter exit

Good point, I added an explicit "destructor", and calling it recursively during `Context` destruction.

> What are the consequences of not doing this at all?

Not very serious in this case, as we don't create very many of these Loggers. We would just end up leaking a small amount of state until process shutdown (logging messages should still get eagerly flushed).

bryevdv (approved):

bryevdv:
:+1: for explicit `destroy`

manopapad:
@bryevdv I added the `want_<level>` methods, as suggested over at https://gitlab.com/StanfordLegion/legion/-/merge_requests/719. We query the information once and cache it locally, then we can early-exit in the logging methods and avoid the CFFI call if the corresponding logging level is not requested. The calling code can also query these levels manually, if they want to avoid doing some expensive query / string manipulation depending on logging level.

bryevdv (approved):

PR #645, opened 2023-03-27
Bug/allow rebuilds with scikit build

robertmaynard:
Currently due to the implicit logic around `SKBUILD` it isn't possible for projects to safely build/install multiple times in a row.

  Consider a project that looks like:
  ```
  legate_add_cpp_subdirectory(cpp TARGET legate_user EXPORT legate_user-export)
  legate_default_python_install(legate_user EXPORT legate_user-export)
  ```
  
  When a local user runs scikit-build the first time everything will work correctly as `legate_user` doesn't exist in the install location. But when invoked a second time it will find the installed version, and not build. So to install an updated version
of the files you need to first clean your env which is not a nice pattern.

So with the PR, projects can explicitly state they always want to build from source and never search for a local version.

### DIFF
--- a/cmake/legate_helper_functions.cmake
+++ b/cmake/legate_helper_functions.cmake
@@ -202,7 +202,7 @@ function(legate_default_python_install target)
 endfunction()
 
 function(legate_add_cpp_subdirectory dir)
-  set(options)
+  set(options ALWAYS_BUILD)
   set(one_value_args EXPORT TARGET)
   set(multi_value_args)
   cmake_parse_arguments(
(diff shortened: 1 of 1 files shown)


trxcllnt (approved):

jjwilke:
@robertmaynard
Thanks for catching this!

You mean the `find_package` below will pick up the installed version instead of the build tree version?

```
rapids_find_package(${target} CONFIG
      GLOBAL_TARGETS legate::${target}
      BUILD_EXPORT_SET ${LEGATE_OPT_EXPORT}
      INSTALL_EXPORT_SET ${LEGATE_OPT_EXPORT})
```

I don't think we would ever want that find package to find an installed version. Should I just restrict the find package call to only look in the build tree? I forget what the option is. `NO_DEFAULT_PATH` or something.

PR #653, opened 2023-03-29
Guard against attaching to non-contiguous buffers

manopapad:
This is currently extraneous, as cffi.from_buffer(.) only works with C-contiguous buffers anyway, but good to have this documented.

### DIFF
--- a/legate/core/_legion/operation.py
+++ b/legate/core/_legion/operation.py
@@ -1145,7 +1145,7 @@ class Attach(Dispatchable[PhysicalRegion]):
         self._launcher = ffi.gc(
             self.launcher, legion.legion_attach_launcher_destroy
         )
-        if not data.c_contiguous and not data.f_contiguous:
+        if not data.contiguous:
             raise RuntimeError("Can only attach to C- or F-contiguous buffers")
         legion.legion_attach_launcher_add_cpu_soa_field(
             self.launcher,
@@ -1314,7 +1314,7 @@ class IndexAttach(Dispatchable[ExternalResources]):
         for sub_region, buf in shard_local_data.items():
             if sub_region.parent is not None:
                 assert sub_region.parent.parent is parent
-            if not buf.c_contiguous and not buf.f_contiguous:
+            if not buf.contiguous:
                 raise RuntimeError(
                     "Can only attach to C- or F-contiguous buffers"
                 )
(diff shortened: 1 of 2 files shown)


legate/core/_legion/operation.py
```diff
@@ -1145,14 +1145,16 @@ def __init__(
         self._launcher = ffi.gc(
             self.launcher, legion.legion_attach_launcher_destroy
         )
+        if not data.c_contiguous and not data.f_contiguous:
+            raise RuntimeError("Can only attach to C- or F-contiguous buffers")
```
bryevdv, line 1149: Woudn't `not data.contiguous` also suffice here? Python docs for [`contiguous`](https://docs.python.org/3/glossary.html#term-contiguous) state

> A buffer is considered contiguous exactly if it is either C-contiguous or Fortran contiguous.

legate/core/_legion/operation.py
```diff
@@ -1145,14 +1145,16 @@ def __init__(
...
         legion.legion_attach_launcher_add_cpu_soa_field(
             self.launcher,
             ffi.cast(
                 "legion_field_id_t",
                 field.fid if isinstance(field, FieldID) else field,
             ),
             ffi.from_buffer(data),
-            data.f_contiguous,
+            not data.c_contiguous,
```
bryevdv, line 1157: What case is missed by the original code, under the condition that at this point, `data` is asserted to be `contiguous` ?

legate/core/_legion/operation.py
manopapad: You're right, I hadn't looked into this very carefully. There are (generally speaking) ways to lay out a 3+ dimension array such that it's "contiguous", but neither in C- nor Fortran-order (e.g. by ordering the dimensions `x > z > y`), but it looks like cpython doesn't consider this case.

legate/core/_legion/operation.py
manopapad: 0d and 1d contiguous array are considered to be both `c_contiguous` and `f_contiguous`, so in that case I wanted to tell the attachment operation to take the C-contiguous path. Not that it matters very much.

bryevdv: Ah that makes sense, a short comment might be advised (and maybe also to the effect that it doesn't really matter)

legate/core/_legion/operation.py
manopapad: Changed to just checking `.contiguous` with d891466071d52d98979134b7e171ed7e5d2b33ae

legate/core/_legion/operation.py
manopapad: Yup, added suggested comments with f01cce9485a47c8287b59975e3ea39b792be4bcc

bryevdv (approved):

PR #655, opened 2023-03-29
Fix documentation issues

marcinz:
Need to generate heading anchors and the link must be absolute to work in Sphinx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,6 +473,6 @@ documentation](https://nv-legate.github.io/legate.core).
 We recommend starting by experimenting with at least one Legate application
 library to test out performance and see how Legate works. If you are interested
 in building your own Legate application library, we recommend that you
-investigate our [Legate Hello World application library](https://github.com/nv-legate/legate.core/tree/branch-23.05/examples/hello) that
+investigate our [Legate Hello World application library](https://github.com/nv-legate/legate.core/tree/HEAD/examples/hello) that
 provides a small example of how to get started developing your own drop-in
 replacement library on top of Legion using the Legate Core library.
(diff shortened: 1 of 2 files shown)


marcinz:
Unfortunately, to create a link to the file, one must hardcode the version of the tree. That creates an issue where this must be updated periodically. @bryevdv, is there a better way to do that?

bryevdv:
> is there a better way to do that?

Not that I am aware of, out of the box. For Bokeh we created a custom sphinx role to link to the right branch automatically:

https://github.com/bokeh/bokeh/blob/74b1f1a227dd1aa4735caa13199e79986b7e67f5/src/bokeh/sphinxext/bokeh_roles.py#L176-L203

```python
    app = inliner.document.settings.env.app

    tag = app.env.config["version"]
    if "-" in tag:
        tag = "main"

    url = f"{BOKEH_GH}/tree/{tag}/{text}"
    options = options or {}
    set_classes(options)
    node = nodes.reference(rawtext, text, refuri=url, **options)
    return [node], []
```
I could add a similar directive to our Sphinx setup if that's desired. We just need to be able to get the version from somewhere in the config

manopapad:
> For the README, we could point a link to HEAD, and that is supposed to always go to the current default branch.

I think we can just do that for now. The plan is to move away from copying the .md files to the sphinx tree anyway, and at that point we can remove this explicit link from README.md.

README.md
```diff
@@ -473,6 +473,6 @@ documentation](https://nv-legate.github.io/legate.core).
 We recommend starting by experimenting with at least one Legate application
 library to test out performance and see how Legate works. If you are interested
 in building your own Legate application library, we recommend that you
-investigate our [Legate Hello World application library](examples/hello) that
+investigate our [Legate Hello World application library](https://github.com/nv-legate/legate.core/tree/branch-23.05/examples/hello) that
```
bryevdv, line 476: @marcinz I think this is the best we can do in the short term
```suggestion
investigate our [Legate Hello World application library](https://github.com/nv-legate/legate.core/tree/HEAD/examples/hello) that
```

bryevdv (approved):
one suggestion

PR #658, opened 2023-03-31
Add support for integer division to ProjExpr

sdalton1:
Expand support of ProjExpr for integer division.

Use case:

```
task.add_input(a, lambda p : (p[0] // rows + row_offset, p[1] // cols + col_offset))
```

### DIFF
--- a/legate/core/projection.py
+++ b/legate/core/projection.py
@@ -83,6 +83,10 @@ class ProjExpr:
             raise ValueError("RHS must be an integer")
         return ProjExpr(self._dim, self._weight, self._offset + other)
 
+    def __floordiv__(self, other: int) -> ProjExpr:
+        if not isinstance(other, int):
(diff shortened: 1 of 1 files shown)


magnatelee:
@sdalton1 I don't think this would work without changing the type of affine transformations used in projection functors.

A bit of background for your understanding: the projection function you pass to a `add_*` call is symbolically analyzed by legate and converted into an affine transformation of integer weights. and in the projection functor, we use it to map points in the launch domain to colors in the partition: https://github.com/nv-legate/legate.core/blob/branch-23.05/src/core/runtime/projection.cc#L150 to support divisions, we need to change the weights to be floats, which could lead to an accuracy issue, or build a separate machinery for divisions.

sdalton1:
Thanks for the clarification. I will close this issue and try restructuring my projection function a different way.

PR #660, opened 2023-03-31
Reduction example

magnatelee:


examples/reduction/examples/bincount.py
```diff
@@ -0,0 +1,41 @@
...
+    store = user_context.create_store(ty.uint64, size)
+    to_cunumeric_array(store)[:] = np.random.randint(
+        low=0, high=num_bins - 1, size=size
+    )
+    print_store(store)
+
+    result = bincount(store, num_bins)
+    print_store(result)
+
+    # Bogus downstream computation to consume the bincount result
```
jjwilke, line 36: Why is this here? Wouldn't the print force execution? Why do we have to consume the result?

examples/reduction/examples/matmul.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-m",
+        type=int,
+        default=3,
+        dest="m",
+        help="Extent of the first dimension",
```
jjwilke, line 72: maybe say "first exterior"

examples/reduction/examples/matmul.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+    parser.add_argument(
+        "-n",
```
jjwilke, line 75: may say "second exterior"

examples/reduction/examples/matmul.py
```diff
@@ -0,0 +1,104 @@
...
+
+        result = sum_over_axis(tmp, 1)
+        if print_stores:
+            print_store(result)
+
+    result = matmul(rhs1, rhs2)
+    if print_stores:
+        print_store(result)
+
+    # Bogus downstream computation to consume the matmul result
```
jjwilke, line 61: Same question here. Why is this needed? what makes it "bogus"?

examples/reduction/examples/matmul.py
```diff
@@ -0,0 +1,104 @@
...
+
+import legate.core.types as ty
+
+
+def test(m: int, n: int, k: int, print_stores: bool, matmul_only: bool):
+    # Generate inputs using cuNumeric
+    rhs1 = user_context.create_store(ty.int64, (m, k))
+    rhs2 = user_context.create_store(ty.int64, (k, n))
+    tmp = np.arange(m * k).reshape(m, k)
+    to_cunumeric_array(rhs1)[:] = tmp
```
jjwilke, line 36: this way of creating a temp array and then assigning through a slice via a cunumeric array was a bit odd to me - and might look a bit funny in a tutorial.

If we're going to depend on cunumeric, I would just create it as a cunumeric array and add a helper function `get_cunumeric_store`?

examples/reduction/reduction/reduction.py
```diff
@@ -0,0 +1,268 @@
...
+        Summation result
+    """
+    sanitized = _sanitize_axis(axis, input.ndim)
+
+    # Compute the output shape by removing the axis being summed over
+    res_shape = tuple(
+        ext for dim, ext in enumerate(input.shape) if dim != sanitized
+    )
+    result = context.create_store(input.type.type, res_shape)
+    to_cunumeric_array(result).fill(0)
```
jjwilke, line 99: same comment here. It seems more intuitive to me to create a cunumeric zeros and then get the store from it.

examples/reduction/examples/matmul.py
```diff
@@ -0,0 +1,104 @@
...
+
+import argparse
+
+import cunumeric as np
+from reduction import (
+    matmul,
+    multiply,
+    print_store,
+    sum_over_axis,
+    to_cunumeric_array,
```
jjwilke, line 24: Personal preference would be to only have a `print_store` function and that be the only place we need a `to_cunumeric_array` function. It seems more intuitive to me to create cunumeric arrays and fetch stores from them rather than create a store and then convert it to a cunumeric array.

examples/reduction/examples/histogram.py
```diff
@@ -0,0 +1,73 @@
...
+    to_cunumeric_array,
+    user_context,
+)
+
+import legate.core.types as ty
+
+
+def test(size: int, num_bins: int):
+    input = user_context.create_store(ty.int32, size)
+    to_cunumeric_array(input)[:] = np.random.randint(
```
jjwilke, line 33: Same comment here. It was a bit weird to me to create a cunumeric array from the store, then assign to a slice of it from a numpy array. Might be more intuitive to create a cunumeric random array and then have a `get_cunumeric_store` helper function?

examples/reduction/reduction/reduction.py
```diff
@@ -0,0 +1,195 @@
...
+    MUL = user_lib.cffi.MUL
+    SUM_OVER_AXIS = user_lib.cffi.SUM_OVER_AXIS
+
+
+class _Wrapper:
+    def __init__(self, store: Store) -> None:
+        self._store = store
+
+    @property
+    def __legate_data_interface__(self) -> dict[str, Any]:
```
jjwilke, line 42: Should this just be a part of the Store? Shouldn't we just have `Store` define its own `__legate_data_interface__` ?

examples/reduction/examples/matmul.py
csadorf: In general, it would probably be better to to implement the `__str__` function of the store (wrapper) so that `print(store)` would yield the desired result.

examples/reduction/reduction/reduction.py
csadorf: Yes, that seems like a better approach. I assume that the implementation here serves both as a stop-gap measure and also guides and motivates the implementation of such needed features.

magnatelee:
@jjwilke I addressed your comments in the latest commit, so please take another look. (I had to wait until the core type system is merged to do this properly.) Note that I simply replaced `to_cunumeric_array` with `np.asarray`, as extracting a store from a cunumeric ndarray is way more verbose than wrapping a store with an ndarray and calling a cunumeric method on it. Let me know if you still think there's a better way.

examples/reduction/src/histogram.cc
```diff
@@ -0,0 +1,84 @@
...
+    auto num_bins = bin_shape.hi[0] - bin_shape.lo[0];
+
+    auto in_acc  = input.read_accessor<VAL, 1>();
+    auto bin_acc = bins.read_accessor<VAL, 1>();
+    auto res_acc = result.reduce_accessor<legate::SumReduction<uint64_t>, true, 1>();
+
+    for (legate::PointInRectIterator<1> it(in_shape); it.valid(); ++it) {
+      auto& value = in_acc[*it];
+      for (auto bin_idx = 0; bin_idx < num_bins; ++bin_idx)
+        if (bin_acc[bin_idx] <= value && value < bin_acc[bin_idx + 1]) {
```
jjwilke, line 47: may wish to have a clarifying comment:

// Use naive algorithm that loops all bin lower bounds to find match

src/core/mapping/mapping.cc
```diff
@@ -211,7 +211,8 @@ void StoreMapping::populate_layout_constraints(
     }
   } else
     fields.push_back(stores.front().get().region_field().field_id());
-  layout_constraints.add_constraint(Legion::FieldConstraint(fields, true /*contiguous*/));
+  layout_constraints.add_constraint(
```
jjwilke, line 214: Why is this change needed? This seems like it could have major implications. Is it required in this PR?

jjwilke (approved):
Overall LGTM. Couple minor questions/nits.

examples/reduction/src/histogram.cc
magnatelee: https://github.com/nv-legate/legate.core/pull/660/commits/7b552355ba524569ebcf84760e825e1fde955f6c

src/core/mapping/mapping.cc
magnatelee: This and other changes for `tree_reduce` have been factored out into #699, so this diff will disappear once this PR is rebased on it. FYI, this diff actually fixes a performance problem for unbound stores; Legion's layout constraint check is currently too strict for single-field instances and it ends up refusing to reuse instances from unbound stores for downstream tasks even when they are reusable.

PR #670, opened 2023-04-07
Clean up handling of Legion config options

manopapad:
* Always build Legion with python support and FFI bindings. This way even if a user starts with a default cmake build, then does `pip install`, the Legion build that cmake did will automatically be compatible with the python layer. We could just change the default to be ON everywhere, but I don't see a reason why we need to support Legion builds w/o python and bindings, but maybe folks have a different opinion.
* Move unconditional Legion configuration flags out of `install.py`, and leave a note to put those in the cmake configuration. As is we run the risk that developers will only update `install.py` and forget to update the cmake-only defaults.
* Document major required Legion configuration flags, so that users know how to build Legion such that it's compatible with Legate. Automatically check all the ones that Legion exports (others will have to wait until the Legion cmake build starts exporting them).

### DIFF
--- a/cmake/thirdparty/get_legion.cmake
+++ b/cmake/thirdparty/get_legion.cmake
@@ -179,6 +179,7 @@ function(find_or_configure_legion)
           FIND_PACKAGE_ARGUMENTS EXACT
           EXCLUDE_FROM_ALL       ${exclude_from_all}
           OPTIONS                ${_legion_cuda_options}
+                                 "BUILD_SHARED_LIBS ON"
                                  "CMAKE_CXX_STANDARD ${_cxx_std}"
                                  "Legion_VERSION ${version}"
                                  "Legion_BUILD_BINDINGS ON"
@@ -197,6 +198,8 @@ function(find_or_configure_legion)
                                  "Legion_USE_OpenMP ${Legion_USE_OpenMP}"
                                  "Legion_USE_Python ON"
                                  "Legion_BOUNDS_CHECKS ${Legion_BOUNDS_CHECKS}"
+                                 "Legion_BUILD_JUPYTER ON"
+                                 "Legion_EMBED_GASNet_CONFIGURE_ARGS --with-ibv-max-hcas=8"
     )
   endif()
 
--- a/cmake/thirdparty/get_legion.cmake
+++ b/cmake/thirdparty/get_legion.cmake
@@ -70,6 +70,16 @@ function(find_or_configure_legion)
 
   if(Legion_FOUND)
     message(STATUS "CPM: using local package Legion@${version}")
+
+    # Check that required Legion options are set
+    if(NOT Legion_USE_Python)
+      message(FATAL_ERROR "Legion was not compiled with Legion_USE_Python")
+    endif()
+
+    # TODO: The following should also be checked, but are not currently reported by Legion:
+    # Legion_BUILD_BINDINGS
+    # Legion_REDOP_HALF
+    # Legion_REDOP_COMPLEX
   else()
 
     include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules/cpm_helpers.cmake)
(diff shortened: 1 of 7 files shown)


cmake/thirdparty/get_legion.cmake
```diff
@@ -195,19 +206,19 @@ function(find_or_configure_legion)
                                  "Legion_USE_CUDA ${Legion_USE_CUDA}"
                                  "Legion_NETWORKS ${Legion_NETWORKS}"
                                  "Legion_USE_OpenMP ${Legion_USE_OpenMP}"
-                                 "Legion_USE_Python ${Legion_USE_Python}"
+                                 "Legion_USE_Python ON"
                                  "Legion_BOUNDS_CHECKS ${Legion_BOUNDS_CHECKS}"
+                                 "Legion_BUILD_JUPYTER ON"
+                                 "Legion_EMBED_GASNet_CONFIGURE_ARGS --with-ibv-max-hcas=8"
     )
   endif()
```
jjwilke, line 215: We should probably validate that whatever Legion is found sets "Legion_USE_PYTHON"

```
macro(_check_required_legion_option option)
if (NOT ${option})
  message(FATAL_EROR "External Legion was not build with ${option}=ON")
endif()
endmacro()

_check_required_legion_option(Legion_USE_PYTHON)
_check_required_legion_option(Legion_REDOP_HALF)

```

install.py
```diff
@@ -747,14 +743,18 @@ def driver():
         dest="legion_dir",
         required=False,
         default=None,
-        help="Path to an existing Legion build directory.",
+        help="Path to an existing Legion build directory. A recent checkout "
+        "of the control_replication branch is required, configured with "
```
jjwilke, line 747: I don't think the `control_replication` branch would be an actual requirement?

See above. We should enforce these requirements in the cmake rather than just documenting them here.

jjwilke (changes requested):

install.py
```diff
@@ -436,13 +438,7 @@ def validate_path(path):
 -DLegion_USE_LLVM={("ON" if llvm else "OFF")}
 -DLegion_NETWORKS={";".join(networks)}
 -DLegion_USE_HDF5={("ON" if hdf else "OFF")}
--DLegion_USE_Python=ON
 -DLegion_Python_Version={pyversion}
--DLegion_REDOP_COMPLEX=ON
```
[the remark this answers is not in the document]
manopapad: It is set, in the `rapids_cpm_find(Legion ...` command in `cmake/thirdparty/get_legion.cmake`.

cmake/thirdparty/get_legion.cmake
manopapad: I am doing that around line 75 in the same file.

Legion doesn't actually export `Legion_REDOP_*` it seems, at least according to what's in https://github.com/StanfordLegion/legion/blob/stable/cmake/LegionConfigCommon.cmake.in. I realized we can actually check that in `legate_defines.h`, so the only major configuration option that's missing is `Legion_BUILD_BINDINGS`.

install.py
manopapad: Unfortunately the Legion API has significant changes between the `master` branch (which is what actually gets released) and the `control_replication` branch, which we depend on. AFAIK there is no exported configuration option that we can use to check whether this is a `master` or `control_replication` Legion build. This is orthogonal to the fact that we require `Legion_REDOP_COMPLEX`.

Perhaps we could use some other trick to confirm that we're on a `control_replication` build, but this still doesn't solve the bigger problem of version compatibility between Legion and Legate (Legion adds an API extension, we start depending on that, but there is no version tagging on the `control_replication` tree that we can use to confirm that the linked Legion is new enough).

I think the best long-term solution here is to wait for the merge of `control_replication` into `master`, and then we can start depending on specific released versions of Legion. But maybe @lightsighter has another suggestion.

cmake/thirdparty/get_legion.cmake
jjwilke: whoops, I can't read

install.py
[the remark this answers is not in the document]
jjwilke: yeah, it is. I apparently can't click on the github "expand" arrows.

install.py
jjwilke: sounds good

jjwilke (approved):
LGTM. Installation works and checks correctly catch errors.

cmake/thirdparty/get_legion.cmake
```diff
@@ -70,6 +70,14 @@ function(find_or_configure_legion)
 
   if(Legion_FOUND)
     message(STATUS "CPM: using local package Legion@${version}")
+
+    # Check that required Legion options are set
+    if(NOT Legion_USE_Python)
+      message(FATAL_ERROR "Legion was not compiled with Legion_USE_Python")
+    endif()
```
trxcllnt, line 77: This makes it impossible to build the legate C++ library without Legion built w/ Python bindings. Is this really what we want?

trxcllnt (changes requested):

cmake/thirdparty/get_legion.cmake
manopapad: This change is my attempt at avoiding the following fell-bad scenario:

* user builds Legion separately, forgets to add python support
* user builds legate.core C++ bits using `cmake` directly
* users does `pip install`
* user realizes nothing is working, because they built Legion w/o Python support (and the build process never complained)

I would also be ok with a situation where the cmake build doesn't enforce `Legion_USE_Python`, and instead that becomes necessary only when you try to install the python pieces of legate.core. Do you think that's possible/preferable? Or maybe you have a different approach to avoid this scenario?

PR #676, opened 2023-04-11
Mapper refactoring

magnatelee:
Per the internal discussion, this PR implements the following changes to the Legate mapper:

- Libraries can now have only one mapper and should provide their mappers at the registration time.
- When no mapper is given, the default mapper will be used for the library.
- Mapper id parameters have been removed from the APIs where they were expected.

This PR builds on #675 and so it currently inherits all of the changes. I will rebase this PR once #675 is merged.

### DIFF
--- a/src/core/mapping/default_mapper.h
+++ b/src/core/mapping/default_mapper.h
@@ -21,7 +21,7 @@
 namespace legate {
 namespace mapping {
 
-class DefaultMapper : public LegateMapper {
+class DefaultMapper : public Mapper {
  public:
   virtual ~DefaultMapper() {}
 
--- a/src/core/mapping/default_mapper.h
+++ b/src/core/mapping/default_mapper.h
@@ -1,0 +1,40 @@
+/* Copyright 2023 NVIDIA Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


src/core/mapping/default_mapper.h
```diff
@@ -0,0 +1,40 @@
...
+class DefaultMapper : public Mapper {
+ public:
+  virtual ~DefaultMapper() {}
+
+ public:
+  void set_machine(const MachineQueryInterface* machine);
+  TaskTarget task_target(const Task& task, const std::vector<TaskTarget>& options);
+  std::vector<StoreMapping> store_mappings(const Task& task,
+                                           const std::vector<StoreTarget>& options);
+  virtual Scalar tunable_value(TunableID tunable_id);
```
jjwilke, line 33: Does this need to be marked virtual?

src/core/mapping/default_mapper.h
```diff
@@ -0,0 +1,40 @@
...
+namespace legate {
+namespace mapping {
+
+class DefaultMapper : public Mapper {
+ public:
+  virtual ~DefaultMapper() {}
+
+ public:
+  void set_machine(const MachineQueryInterface* machine);
+  TaskTarget task_target(const Task& task, const std::vector<TaskTarget>& options);
```
jjwilke, line 30: Should these be marked override?

jjwilke (approved):
LGTM. A couple minor nits.

src/core/mapping/default_mapper.h
magnatelee: my initial thought is that there could be some mappers that would only have custom tunables (though that'd need the machine query interface to be `protected`), but now I think that'd rather need a custom mapper. let me remove that.

src/core/mapping/default_mapper.h
magnatelee: yup, I forgot to put those. Will add them.

manopapad (approved):

PR #681, opened 2023-04-12
Safe vector accesses in examples

magnatelee:
This PR replaces all direct index accesses to vectors with the `.at` calls, which would give us meaningful error messages instead of segfault on out-of-bound accesses. This also unifies the clang format configuration for the project.


examples/io/src/write_even_tiles.cc
```diff
@@ -49,10 +49,10 @@ class WriteEvenTilesTask : public Task<WriteEvenTilesTask, WRITE_EVEN_TILES> {
...
   static void cpu_variant(legate::TaskContext& context)
   {
-    auto dirname                           = context.scalars()[0].value<std::string>();
-    legate::Span<const int32_t> shape      = context.scalars()[1].values<int32_t>();
-    legate::Span<const int32_t> tile_shape = context.scalars()[2].values<int32_t>();
-    auto& input                            = context.inputs()[0];
+    auto dirname                           = context.scalars().at(0).value<std::string>();
+    legate::Span<const int32_t> shape      = context.scalars().at(0).values<int32_t>();
+    legate::Span<const int32_t> tile_shape = context.scalars().at(0).values<int32_t>();
+    auto& input                            = context.inputs().at(0);
```
manopapad, line 55:
```suggestion
    auto dirname                           = context.scalars().at(0).value<std::string>();
    legate::Span<const int32_t> shape      = context.scalars().at(1).values<int32_t>();
    legate::Span<const int32_t> tile_shape = context.scalars().at(2).values<int32_t>();
    auto& input                            = context.inputs().at(0);
```

manopapad (approved):

PR #684, opened 2023-04-12
Document read-only / env-only settings

bryevdv:
This PR add s support for documenting the env var settings that are read at the CPP level. 

@manopapad before adding tests and finishing up I wanted to ask a couple of preliminary questions:

* What do you want to do about defaults? Currently if the env var is not set, None is returned. If we want instead to return default value, we can, but (I assume) we would have to manually keep the Python value in sync with any CPP `getenv` counterparts. Is there any way we could at least arrange the CPP code so that we could ensure things match under test?
* Help strings are stubs, will need someone to supply better copy for them
* Does the readonly-ness need to be even more forceful e.g. with  `__set__` that raises a more informative exception, or a stronger note in the docs?

image

### DIFF
--- a/legate/settings.py
+++ b/legate/settings.py
@@ -14,7 +14,6 @@
 #
 from __future__ import annotations
 
-from .install_info import env_defaults_header
 from .util.settings import (
     EnvOnlySetting,
     PrioritizedSetting,
@@ -26,7 +25,9 @@ from .util.settings import (
 __all__ = ("settings",)
 
 
-def _extract_define(header: str, name: str) -> str:
+def _extract_define(name: str) -> str:
+    from .install_info import env_defaults_header as header
+
     lines = (line for line in header.split("\n") if line.startswith("#define"))
     for line in lines:
         tokens = line.split(" ")
@@ -89,12 +90,8 @@ class LegateRuntimeSettings(Settings):
     min_gpu_chunk: EnvOnlySetting[int] = EnvOnlySetting(
         "min_gpu_chunk",
         "LEGATE_MIN_GPU_CHUNK",
-        default=int(
-            _extract_define(env_defaults_header, "MIN_GPU_CHUNK_DEFAULT")
-        ),
-        test_default=int(
-            _extract_define(env_defaults_header, "MIN_GPU_CHUNK_TEST")
-        ),
+        default=int(_extract_define("MIN_GPU_CHUNK_DEFAULT")),
+        test_default=int(_extract_define("MIN_GPU_CHUNK_TEST")),
         convert=convert_int,
         help="""
         Minimum chunk size to enable GPU execution.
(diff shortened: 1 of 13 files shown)


bryevdv:
@manopapad I don't love all the ugly complication from trying to scrape the C header. If you tell me that these values are unlikely to change and that a comment in the header about updating the corresponding Python values would suffice, I would be convinced to just pass the default and test default values as literal and rip out all the complexity. 

Updated docs with defaults now look lik:

image

legate/settings.py
```diff
@@ -58,5 +64,134 @@ class LegateRuntimeSettings(Settings):
...
+        This is a read-only environment variable setting used by the runtime.
+        """,
+    )
+
+    precise_exception_trace: EnvOnlySetting[bool] = EnvOnlySetting(
+        "precise_exception_trace",
+        "LEGATE_PRECISE_EXCEPTION_TRACE",
+        default=False,
+        test_default=False,
+        convert=convert_bool,
```
manopapad, line 149: Of note here, the C++ code will not accept all the "truthy" values that `convert_bool` will accept (i.e. "yes", "Yes", "on" etc.), it will only accept 0 and 1 (and other integers I suppose, anything non-zero will count as "true").

We should either add a note specifically for C++-only boolean options, that they need to be passed strictly with 0/1, or add an `extract_env_bool` function to the C++ side that accepts the same strings as `convert_bool`, and use it when a boolean envvar is expected.

manopapad: Yes, that's reasonable. I would suggest also adding a comment on the top of the "settings" page saying that boolean values are passed through the environment as 0/1.

bryevdv: https://github.com/nv-legate/legate.core/pull/684/commits/4c4362708e84501450163f1c110274dd3f5c7ad2 and https://github.com/nv-legate/legate.core/pull/684/commits/6b8846ac642fc1bdcd18a8d82fdbca9426230fde made the comment show on every individual property:

image

manopapad (approved):
I found some more undocumented options in the C++ sources:
```
const char* show_usage = getenv("LEGATE_SHOW_USAGE");
parse_variable("LEGATE_SHOW_PROGRESS", show_progress_requested);
parse_variable("LEGATE_EMPTY_TASK", use_empty_task);
parse_variable("LEGATE_SYNC_STREAM_VIEW", synchronize_stream_view);
parse_variable("LEGATE_LOG_MAPPING", log_mapping_decisions);
```
We can handle these on a follow-up PR. Note that two of them are controlled by legate driver flags (`--mem-usage`, `--progress`). Once the corresponding env vars are documented, these flags should probably be removed, since they're legate.core library options, not launcher options.

bryevdv:
@marcinz any idea why these two tests are failing? They both say no "artifacts"

PR #687, opened 2023-04-14
Collective fix

ipdemes:
Fixing logic for choosing collective instances


ipdemes:
The collective test should be run with new `-spy-validate-collective` and `--spy-assert-warning` flags to check for errors in the logic

magnatelee:
> The collective test should be run with new -spy-validate-collective and --spy-assert-warning flags to check for errors in the logic

can you run cunumeric tests with these flags and see if you hit any errors?

ipdemes:
> > The collective test should be run with new -spy-validate-collective and --spy-assert-warning flags to check for errors in the logic
> 
> can you run cunumeric tests with these flags and see if you hit any errors?

@magnatelee : I tested all cunumeric examples and tests with new flags and got all tests to pass the check except for the `cholesky` test. It seems like runtime is adding one of the regions to the list of collective views when it shouldn't be added (legate.core doesn't add anything to the list of collective views).  I believe this issue is not related to this PR,  I will try to find a solution as a separate task

src/core/data/transform.h
```diff
@@ -51,6 +52,13 @@ struct TransformStack : public Transform {
  public:
   void dump() const;
 
+ public:
+  void return_promoted_dims(std::vector<int32_t>& dims)
+  {
+    if (nullptr != transform_) transform_->return_promoted_dims(dims);
+    if (nullptr != parent_) parent_->return_promoted_dims(dims);
```
manopapad, line 59: You probably need to process the parent before the current node in the stack, see comments in `transform.cc`.

src/core/data/transform.cc
```diff
@@ -239,6 +246,11 @@ void Project::print(std::ostream& out) const
 
 int32_t Project::target_ndim(int32_t source_ndim) const { return source_ndim + 1; }
 
+void Project::return_promoted_dims(std::vector<int32_t>& dims) const
+{
+  dims.push_back((-dim_ - 1));
```
manopapad, line 251: If you process the parent before the child, then I don't think you need to leave this "sentinel" value, you can just remove the projected-out dimension from the `promoted_dims` that the parent returned (if it's there), and shift all larger dims down by one.

src/core/data/transform.cc
```diff
@@ -301,6 +313,8 @@ void Transpose::print(std::ostream& out) const
 
 int32_t Transpose::target_ndim(int32_t source_ndim) const { return source_ndim; }
 
+void Transpose::return_promoted_dims(std::vector<int32_t>&) const {};
```
manopapad, line 316: I think we need to do something to handle this case, where dimensions are reordered. Consider that if dimension 1 of `A` is fake, then dimension 0 of `A.transpose([1,2,0])` is fake. See the mapper logs I get for this example:

```
out = cn.zeros((2,3,4))
x = cn.ones((4,3)) # shape=(4,3), promoted_dims=[]
y = x[:,cn.newaxis,:] # shape=(4,1,3), promoted_dims=[1]
z = y.transpose((1,2,0)) # shape=(1,3,4), promoted_dims=[0]
cn.exp(z, out=out)
```

```
MAP_TASK for cunumeric::UnaryOpTask(index_point=(0,1,2))<106> @ a.py:30
    Requirement[0](privilege=READ_ONLY,region=(2,(49,3),2),domain=<2,1>..<2,1>,fields=1000)
    Requirement[1](privilege=WRITE_DISCARD,region=(1,(16,1),1),domain=<0,1,2>..<0,1,2>,fields=1000)
MAP_TASK for cunumeric::UnaryOpTask(index_point=(1,1,2))<107> @ a.py:30
    Requirement[0](privilege=READ_ONLY,region=(2,(49,3),2),domain=<2,1>..<2,1>,fields=1000)
    Requirement[1](privilege=WRITE_DISCARD,region=(1,(17,1),1),domain=<1,1,2>..<1,1,2>,fields=1000)
```

Indeed, it is on dimension 0 of the launch space that the "collective" access of `x` is happening. 

I think you probably want to have processed the parent transform first, before you can shift the promoted dims around.

manopapad (changes requested):

src/core/mapping/base_mapper.cc
```diff
@@ -208,20 +209,37 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
+    auto store = legate_task.reductions()[i];
+    if (store.is_future()) continue;
+    auto idx = legate_task.reductions()[i].requirement_index();
+    auto req = task.regions[idx];
     if (req.privilege & LEGION_WRITE_PRIV) continue;
-    // Look up the projection for the input region. There are cases where
-    // Legate libraries register their own projection functors that are
-    // not recorded by Legate Core. So, handle the case when these functors
-    // are not present and allow for them to be missing.
     auto projection = find_legate_projection_functor(req.projection, true /* allow_mising */);
```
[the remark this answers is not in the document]
ipdemes: I thought we could avoid going through collecting promoted dims for reductions, since we only need to check if projection!=0 for them

src/core/mapping/base_mapper.cc
```diff
@@ -208,20 +209,35 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
     if ((req.handle_type == LEGION_SINGULAR_PROJECTION) ||
-        (projection != nullptr && projection->is_collective())) {
+        ((projection != nullptr) && (projection != 0))) {
```
manopapad, line 236: If all you need to do is check for non-identity projection functor, I think you just need to:

```
    if (req.handle_type == LEGION_SINGULAR_PROJECTION || req.projection != 0) {
```

And then presumably you don't need to do `find_legate_projection_functor` at all.

Notably, the `projection != 0` check that exists currently doesn't do anything, since you've already checked the equivalent  `projection != nullptr`.

src/core/data/transform.cc
```diff
@@ -239,6 +248,12 @@ void Project::print(std::ostream& out) const
 
 int32_t Project::target_ndim(int32_t source_ndim) const { return source_ndim + 1; }
 
+void Project::return_promoted_dims(std::vector<int32_t>& dims) const
+{
+  auto finder = std::find(dims.begin(), dims.end(), (dim_));
+  if (finder != dims.end()) { dims.erase(finder); }
```
manopapad, line 254: Note that you also need to "shift down" everything in `dims` that is larger than `dim_`.

src/core/data/transform.cc
```diff
@@ -185,6 +187,13 @@ void Promote::print(std::ostream& out) const
 
 int32_t Promote::target_ndim(int32_t source_ndim) const { return source_ndim - 1; }
 
+void Promote::return_promoted_dims(std::vector<int32_t>& dims) const
+{
+  // avoid pushing 2 times
+  auto finder = std::find(dims.begin(), dims.end(), (extra_dim_));
+  if (finder == dims.end()) { dims.push_back(extra_dim_); }
```
manopapad, line 194:
```suggestion
  for (auto& dim : dims) if (dim >= extra_dim_) dim++;
  dims.push_back(extra_dim_);
```

IIUC the added dimension is "bumping up" all the existing dimensions. Consider this example:

```
x = cn.ones((4,5)) # shape=(4,5), promoted_dims=[]
x = x[:,np.newaxis,:] # shape=(4,1,5), promoted_dims=[1]
x = x[np.newaxis,:,:,:] # shape=(1,4,1,5), promoted_dims=[0,2]
```

Therefore, I believe all existing promoted dims should remain, and adjusted up if they came after the added dim.

src/core/data/transform.cc
```diff
@@ -301,6 +316,19 @@ void Transpose::print(std::ostream& out) const
...
+{
+  for (size_t i = 0; i < axes_.size(); i++) {
+    if (i != axes_[i]) {
+      auto finder_dims = std::find(dims.begin(), dims.end(), (i));
+      if (finder_dims != dims.end()) {
+        auto finder_axes   = std::find(axes_.begin(), axes_.end(), (i));
+        dims[*finder_dims] = *finder_axes;
+      }
+    }
+  }
```
manopapad, line 329:
```suggestion
  // i should be added to X.tranpose(axes).promoted iff axes[i] is in X.promoted
  // e.g. X.promoted = [0] => X.transpose((1,2,0)).promoted = [2]
  for (auto& promoted : dims) {
    auto finder = std::find(axes_.begin(), axes_.end(), promoted);
#ifdef DEBUG_LEGATE
    assert(finder != axes_.end());
#endif
    promoted = finder - axes_.begin();
  }
```

The current code has the potential to blow up, because you're making multiple passes over the promoted dims, so the second pass has the potential to see the partial result from the first pass. I can't immediately spot a problem, but I think it's more safe and understandable if you iterate over the promoted dims instead.

src/core/data/transform.cc
```diff
@@ -376,6 +404,8 @@ int32_t Delinearize::target_ndim(int32_t source_ndim) const
   return source_ndim - strides_.size() + 1;
 }
 
+void Delinearize::return_promoted_dims(std::vector<int32_t>&) const {};
```
manopapad, line 407:
```suggestion
void Delinearize::return_promoted_dims(std::vector<int32_t>& dims) const
{
  bool delinearized_dim_was_promoted = false;
  for (auto& promoted : dims) {
    if (promoted == dim_) delinearized_dim_was_promoted = true;
    if (promoted > dim_) promoted += strides_.size() - 1;
  }
  if (delinearized_dim_was_promoted)
    // all the new dims are also therefore considered promoted
    for (int32_t i = 1; i < strides_.size(); ++i) dims.push_back(dim_ + i);
}
```

I think this is what needs to happen here, but @magnatelee should really confirm the logic here (and all the other transforms really).

src/core/data/transform.cc
manopapad: I don't think there's anything to do for `Shift`. I am not 100% sure what needs to be done for `Delinearize`, I took a crack at it, we should get Wonchan's confirmation though.

ipdemes: OK, I will ask Wonchan on slack.

src/core/data/transform.cc
ipdemes: yes, that is right

src/core/mapping/base_mapper.cc
ipdemes: that is right, thank you

src/core/data/transform.cc
ipdemes: I actually think the partitioning along of the new dimensions created after delinearization is not supported.
I have tried to execute following
```
  a = create_int64_store(shape=(100,))
   a=a.promote(0,10000)
   a=a.delinearize(0,[100,100])
   store_partition = store.partition_by_tiling(100,10,100) #partition by dim 1
   launch_shape = store_partition.partition.color_shape

    task = context.create_manual_task(
        OpCode.COLLECTIVE,
        launch_domain=Rect(launch_shape),
    )
    task.add_input(store_partition)
    task.execute()

```

and got error when from where legate is trying to create partition

tests/integration/collective/examples/test_collective.py
```diff
@@ -0,0 +1,127 @@
...
+#
+import pytest
+from collective import (
+    collective_test,
+    collective_test_matvec,
+    create_int64_store,
+)
+
+
+def test_no_collective() -> None:
```
magnatelee, line 23: formatting of this file is awful. I think we should go check the formatter configuration.

src/core/data/transform.cc
```diff
@@ -130,6 +130,8 @@ void Shift::print(std::ostream& out) const
 
 int32_t Shift::target_ndim(int32_t source_ndim) const { return source_ndim; }
 
+void Shift::return_promoted_dims(std::vector<int32_t>&) const {};
```
magnatelee, line 133: I don't think you need a semicolon at the end.

src/core/data/transform.cc
```diff
@@ -239,6 +248,17 @@ void Project::print(std::ostream& out) const
 
 int32_t Project::target_ndim(int32_t source_ndim) const { return source_ndim + 1; }
 
+void Project::return_promoted_dims(std::vector<int32_t>& dims) const
+{
+  auto finder = std::find(dims.begin(), dims.end(), (dim_));
```
magnatelee, line 253: do we need to wrap `dim_` in parentheses?

src/core/data/transform.cc
```diff
@@ -239,6 +248,17 @@ void Project::print(std::ostream& out) const
...
+  if (finder != dims.end()) {
+    dims.erase(finder);
+    for (size_t i = 0; i < dims.size(); i++) {
+      if (dims[i] > dim_) { dims[i] -= 1; }
+    }
```
magnatelee, line 258:
```suggestion
    for (auto& dim : dims)
      if (dim > dim_) --dim;
```

src/core/data/transform.h
```diff
@@ -30,7 +30,8 @@ struct Transform {
 
 struct StoreTransform : public Transform {
   virtual ~StoreTransform() {}
-  virtual int32_t target_ndim(int32_t source_ndim) const = 0;
+  virtual int32_t target_ndim(int32_t source_ndim) const         = 0;
+  virtual void return_promoted_dims(std::vector<int32_t>&) const = 0;
```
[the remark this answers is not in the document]
magnatelee: also, I feel this method could have a return value, instead of taking an inout argument.

src/core/data/transform.cc
magnatelee: also, don't you need to do this even when `dim_` wasn't found in the `dims`?

magnatelee: wouldn't this function return `{1}` for `store.promote(1, 10).project(0, 0).project(0, 0)`?

src/core/data/transform.h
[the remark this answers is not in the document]
ipdemes: I have changed the name to `find_imaginary_dims`.
However I don't think we can have a return value instead of the argument since we need to process `parent` transform before our transform here: https://github.com/ipdemes/legate.core/blob/collective/src/core/data/transform.h#L58-L59

src/core/data/transform.cc
ipdemes: Yes, we need to do `--dim` even if `dim_` is not found in `dims`

src/core/data/transform.h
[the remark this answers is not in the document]
magnatelee: that doesn't make sense. if `parent_` is `nullptr`, you're at the bottom of the stack and you just need to forward what's coming out of the `transform_`.

src/core/data/transform.cc
```diff
@@ -301,6 +318,19 @@ void Transpose::print(std::ostream& out) const
...
+  // i should be added to X.tranpose(axes).promoted iff axes[i] is in X.promoted
+  // e.g. X.promoted = [0] => X.transpose((1,2,0)).promoted = [2]
+  for (auto& promoted : dims) {
+    auto finder = std::find(axes_.begin(), axes_.end(), promoted);
+#ifdef DEBUG_LEGATE
+    assert(finder != axes_.end());
+#endif
+    promoted = finder - axes_.begin();
+  }
+};
```
magnatelee, line 332:
```suggestion
}
```

src/core/data/transform.cc
```diff
@@ -376,6 +406,8 @@ int32_t Delinearize::target_ndim(int32_t source_ndim) const
   return source_ndim - strides_.size() + 1;
 }
 
+void Delinearize::find_imaginary_dims(std::vector<int32_t>&) const {};
```
magnatelee, line 409:
```suggestion
void Delinearize::find_imaginary_dims(std::vector<int32_t>&) const {}
```

src/core/data/transform.h
```diff
@@ -51,6 +52,16 @@ struct TransformStack : public Transform {
  public:
   void dump() const;
 
+ public:
+  std::vector<int32_t> find_imaginary_dims() const
```
magnatelee, line 56: missed `override` keyword

src/core/data/transform.h
```diff
@@ -51,6 +52,16 @@ struct TransformStack : public Transform {
...
+  {
+    std::vector<int32_t> dims;
+
+    if (nullptr != parent_) { dims = parent_->find_imaginary_dims(); }
+    if (nullptr != transform_) transform_->find_imaginary_dims(dims);
+    return std::move(dims);
+  }
```
magnatelee, line 63: any reason to keep this in a header? please move this to a cc file

src/core/mapping/base_mapper.cc
```diff
@@ -18,6 +18,7 @@
 #include <sstream>
 #include <unordered_map>
 
+#include "legion.h"
```
magnatelee, line 21: I feel the existing headers already transitively include legion.h. did you actually need this to get the code compiled?

src/core/mapping/base_mapper.cc
```diff
@@ -135,23 +136,34 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
-  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
-    auto& req = task.regions[idx];
+  for (auto& store : legate_task.inputs()) {
+    if (store.is_future()) continue;
+    auto idx = store.requirement_index();
+    auto req = task.regions[idx];
+    if (req.privilege & LEGION_WRITE_PRIV) continue;
+    std::vector<int32_t> promoted_dims = store.find_imaginary_dims();
+    for (auto& d : promoted_dims) {
+      if (d < 0) continue;
```
magnatelee, line 148: this doesn't sound right. dimensions returned from `find_imaginary_dims` should never be negative.

src/core/mapping/base_mapper.cc
```diff
@@ -135,23 +136,34 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
+      if ((task.index_domain.hi()[d] - task.index_domain.lo()[d]) >= 1) {
```
magnatelee, line 149: you want to cache the domain points from `hi()` and `lo()`; they return copies, not references.

src/core/mapping/base_mapper.cc
```diff
@@ -135,23 +136,34 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
 {
+  Task legate_task(&task, context, runtime, ctx);
 #ifdef LEGATE_USE_COLLECTIVE
-  for (uint32_t idx = 0; idx < task.regions.size(); ++idx) {
-    auto& req = task.regions[idx];
+  for (auto& store : legate_task.inputs()) {
+    if (store.is_future()) continue;
+    auto idx = store.requirement_index();
+    auto req = task.regions[idx];
+    if (req.privilege & LEGION_WRITE_PRIV) continue;
```
magnatelee, line 145: input stores will always be mapped to read-only region requirements, so I don't think this check is necessary.

src/core/mapping/store.h
```diff
@@ -195,6 +195,13 @@ class Store {
    */
   Domain domain() const;
 
+ public:
+  std::vector<int32_t> find_imaginary_dims() const
+  {
```
magnatelee, line 200: let's move this body to the .cc file

tests/integration/collective/collective/collective.py
```diff
@@ -0,0 +1,89 @@
...
+
+class OpCode(IntEnum):
+    COLLECTIVE = user_lib.cffi.COLLECTIVE
+
+
+def create_int64_store(shape: Tuple[Any, ...]) -> Store:
+    store = context.create_store(ty.int64, shape=shape)
+    # call empty function on the store to make Legion to think
+    # that we initialized the store
+    task = context.create_auto_task(
```
magnatelee, line 33: you could have used fill instead

tests/integration/collective/collective/collective.py
```diff
@@ -0,0 +1,89 @@
...
+        OpCode.COLLECTIVE,
+    )
+    task.add_output(store)
+    task.execute()
+    return store
+
+
+def _broadcast(store: Store, shape: Tuple[Any, ...]) -> Store:
```
magnatelee, line 41: why do we need this general broadcast implementation here? I feel we could just directly transform the stores below.

tests/integration/collective/src/collective_test.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "collective_cffi.h"
+#include "legate_library.h"
+
+namespace collective {
+
+extern Legion::Logger logger;
```
magnatelee, line 8: if this logger isn't used, please remove it. I feel we can remove this header file.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+#=============================================================================
```
magnatelee, line 1: is this file necessary? if you meant to add this file, then I feel you should also update this line: https://github.com/nv-legate/legate.core/blob/branch-23.07/legate_core_cpp.cmake#L521

tests/integration/collective/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+set(CMAKE_CXX_STANDARD 17)
+set(BUILD_SHARED_LIBS ON)
+
+find_package(legate_core REQUIRED)
+
+legate_add_cpp_subdirectory(src TARGET collective  EXPORT collective-export)
+
+legate_add_cffi(${CMAKE_CURRENT_SOURCE_DIR}/src/collective_cffi.h TARGET collective)
+legate_python_library_template(collective)
```
magnatelee, line 33: note that the python library template is going to be deprecated, and it'd be easier to directly spell out boilerplates, which isn't that much at this point. I suggest you look at other recent integration tests that do that (e.g., https://github.com/nv-legate/legate.core/tree/branch-23.07/tests/integration/tree_reduce)

src/core/data/transform.h
ipdemes: I believe there should be no override since this method is unique to TrasformStack (has different arguments)

src/core/mapping/base_mapper.cc
ipdemes: Sorry, this is a leftover from a previous implementation

src/core/mapping/base_mapper.cc
```diff
@@ -134,23 +134,34 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
+    std::vector<int32_t> promoted_dims = store.find_imaginary_dims();
+    auto hi                            = task.index_domain.hi();
+    auto lo                            = task.index_domain.lo();
+    for (auto& d : promoted_dims) {
+      if ((hi[d] - lo[d]) >= 1) {
+        output.check_collective_regions.insert(idx);
+        break;
+      }
+    }
+  }
```
magnatelee, line 152:
```suggestion
    auto lo = task.index_domain.lo();
    auto hi = task.index_domain.hi();
    for (auto& store : legate_task.inputs()) {
    if (store.is_future()) continue;
    for (auto& d : store.find_imaginary_dims()) {
      if ((hi[d] - lo[d]) >= 1) {
        output.check_collective_regions.insert(idx);
        break;
      }
    }
  }
```

tests/integration/CMakeLists.txt
```diff
@@ -18,5 +18,6 @@
 # 'find' the current build tree to test package builds
 
 add_subdirectory(registry)
+add_subdirectory(collective)
```
magnatelee, line 21: let's keep the tests sorted alphabetically.

src/core/mapping/base_mapper.cc
```diff
@@ -134,23 +134,33 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
-    auto& req = task.regions[idx];
+  auto hi = task.index_domain.hi();
+  auto lo = task.index_domain.lo();
+  for (auto& store : legate_task.inputs()) {
+    if (store.is_future()) continue;
+    auto idx                           = store.requirement_index();
+    std::vector<int32_t> promoted_dims = store.find_imaginary_dims();
+    for (auto& d : promoted_dims) {
+      if ((hi[d] - lo[d]) >= 1) {
+        output.check_collective_regions.insert(idx);
```
magnatelee, line 147:
```suggestion
    std::vector<int32_t> promoted_dims = store.find_imaginary_dims();
    for (auto& d : promoted_dims) {
      if ((hi[d] - lo[d]) >= 1) {
        output.check_collective_regions.insert(store.requirement_index());
```

magnatelee: this is definitely a nitpick. in general I believe it's a good writing practice to put the definition close to the use, unless that leads to a lot of repetitions.

magnatelee (approved):

magnatelee:
@ipdemes left a minor comment. otherwise LGTM

PR #688, opened 2023-04-14
Fix CMake template directories to use current_dir for subfolders

jjwilke:


manopapad (approved):
That fixes my issue. Now the file is under
```
tests/integration/registry/registry/install_info.py
```
That still gets included in git untracked files. Might I suggest changing `.gitignore` from just `legate/install_info.py` to any `install_info.py` in general?

magnatelee (approved):

PR #694, opened 2023-04-19
Support PYTHONFAULTHANDLER

bryevdv:
This PR conditions the setting of `REALM_BACKTRACE` depending on whether `PYTHONFAULTHANDLER` is already set. 

@marcinz if we want to set `PYTHONFAULTHANDLER=1` but *only* for the "Eager" job, where can that change be made?

### DIFF
--- a/legate/driver/launcher.py
+++ b/legate/driver/launcher.py
@@ -257,7 +257,7 @@ class Launcher:
             env["LEGION_FREEZE_ON_ERROR"] = "1"
 
         # Debugging options
-        if system.env.get("PYTHONFAULTHANDLER", None) != "1":
+        if system.env.get("PYTHONFAULTHANDLER", "") != "":
             env["REALM_BACKTRACE"] = "1"
 
         if config.debugging.gasnet_trace:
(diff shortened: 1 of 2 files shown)


manopapad (approved):
One comment, looks good otherwise

marcinz:
I think that we could probably insert a test here: https://github.com/nv-legate/cunumeric/blob/branch-23.05/.github/workflows/ci.yml#L121

I think that we could test `${{ matrix.name }}` and export `PYTHONFAULTHANDLER=1` if the name is `eager`. I'll create a PR to test this out.

bryevdv:
Not entirely sure why the new `test_realm_backtrace_with_faulthandler` unit test is failing in CI, it passes for me locally.

legate/driver/launcher.py
```diff
@@ -257,7 +257,7 @@ def _compute_env(self) -> tuple[EnvDict, set[str]]:
             env["LEGION_FREEZE_ON_ERROR"] = "1"
 
         # Debugging options
-        if system.env.get("PYTHONFAULTHANDLER", None) != "1":
+        if system.env.get("PYTHONFAULTHANDLER", "") != "":
```
bryevdv, line 260: logic was reversed here

PR #698, opened 2023-04-21
Driver error if legate was compiled w/o required options

bryevdv:
This PR will make the driver error out with an informative RuntimeError if options like `--gpus` are requested, but the required support is not compiled into the build.


manopapad (approved):

bryevdv:
@manopapad @marcinz should build and tests be expected to run?

marcinz:
@bryevdv I know this is suboptimal, but could you make another commit and revert it? The tests just seem to be gone from the current commit.

PR #699, opened 2023-04-21
Make `tree_reduce` reuse the existing partition

magnatelee:
The `tree_reduce` operation hasn't been able to reuse the input's existing key partition, because key partitions for unbound stores haven't been properly recorded. This PR fixes the issue and thereby allows `tree_reduce` to reuse the existing partitions.


magnatelee:
CC @RAMitchell this is the fix I promised.

legate/core/launcher.py
```diff
@@ -975,12 +983,15 @@ def build_single_task(self, argbuf: BufferBuilder) -> SingleTask:
...
+    def execute(
+        self, launch_domain: Rect
+    ) -> tuple[FutureMap, dict[Store, LegionPartition]]:
         task = self.build_task(launch_domain, BufferBuilder())
         result = self._context.dispatch(task)
         assert isinstance(result, FutureMap)
         self._out_analyzer.update_storages()
-        return result
+        out_partitions = self._out_analyzer.collect_partitions()
+        return result, out_partitions
```
bryevdv, line 994: dataclass could be nice here rather than a tuple

bryevdv (approved):

manopapad (approved):

legate/core/launcher.py
magnatelee: https://github.com/nv-legate/legate.core/pull/699/commits/81bd87d21c59aaa6ad06f9db3b42dc5f8caf59d6

RAMitchell:
> CC @RAMitchell this is the fix I promised.

Thanks!

RAMitchell:
`tree_reduce` is still dividing 1 input into 4 for me, after applying this fix.

legate/core/operation.py
```diff
@@ -1463,21 +1470,26 @@ def launch(self, strategy: Strategy) -> None:
...
 
-            output = self._context.create_store(input.type)
+            num_tasks = (fan_in + self._radix - 1) // self._radix
+            fan_in = num_tasks
+            done = fan_in == 1
+
+            if done:
+                output = self._outputs[0]
+            else:
+                output = self._context.create_store(input.type)
```
bryevdv, line 1480: defining `done` seems superfluous since it is not used anywhere else. Suggest just doing 
```
    if fan_in == 1:
```
directly

legate/core/partition.py
```diff
@@ -430,3 +430,9 @@ def construct(
                 index_space, self, index_partition
             )
         return region.get_child(index_partition)
+
+    def import_partition(self, partition: LegionPartition) -> None:
+        index_partition = partition.index_partition
+        runtime.partition_manager.record_index_partition(
+            index_partition.parent, self, index_partition
```
bryevdv, line 437: It seems a little odd that this function takes both `index_partition` and `index_partition.parent` as individual parameters. Is that unavoidable?

bryevdv (approved):
just a few very minor suggestions

magnatelee:
> tree_reduce is still dividing 1 input into 4 for me, after applying this fix.

@RAMitchell that actually raises a question as to how `tree_reduce` behaves in that case. I assume your input to tree_reduce is created by a single task, right? do you want it to be not partitioned in that case? the current implementation uses the same partitioner for the very first input to find a reasonable partition if it doesn't have one already.

RAMitchell:
I have a first task that accepts data and compacts it into a sketch object. Each object gets serialised to an unbound store output. My issue is that tree_reduce is dividing these serialised objects up. If I write a buffer on one worker then it would need to stay together for it to be usable. If only one worker writes a buffer then only one buffer would appear in the tree reduce task - at least that is my hope.

magnatelee:
@RAMitchell pushed some changes to fix the issue. can you pull and try again?

legate/core/operation.py
```diff
@@ -810,12 +816,16 @@ def launch(self, strategy: Strategy) -> None:
...
-            result = launcher.execute(launch_domain)
+            res = launcher.execute(launch_domain)
+            result = res.future_map
+            out_partitions = res.output_partitions
         else:
             result = launcher.execute_single()
+            out_partitions = dict()
 
-        self._demux_scalar_stores(result, launch_domain)
+        self._demux_scalar_stores(result, out_partitions, launch_domain)
```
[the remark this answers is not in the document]
magnatelee: [473a832](https://github.com/nv-legate/legate.core/pull/699/commits/473a832f369229b882464d3e232bd40fa579eec2)

legate/core/operation.py
magnatelee: [473a832](https://github.com/nv-legate/legate.core/pull/699/commits/473a832f369229b882464d3e232bd40fa579eec2)

legate/core/partition.py
magnatelee: [473a832](https://github.com/nv-legate/legate.core/pull/699/commits/473a832f369229b882464d3e232bd40fa579eec2)

magnatelee:
@bryevdv I think I address your comments. Can you take another look?

bryevdv:
@magnatelee LGTM Python-wise

PR #707, opened 2023-05-01
Remove Context dependency from annotations

bryevdv:
This PR moves the provenance and annotation APIs to `runtime.py` and removes the explicit dependency on a context. A follow-on PR will split up `runtime.py` (now > 2k LOC) into sub-modules.


bryevdv:
@magnatelee @manopapad there were conflicts after the resource scope PR, please check the resolution looks good

magnatelee:
> @magnatelee @manopapad there were conflicts after the resource scope PR, please check the resolution looks good

LGTM

manopapad (approved):

PR #715, opened 2023-05-05
Unify env test flags

bryevdv:
This PR removes the codepaths that set `CUNUMERIC_TEST` from the `legate.tester`. Corresponding PR in cunumeric removes usage of this flag in favor or `LEGATE_TEST`: https://github.com/nv-legate/cunumeric/pull/922

### DIFF
--- a/legate/tester/stages/_linux/gpu.py
+++ b/legate/tester/stages/_linux/gpu.py
@@ -18,7 +18,7 @@ import time
 from typing import TYPE_CHECKING
 
 from ..test_stage import TestStage
-from ..util import CUNUMERIC_TEST_ENV, Shard, StageSpec, adjust_workers
+from ..util import Shard, StageSpec, adjust_workers
 
 if TYPE_CHECKING:
     from ....util.types import ArgList, EnvDict
@@ -50,7 +50,7 @@ class GPU(TestStage):
         self._init(config, system)
 
     def env(self, config: Config, system: TestSystem) -> EnvDict:
-        return dict(CUNUMERIC_TEST_ENV)
+        return {}
 
     def delay(self, shard: Shard, config: Config, system: TestSystem) -> None:
         time.sleep(config.gpu_delay / 1000)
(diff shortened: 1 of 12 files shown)


legate/tester/stages/_linux/gpu.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, config: Config, system: TestSystem) -> None:
         self._init(config, system)
 
     def env(self, config: Config, system: TestSystem) -> EnvDict:
-        return dict(CUNUMERIC_TEST_ENV)
```
manopapad, line 53: Not immediately relevant to this PR, but shouldn't this also be using UNPIN_ENV?

legate/tester/stages/util.py
```diff
@@ -27,11 +27,8 @@
 
 UNPIN_ENV = {"REALM_SYNTHETIC_CORE_MAP": ""}
 
-CUNUMERIC_TEST_ENV = {"CUNUMERIC_TEST": "1"}
-
 # Raise min chunk sizes for deferred codepaths to force eager execution
 EAGER_ENV = {
-    "CUNUMERIC_TEST": "0",
     "CUNUMERIC_MIN_CPU_CHUNK": "2000000000",
```
manopapad, line 32: Now the NumPy fallback is controlled by `LEGATE_TEST`. Because we set `LEGATE_TEST=1` unconditionally on all tests, the eager stage will now always disable the fallback, even through we're trying to set the min_chunk sizes high enough so we can force-enable it.

Therefore, I think we want to override `CUNUMERIC_DISABLE_FALLBACK=0` in this eager env (once that is an option, as suggested in https://github.com/nv-legate/cunumeric/pull/922).

bryevdv: This honestly seems worse than just keeping the status quo (no reduction in env vars, current usage stops working)

legate/tester/stages/_linux/gpu.py
bryevdv: I'm not sure cc @magnatelee

legate/tester/stages/util.py
manopapad: Yes, but only for the special "eager" case. The more other, more typical testing modes (cpu, gpu, openmp) now have one less variable.

bryevdv: https://github.com/nv-legate/legate.core/pull/715/commits/20b4e7c40d165fc704ef34c47bc092f51c17ce68 for update in cn PR

manopapad (approved):

PR #716, opened 2023-05-05
Fix a "file-not-found" bug during repeated editable installs

manopapad:
Committing the fix by @jjwilke


bryevdv (approved):

bryevdv:
Seems to resolve issue on perlmutter

jjwilke (approved):

PR #718, opened 2023-05-05
Ask cuTensor for detailed error messages

manopapad:


legate/driver/launcher.py
```diff
@@ -260,6 +260,9 @@ def _compute_env(self) -> tuple[EnvDict, set[str]]:
         if system.env.get("PYTHONFAULTHANDLER", "") == "":
             env["REALM_BACKTRACE"] = "1"
 
+        if "CUTENSOR_LOG_LEVEL" not in system.env:
+            env["CUTENSOR_LOG_LEVEL"] = "1"
```
bryevdv, line 264: could use `env.setdefault("CUTENSOR_LOG_LEVEL", "1")`

six / half-dozen

bryevdv (approved):

legate/driver/launcher.py
manopapad: I think `env` is not initialized from `system.env`, it starts out empty.

bryevdv: :+1: I read too quickly

PR #720, opened 2023-05-06
Add num_procs property

jjwilke:


legate/core/machine.py
```diff
@@ -168,6 +168,11 @@ def __eq__(self, other: object) -> bool:
                 return False
         return True
 
+    @property
+    def num_procs(self) -> int:
+        range = self.get_processor_range()
+        return range.high - range.low
```
magnatelee, line 174: Let's not duplicate the size calculation, as we may later change the range back to a closed interval.
```suggestion
        return len(range)
```

legate/core/machine.py
```diff
@@ -168,6 +168,11 @@ def __eq__(self, other: object) -> bool:
                 return False
         return True
 
+    @property
+    def num_procs(self) -> int:
```
magnatelee, line 172: Just curious, any reason that you find `machine.num_procs` better than `len(machine)`? Both would compute the same value.

jjwilke:
Not needed. `__len__` returns the same value.

PR #723, opened 2023-05-09
Add interface files to the manifest

marcinz:
We will install the interface files ending in the `.pyi` extension to enable proper mypy type checking.


marcinz:
@bryevdv After the CI is completed, I will double check that the file gets installed.

bryevdv:
:+1: if this does not work for some reason, then `global-include` is also worth a try.

PR #724, opened 2023-05-09
[DO NOT MERGE] Move away from flat-layout

bryevdv:
This PR moves the legate repo struct away from a [flat-layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) and towards (effectively) a [src-layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout).  ("Effectivley" since our top-level sub-folder is `python` rather than `src`)

I have tested locally with both editable and non-editable installs (running all cunumeric tests and stages) but this will require careful review and testing.

As an aside, `isort` seems to behave much better wrt to `legate` imports (not grouping with external/third-party)


magnatelee:
@trxcllnt @marcinz looks like we need to teach CI how to deal with the new directory structure. can relevant changes be isolated in this PR or does it need changes in the internal system?

marcinz:
> @trxcllnt @marcinz looks like we need to teach CI how to deal with the new directory structure. can relevant changes be isolated in this PR or does it need changes in the internal system?

So I see this problem:

```
-- Installing: $PREFIX/lib/cmake/legate_core/legate_core-config.cmake
  Running command Preparing metadata (pyproject.toml)
  /opt/conda/conda-bld/legate-core_1683753114061/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/lib/python3.11/site-packages/setuptools/dist.py:520: SetuptoolsDeprecationWarning: Invalid version: 'core_1683753114061'.
 ```
 
 I'm not sure that this is specifically related to CI. I'll have to look at this error and see what causes it.

bryevdv:
I merged and resolved the two seemingly trivial conflicts but it seems editable installs no longer work (build fails) locally for me

bryevdv:
I'm going to go ahead and close this as it has diverged from latest dev branches considerably. Most of the PR is simple file movement so we can try again whenever the time seems opportune to take on a somewhat disruptive change like this.

PR #725, opened 2023-05-12
Use Legion with CMake's native CUDA language

trxcllnt:
This PR updates legate.core to use a Legion branch that incorporates the changes in the [drop-find-cuda](https://gitlab.com/StanfordLegion/legion/-/merge_requests/739#6d9c9297ab43e0b6df8fe6898c740e1683f74b6b) PR.


scripts/build-install.sh
```diff
@@ -22,7 +22,7 @@ if [[ -n "$(which ninja)" ]]; then cmake_args+=" -GNinja"; fi
 cmake_args+="
 -D Legion_USE_CUDA=ON
 -D Legion_USE_OpenMP=ON
--D CMAKE_CUDA_ARCHITECTURES=NATIVE
+-D Legion_CUDA_ARCH=native
```
jjwilke, line 25: Are we encouraging users not to use `CMAKE_CUDA_ARCHITECTURES`?

install.py
```diff
@@ -445,6 +445,8 @@ def validate_path(path):
 -DLegion_EMBED_GASNet_CONFIGURE_ARGS="--with-ibv-max-hcas=8"
 """.splitlines()
 
+    if cuda:
+        cmake_flags += [f"-DLegion_CUDA_ARCH={arch}"]
```
jjwilke, line 449: I probably missed some discussion on this. Why the switch from `CMAKE_CUDA_ARCHITECTURES` to `Legion_CUDA_ARCH`?

legate_core_cpp.cmake
```diff
@@ -257,6 +233,14 @@ set_target_properties(legate_core
                       INTERFACE_POSITION_INDEPENDENT_CODE ON
                       LIBRARY_OUTPUT_DIRECTORY            lib)
 
+if(Legion_USE_CUDA)
+  set_property(TARGET legate_core PROPERTY CUDA_ARCHITECTURES ${Legion_CUDA_ARCH})
```
jjwilke, line 237: It seems like we now have a potential conflict between `CMAKE_CUDA_ARCHITECTURES` and `Legion_CUDA_ARCH`.  If someone were to specify `-DCMAKE_CUDA_ARCHITECTURES=80`, e.g., it will just be silently ignored because of this line.

Or is this okay because of the new line in Legion:
```
    if(CMAKE_CUDA_ARCHITECTURES)
      set(Legion_CUDA_ARCH ${CMAKE_CUDA_ARCHITECTURES})
```
?

scripts/build-install.sh
trxcllnt: `CMAKE_CUDA_ARCHITECTURES` applies to all CUDA targets in the configuration, but I've made Legion [set the `CUDA_ARCHITECTURES` property](https://gitlab.com/StanfordLegion/legion/-/merge_requests/739/diffs#6d9c9297ab43e0b6df8fe6898c740e1683f74b6b_507_503) on each target, then [expose](https://gitlab.com/StanfordLegion/legion/-/merge_requests/739/diffs#eb3b097f0162225e2fcae76f74c070e17c5552c2_37_28) its archs in its package config via `Legion_CUDA_ARCH`.

I updated legate.core to inherit the archs list that Legion uses (similar to what we do for `Legion_USE_CUDA` etc.), so that way we don't get into a situation where legate.core is compiled for an arch not supported by Legion.

The two reasons Legion is setting the `CUDA_ARCHITECTURES` property per target instead of setting `CMAKE_CUDA_ARCHITECTURES` globally is:
1. This prevents Legion's archs list doesn't leak out of Legion's CMakeLists.txt and affect unrelated targets. If someone is building Legion as a submodule (like we do in legate.core), setting `CMAKE_CUDA_ARCHITECTURES` globally can affect the rest of their build.
3. Legion includes a polyfill for `CUDA_ARCHITECTURES` that translates each arch to its equivalent `-gencode` flag in < CMake 3.18. We can either append the `-gencode` flags to the `CMAKE_CUDA_FLAGS` global or set them per-target via `target_compile_options()`, and the strong preference is to avoid touching the global.

legate_core_cpp.cmake
trxcllnt: Yeah right now if someone sets `-DCMAKE_CUDA_ARCHITECTURES=` (or has set it before `add_subdirectory(./legion)`), then I made Legion use it instead of `-DLegion_CUDA_ARCH=`.

It feels a bit backward at first since `-DLegion_CUDA_ARCH` seems more project-specific than `-DCMAKE_CUDA_ARCHITECTURES=`. But `Legion_CUDA_ARCH` is parsed (commas -> semis) and defaults to `all-major` if unset, so I wanted to have an "escape hatch" if someone wants to force an exact list of GPU architectures that skips (or resets) all of Legion's arch logic.

install.py
trxcllnt: See above.

legate_core_cpp.cmake
trxcllnt: Oh but you're right -- if `find_package(Legion)` succeeds, now legate.core will ignores the `-DCMAKE_CUDA_ARCHITECTURES` argument and use whatever archs Legion uses. This mirrors how if we find a Legion that built with `-DLegion_USE_CUDA=OFF`, legate.core uses that value even if the user configures with `-DLegion_USE_CUDA=ON`.

jjwilke: Hmm... I think that's fine. I guess might at least want a warning in that situation. Or do we want to FATAL_ERROR if someone tries to set `CMAKE_CUDA_ARCHITECTURES` in that situation and it's totally disallowed?

cmake/versions.json
```diff
@@ -6,8 +6,8 @@
       "git_tag" : "1.17.0"
     },
     "Legion": {
-      "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "915f481e5daa9485344866062441d4ce4d7fb365"
+      "git_url" : "https://gitlab.com/inlineptx/legion.git",
+      "git_tag" : "0de432af68dc8a8274dd5412945e946cea08f3bc"
```
manopapad, line 10: Just putting a reminder here to reset this before merging with mainline.

cmake/thirdparty/get_legion.cmake
```diff
@@ -189,25 +127,21 @@ function(find_or_configure_legion)
                                  "Legion_REDOP_COMPLEX ON"
                                  "Legion_GPU_REDUCTIONS OFF"
                                  "Legion_BUILD_RUST_PROFILER ON"
-                                 "Legion_SPY ${Legion_SPY}"
-                                 "Legion_USE_LLVM ${Legion_USE_LLVM}"
-                                 "Legion_USE_HDF5 ${Legion_USE_HDF5}"
-                                 "Legion_USE_CUDA ${Legion_USE_CUDA}"
-                                 "Legion_NETWORKS ${Legion_NETWORKS}"
-                                 "Legion_USE_OpenMP ${Legion_USE_OpenMP}"
```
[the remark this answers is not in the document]
trxcllnt: Yes!

trxcllnt:
Closing in favor of https://github.com/nv-legate/legate.core/pull/828.

PR #726, opened 2023-05-16
Fixes for resource scoping

magnatelee:
This PR fixes two bugs in the resource scoping code:

1) Node ranges weren't correctly computed to be half-open intervals
2) The base mapper was trying to find initial processors out of empty processor ranges

### DIFF
--- a/src/core/mapping/base_mapper.cc
+++ b/src/core/mapping/base_mapper.cc
@@ -165,12 +165,9 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
     LEGATE_ABORT;
   }
 
-  auto target      = legate_mapper_->task_target(legate_task, options);
-  auto local_range = machine.slice(target, machine_desc);
-#ifdef DEBUG_LEGATE
-  assert(!local_range.empty());
-#endif
-  output.initial_proc = local_range.first();
+  auto target = legate_mapper_->task_target(legate_task, options);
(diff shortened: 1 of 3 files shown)


src/core/mapping/base_mapper.cc
```diff
@@ -165,12 +165,9 @@ void BaseMapper::select_task_options(const Legion::Mapping::MapperContext ctx,
...
 
-  auto target      = legate_mapper_->task_target(legate_task, options);
-  auto local_range = machine.slice(target, machine_desc);
-#ifdef DEBUG_LEGATE
-  assert(!local_range.empty());
-#endif
-  output.initial_proc = local_range.first();
+  auto target = legate_mapper_->task_target(legate_task, options);
+  // The initial processor just needs to have the same kind as the eventual target of this task
+  output.initial_proc = machine.procs(target).front();
```
jjwilke, line 170: Is there a scenario in which even this might be empty?

jjwilke (approved):
LGTM. Appears to be working in tests. 

One nit question.

jjwilke (changes requested):
New segfault... hang on...

src/core/mapping/base_mapper.cc
magnatelee: no it can't. that should have already been caught at this line: https://github.com/nv-legate/legate.core/blob/branch-23.05/legate/core/context.py#L177

PR #727, opened 2023-05-16
Use the NCCL workaround unconditionally

magnatelee:


manopapad (approved):
Sounds good to me. However, I would note @lightsighter has suggested that by denoting an index launch as concurrent (which we always do for launches that request a NCCL communicator) we should automatically be guaranteed that no point task will begin until all point tasks' preconditions have been satisfied, making the barrier workaround presumably unnecessary.

magnatelee:
> However, I would note @lightsighter has suggested that by denoting an index launch as concurrent (which we always do for launches that request a NCCL communicator) we should automatically be guaranteed that no point task will begin until all point tasks' preconditions have been satisfied, making the barrier workaround presumably unnecessary.

@manopapad Do you mean that concurrent launches should already have obviated the workaround or that @lightsighter suggested he would modify the concurrent task launch to guarantee that actually? Note that the workaround has been needed despite index launches for NCCL tasks being tagged concurrent.

manopapad:
> concurrent launches should already have obviated the workaround

That is what I was led to believe based on this slack discussion from May 2nd:

> Sean Treichler
> my understanding was that there was a way in legion to say that a given index task is doing collective stuff and make sure no point in the index task launches before everybody's preconditions are satisfied?
> Michael Bauer
> That is what the concurrent index space task launches are doing

lightsighter:
>  Note that the workaround has been needed despite index launches for NCCL tasks being tagged concurrent.

So here's the rub. If you do a concurrent index space task launch AND all the CUDA work that happens inside those point tasks of the collective index space task launch are local to the GPU that they are mapped to, then you are safe and you should not encounter any deadlocks. If however, you do anything in those point tasks that requires interactions with other GPUs inside the concurrent point tasks, then you don't have a forward progress guarantee. Things that can induce implicit synchronization between two GPUs in CUDA include, but are not limited to:

* Issuing a cudaMemcpy to/from another GPU
* Issuing a cudaMalloc (which synchronizes between all the GPUs in the process)

Legion can only protect you at the task boundaries from GPU work that it (really Realm) does before the point tasks of a concurrent index space task run. Once they start running, it's up to the client to make sure those tasks don't do any kind of work which might induce implicit synchronization between the GPUs before/after NCCL kernel calls. If they are going to do implicitly synchronizing calls between the GPUs then they need to put in a barrier before calling NCCL to make sure that all the synchronizing work is done before starting any NCCL calls.

PR #730, opened 2023-05-17
Make sure local fields are not enabled for any Python interpreter

magnatelee:
We've been suppressing reservation for local fields in each field space by [explicitly passing `-lg:local 0`](https://github.com/nv-legate/legate.core/blob/branch-23.05/legate/driver/command.py#L217), but never been doing the same in the code path for the canonical Python interpreter. So, any code that tries to have more than 252 active stores crashes with the canonical Python interpreter. This PR fixes the issue by just configuring Legion with no local fields. This PR also bumps up the commit hash for Legion, as [Legion's cmake build also had to be updated](https://gitlab.com/StanfordLegion/legion/-/commit/bc69f142bfffba8781c999dc0637ebbc07b2cd10) for this.

### DIFF
--- a/cmake/versions.json
+++ b/cmake/versions.json
@@ -7,7 +7,7 @@
     },
     "Legion": {
       "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "9663e2a0d9f26259dc7ecdac63925c0b5d7408b2"
+      "git_tag" : "83a417c848ba5732a53b870bb0119f23d979a9d0"
     }
   }
 }
--- a/cmake/versions.json
+++ b/cmake/versions.json
@@ -7,7 +7,7 @@
     },
     "Legion": {
       "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "bc69f142bfffba8781c999dc0637ebbc07b2cd10"
+      "git_tag" : "9663e2a0d9f26259dc7ecdac63925c0b5d7408b2"
     }
   }
 }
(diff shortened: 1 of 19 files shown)


magnatelee:
@trxcllnt here's a cmake build change for legion: https://gitlab.com/StanfordLegion/legion/-/commit/bc69f142bfffba8781c999dc0637ebbc07b2cd10 can you take a look at it? one peculiar thing about cmake that I figured out was that setting `0` to a variable seems to have an effect of unsetting it, so the `0` value doesn't get propagated to `legion_defines.h.in` but it un-defines that macro. so I had to wrap it with parentheses so it can survive, but it looks like our CI seems to complain about the zero value in a different place.

cmake/versions.json
```diff
@@ -7,7 +7,7 @@
     },
     "Legion": {
       "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "0ff4a18c08590c28bf3e416581790ceee974665c"
+      "git_tag" : "bc69f142bfffba8781c999dc0637ebbc07b2cd10"
```
manopapad, line 10: Can you please bump to 9663e2a0d9f26259dc7ecdac63925c0b5d7408b2, to include a fix for the CI failures seen in https://github.com/nv-legate/cunumeric/pull/932?

magnatelee: https://github.com/nv-legate/legate.core/pull/730/commits/f465ea737bda2c2a3eb2c640644daf49734e56a5

magnatelee:
>  but it looks like our CI is complaning about the zero value in a different place.
@trxcllnt I figured out what's causing the build issue.

PR #731, opened 2023-05-17
Fix bugs in corner cases of `tree_reduce`

magnatelee:
`tree_reduce` wasn't correctly handling cases where the input is produced by a single task.


csadorf (approved):
Looks good and can confirm that it resolves the issue I faced.

PR #735, opened 2023-05-23
Don't pass UCX options when we know we're not using it

manopapad:


legate/driver/command.py
```diff
@@ -413,7 +413,10 @@ def cmd_eager_alloc(
 def cmd_ucx(
     config: ConfigProtocol, system: System, launcher: Launcher
 ) -> CommandPart:
-    return ("-ucx:tls_host", "rc,tcp,cuda_copy,cuda_ipc,sm,self")
+    if config.multi_node.ranks > 1 and "ucx" in install_info.networks:
+        return ("-ucx:tls_host", "rc,tcp,cuda_copy,cuda_ipc,sm,self")
```
SeyedMir, line 417: This is not needed anymore.

legate/driver/launcher.py
```diff
@@ -186,16 +187,18 @@ def _compute_env(self) -> tuple[EnvDict, set[str]]:
...
-        env["UCX_RC_TX_POLL_ALWAYS"] = "y"
-
-        # Link to the UCX bootstrap plugin, in case Realm is using UCX
-        env["REALM_UCP_BOOTSTRAP_PLUGIN"] = str(
-            system.legion_paths.legion_lib_path / "realm_ucp_bootstrap_mpi.so"
-        )
+        if config.multi_node.ranks > 1 and "ucx" in install_info.networks:
+            # UCX-related environment variables
+            env["UCX_CUDA_COPY_MAX_REG_RATIO"] = "1.0"
+            env["UCX_MULTI_LANE_MAX_RATIO"] = "1.0"
```
SeyedMir, line 193: Let's remove this one too, i.e., not set `UCX_MULTI_LANE_MAX_RATIO`

legate/driver/command.py
bryevdv: Just clarifying: `cmd_ucx` can then be removed entirely?

legate/driver/launcher.py
```diff
@@ -186,16 +187,18 @@ def _compute_env(self) -> tuple[EnvDict, set[str]]:
...
+            env["UCX_IB_RCACHE_PURGE_ON_FORK"] = "n"
+            env["UCX_RC_TX_POLL_ALWAYS"] = "y"
+
+            # Link to the UCX bootstrap plugin
+            env["REALM_UCP_BOOTSTRAP_PLUGIN"] = str(
+                system.legion_paths.legion_lib_path
+                / "realm_ucp_bootstrap_mpi.so"
+            )
```
bryevdv, line 201: Maybe time to think about splitting `_compute_env` up into `env_nccl`, `env_ucc`, etc but no need to do that in this PR.

bryevdv (approved):
LGTM modulo question about removing `cmd_ucx`

legate/driver/command.py
SeyedMir: yes, at this point we can remove `cmd_ucx` everywhere in that file.

manopapad: 9d95834957301e54eaa9bfecc25e407e7d3cf6fe

legate/driver/launcher.py
manopapad: 9d95834957301e54eaa9bfecc25e407e7d3cf6fe

PR #737, opened 2023-05-25
Load Legate .so's using RTLD_GLOBAL, so they're included in dlopen(NULL)

manopapad:
@rohany please confirm this fixes your LLVM JIT loading issue


rohany:
yup, this works

PR #739, opened 2023-05-26
Add MPI escape hatch flag

bryevdv:
This PR adds an env-only setting `LEGATE_DISABLE_MPI` that can be used on systems where MPI cannot currently be supported, e.g. perlmutter. @manopapad I have verified that the example runs on perlmutter with this patch. OPen questions:

* help-text for the setting
* exception type/message for comms
* is runtime conditional OK in the cpp code? not sure any alternative


manopapad (changes requested):
Can we add a `has_cpu_communicator` to the `Runtime` class (and one in `Context` that falls back to `Runtime.has_cpu_communicator`, similar to `add_cpu_communicator`)? Then cuNumeric can attempt to handle this situation gracefully (which we can do in a follow-up PR).

bryevdv:
>  add a has_cpu_communicator to the Runtime

https://github.com/nv-legate/legate.core/pull/739/commits/01bca89e4b85015490334e80f91bb085e3a9929f

manopapad (approved):

PR #742, opened 2023-05-31
Enable symbols on REALM_BACKTRACE through libdw

manopapad:
This is enabling the recently-added Realm functionality, that prints out file/line/symbol information on backtraces by default:

```
  [0] = unknown symbol at unknown file:0 [0000153c9392608f]
  [1] = raise at ../sysdeps/unix/sysv/linux/raise.c:51 [0000153c9392600b]
  [2] = abort at /build/glibc-SzIz7B/glibc-2.31/stdlib/abort.c:79 [0000153c93905858]
  [3] = __assert_fail_base.cold at /build/glibc-SzIz7B/glibc-2.31/assert/assert.c:92 [0000153c93905728]
  [4] = __assert_fail at /build/glibc-SzIz7B/glibc-2.31/assert/assert.c:101 [0000153c93916fd5]
  [5] = void cunumeric::contract<float>(float*, unsigned long, long*, long*, int*, float const*, unsigned long, long*, long*, int*, float const*, unsigned long, long*, long*, int*, bool) at /home/scratch.mpapadakis_sw/py-legate/cunumeric/src/cunumeric/matrix/contract.cu:88 [00001537c740a09c]
  [6] = cunumeric::ContractImplBody<(cunumeric::VariantKind)2, (legate::Type::Code)10>::operator()(float*, unsigned long, long*, long*, int*, float const*, unsigned long, long*, long*, int*, float const*, unsigned long, long*, long*, int*, bool) at /home/scratch.mpapadakis_sw/py-legate/cunumeric/src/cunumeric/matrix/contract.cu:213 [00001537c7409d5b]
  [7] = void cunumeric::ContractImpl<(cunumeric::VariantKind)2>::operator()<(legate::Type::Code)10, 4, (void*)0>(cunumeric::ContractArgs&) const at /home/scratch.mpapadakis_sw/py-legate/cunumeric/src/cunumeric/matrix/contract_template.inl:169 [00001537c7415214]
  [8] = decltype(auto) legate::inner_type_dispatch_fn<4>::operator()<cunumeric::ContractImpl<(cunumeric::VariantKind)2>, cunumeric::ContractArgs&>(legate::Type::Code, cunumeric::ContractImpl<(cunumeric::VariantKind)2>, cunumeric::ContractArgs&) at /home/scratch.mpapadakis_sw/py-legate/legate.core/src/core/utilities/dispatch.h:64 [00001537c740ac0c]
  [9] = decltype(auto) legate::double_dispatch<cunumeric::ContractImpl<(cunumeric::VariantKind)2>, cunumeric::ContractArgs&>(int, legate::Type::Code, cunumeric::ContractImpl<(cunumeric::VariantKind)2>, cunumeric::ContractArgs&) at /home/scratch.mpapadakis_sw/py-legate/legate.core/src/core/utilities/dispatch.h:172 [00001537c740ad76]
```

This currently only works on Linux, because the implementation depends on some ELF-specific functionality.

The required libraries are available on conda, so for simplicity I made them a hard requirement, and I'm enabling this mode unconditionally on Linux. If people disagree, we could try to be nicer and check for the requirements before enabling this feature, but I would request @trxcllnt for guidance if we want to do that.

@m3vaz I tried adding these libraries to conda's build requirements only when building under Linux, not sure I did it right. Also note that `libdw` is dynamically-linked, so not sure if it should be under `runtime` rather than `build`.


trxcllnt (approved):

m3vaz:
@manopapad Because you need to link against it, it needs to be present in the build environment and so is a build requirement.

`libdw` likely adds the run requirement automatically via run_exports, so when you include it as a build requirement, conda-build will automatically know to list it as a run requirement as well.

manopapad:
@trxcllnt it appears that [FindLIBDW.cmake](https://github.com/StanfordLegion/legion/blob/control_replication/cmake/FindLIBDW.cmake) in Legion is not finding the `libdw` under `$CONDA_PREFIX`, most likely the fault of `find_path`/`find_library`: https://github.com/nv-legate/legate.core/actions/runs/5446823583/jobs/9911450722?pr=742#step:9:693.

What is the best way to fix this? Update `FindLIBDW.cmake` to consider `$CONDA_PREFIX` as well? Figure this out during legate.core build and pass it to the Legion build through `LIBDW_PATH` before calling into `get_legion.cmake`?

conda/conda-build/meta.yaml
```diff
@@ -98,6 +98,8 @@ requirements:
     - cmake {{ cmake_version }}
     - {{ compiler('c') }} =11.2
     - {{ compiler('cxx') }} =11.2
+    - elfutils # [linux]
```
sandeepd-nv, line 101: CMake is not able to find libdw during the conda package build process.

manopapad: Can you confirm that packages `libdwarf` and `elfutils` are included in the conda enviroment, and it is activated before the build starts?

PR #745, opened 2023-06-05
Opt-out validation for C++ accessor types and dimension

RAMitchell:
Check the type and dimension of C++ accessors. Adds a template parameter to optionally disable checking.

In case the type is not primative, check the size of the type.

It would be nice to also do this for reduce accessors, but reduce operators have a LHS and RHS type and I'm not sure if its correct to test one of these types.


RAMitchell:
I would have liked to write some tests for this, but not sure how in the current organisation of tests.

jjwilke (changes requested):
I like the change. Thanks for putting this together.

I think this changes the existing behavior in a way we wouldn't want? Double check me?

RAMitchell:
It's intended to change behavior to increase type safety. If a type mismatch occurs it seems much more likely to me to be a programmer mistake than intended, hence the opt in instead of opt out behavior.

jjwilke (approved):
I guess the check should be cheap and only done once per task. I'm fine with the change in existing behavior.

src/core/data/store.h
```diff
@@ -373,37 +377,42 @@ class Store {
...
    * @return A read-only accessor to the store
    */
-  template <typename T, int32_t DIM>
+  template <typename T, int32_t DIM, bool VALIDATE_TYPE = true>
   AccessorRO<T, DIM> read_accessor(const Rect<DIM>& bounds) const;
   /**
-   * @brief Returns a write-only accessor to the store for the entire domain
+   * @brief Returns a write-only accessor to the store for the entire domain. Validates type and
+   dimension by default, disable with VALIDATE_TYPE = false.
+
```
manopapad, line 394:
```suggestion
   *
```

missing asterisk

manopapad, line 394:
```suggestion
   * dimension by default, disable with VALIDATE_TYPE = false.
   *
```

missing asterisks

src/core/data/store.h
```diff
@@ -373,37 +377,42 @@ class Store {
...
    * @return A write-only accessor to the store
    */
-  template <typename T, int32_t DIM>
+  template <typename T, int32_t DIM, bool VALIDATE_TYPE = true>
   AccessorWO<T, DIM> write_accessor(const Rect<DIM>& bounds) const;
   /**
-   * @brief Returns a read-write accessor to the store for the entire domain
+   * @brief Returns a read-write accessor to the store for the entire domain. Validates type and
+   dimension by default, disable with VALIDATE_TYPE = false.
+
```
manopapad, line 407: same as above, missing asterisks

src/core/data/store.inl
```diff
@@ -269,12 +269,40 @@ void UnboundRegionField::bind_data(Buffer<T, DIM>& buffer, const Point<DIM>& ext
...
+      this->type().to_string().c_str());
+    LEGATE_ABORT;
+  }
+
+  // Test size matches for other types
+  if (sizeof(T) != this->type().size()) {
+    log_legate.error(
+      "Type size mismatch: store type %s has size %d, requested type has size %d. Disable type "
+      "checking via "
+      "accessor template parameter if this is intended.",
```
manopapad, line 291: I think these two strings can be combined

manopapad (approved):

RAMitchell:
Can this be merged?

src/core/data/store.inl
```diff
@@ -333,12 +363,13 @@ AccessorRD<OP, EXCLUSIVE, DIM> Store::reduce_accessor() const
...
 }
 
-template <typename T, int DIM>
+template <typename T, int DIM, bool VALIDATE_TYPE>
 AccessorRO<T, DIM> Store::read_accessor(const Rect<DIM>& bounds) const
 {
-#ifdef DEBUG_LEGATE
-  check_accessor_dimension(DIM);
-#endif
+  if constexpr (VALIDATE_TYPE) {
```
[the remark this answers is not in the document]
magnatelee: I guess it wouldn't be too bad to turn it on by default even in release mode, as it'd only introduce a couple of checks

src/core/data/store.inl
```diff
@@ -269,12 +269,39 @@ void UnboundRegionField::bind_data(Buffer<T, DIM>& buffer, const Point<DIM>& ext
...
+  // Test size matches for other types
+  if (sizeof(T) != this->type().size()) {
+    log_legate.error(
+      "Type size mismatch: store type %s has size %d, requested type has size %d. Disable type "
+      "checking via accessor template parameter if this is intended.",
+      this->type().to_string().c_str(),
+      this->type().size(),
+      sizeof(T));
+    LEGATE_ABORT;
+  }
```
magnatelee, line 295:
```suggestion
  auto in_type = legate_type_code_of<T>;
  if (in_type == this->code()) return;
  // Test exact match for primitive types
  if (in_type != Type::Code::INVALID) {
    log_legate.error(
      "Type mismatch: %s accessor to a %s store. Disable type checking via "
      "accessor template parameter if this is intended.",
      typeid(T).name(),
      this->type().to_string().c_str());
    LEGATE_ABORT;
  }

  // Test size matches for other types
  if (sizeof(T) != this->type().size()) {
    log_legate.error(
      "Type size mismatch: store type %s has size %d, requested type has size %d. Disable type "
      "checking via accessor template parameter if this is intended.",
      this->type().to_string().c_str(),
      this->type().size(),
      sizeof(T));
    LEGATE_ABORT;
  }
```

magnatelee: suggested a minor tweak to optimize the common case.

src/core/data/store.inl
```diff
@@ -269,12 +269,39 @@ void UnboundRegionField::bind_data(Buffer<T, DIM>& buffer, const Point<DIM>& ext
...
+template <typename T>
+void Store::check_accessor_type() const
+{
+  auto in_type = legate_type_code_of<T>;
+  // Test exact match for primitive types
+  if (in_type != this->code() && in_type != Type::Code::INVALID) {
+    log_legate.error(
+      "Type mismatch: %s accessor to a %s store. Disable type checking via "
+      "accessor template parameter if this is intended.",
+      typeid(T).name(),
```
magnatelee, line 281: @RAMitchell have you actually tried to exercise this branch? I'd make sure `typeid` prints out the right type name. (I believe it should, but I also vaguely remember `typeid(T).name()` returned a weird string when `T` was one of the types defined in stdint.h.)

RAMitchell: I've been using this branch for a couple weeks successfully. The name is implementation defined, you get something like 'f' for float on gcc. Let me see if I can convert it to a legate type then print it for consistency.

RAMitchell:
Test failures appear unrelated as no python code was changed.

magnatelee:
@RAMitchell can you rebase this on the latest branch-23.07? I know your changes have nothing to do with those failures, but I want to make sure everything is green after the merge.

magnatelee:
@RAMitchell looks like all tests are now passing, so feel free to merge this PR. (I believe you're now a member of the group, so you should have permission.)

PR #746, opened 2023-06-05
Gh workflow unit tests

sandeepd-nv:


m3vaz (commented):
LGTM. The change as is should be ok to go. We will eventually need to sort this into a reusable workflow, but for now this should be ok.

We do have some failures in the checks, and we need to either waive those tests, fix them so that they aren't so dependent on the environment, or provide whatever environment it seems to be expecting.

Will mark as approved once we have resolution on the failing tests, but no further change should be required here.

m3vaz:
The two failing tests in the CI are a missing environment variable and an assertion error.
A fix for the assertion error has already gone in with https://github.com/nv-legate/legate.core/pull/747

PR #751, opened 2023-06-06
Remove dependency on pyarrow

manopapad:
This is no longer used in Legate, and the `libarrow` conda package (required by `pyarrow`) is actually causing issues when building GASNet, because it pulls `ucx` (required for Arrow's `feather` component, which we don't use, but is bundled in the `libarrow` package regardless), which nowadays pulls `rdma-core`, whose `libibverbs.so` conflicts with the system version.


m3vaz (approved):
LGTM Thanks!

magnatelee (approved):

PR #753, opened 2023-06-07
Locking cuda-cudart to a major version

m3vaz:
Libraries are linking to cudart.so.11 so we need to lock the run time dependency to the major version being used during compilation


m3vaz:
This wouldn't have been a concern because previously we would have resolved against cudatoolkit 11 as well as cuda-cudart 12, which would have always given us cudart 11 from cudatoolkit

m3vaz:
Showing passing on the unit tests with the major rev lock, will merge once remaining CI jobs finish.

PR #756, opened 2023-06-07
Add support for test timeouts

bryevdv:
This PR adds an new `--timeout` option to the tester code that can be used to diagnose infinite hangs. When `--timeout` is specified, and a test exceeds this amount, it will be terminated, and report as `[TIME]` in the running test log. All timed-out tests are also reported in one place at the end:

Additionally, this PR adds support to display per-test time information:

image

## Notes

This feature is probably best used as a diagnostic tool for specific situations, rather than enabled by default. Given the extreme ranges of hardware and tests that may be encountered, there is not a reasonable "global default". Instead, when a hanging test is encountered, the `--timeout` may be set in order to determine what test is hanging. 

Lastly, if GPU tests timeout it appears that some subsequent tests run before resources are released, resulting in spurious `(exit: 250)` failures. Perhaps adding a delay after a timeout could remedy this.

### DIFF
--- a/legate/tester/test_system.py
+++ b/legate/tester/test_system.py
@@ -93,7 +93,7 @@ class TestSystem(System):
         *,
         env: EnvDict | None = None,
         cwd: str | None = None,
-        timeout: int | None = 10,
+        timeout: int | None = None,
     ) -> ProcessResult:
         """Wrapper for subprocess.run that encapsulates logging.
 
--- a/legate/tester/test_system.py
+++ b/legate/tester/test_system.py
@@ -22,7 +22,7 @@ import multiprocessing
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from subprocess import PIPE, STDOUT, run as stdlib_run
+from subprocess import PIPE, STDOUT, TimeoutExpired, run as stdlib_run
 from typing import Sequence
 
 from ..util.system import System
@@ -48,11 +48,18 @@ class ProcessResult:
     #: Whether this process was actually invoked
     skipped: bool = False
 
+    #: Whether this process timed-out
+    timeout: bool = False
+
     #: The returncode from the process
     returncode: int = 0
 
     #: The collected stdout and stderr output from the process
     output: str = ""
+
+    @property
+    def passed(self) -> bool:
+        return self.returncode == 0 and not self.timeout
 
 
 class TestSystem(System):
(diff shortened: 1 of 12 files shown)


legate/tester/test_system.py
```diff
@@ -82,6 +93,7 @@ def run(
         *,
         env: EnvDict | None = None,
         cwd: str | None = None,
+        timeout: int | None = 10,
```
manopapad, line 96: Doesn't really matter, since we always set this explicitly, but maybe set the default value to `None`, just so it doesn't clash with the default behavior (no default timeout) higher up in the stack?

legate/tester/test_system.py
```diff
@@ -117,13 +129,25 @@ def run(
...
+                cmd,
+                cwd=cwd,
+                env=full_env,
+                stdout=PIPE,
+                stderr=STDOUT,
+                text=True,
+                timeout=timeout,
+            )
+        except TimeoutExpired:
+            return ProcessResult(invocation, test_file, timeout=True)
```
manopapad, line 144: Does it make sense to save the timed-out test's output (and present it on the final output), in case it's useful to find out what specific function hung?

manopapad (approved):
Some small suggestions, otherwise looks good

legate/tester/test_system.py
bryevdv: It's not available, AFAIK
```
In [10]: try:
    ...:     run(["yes"], timeout=1)
    ...: except Exception as e:
    ...:     print(e.stderr)

y
y
y
y
y
None
```
same for `stdout` on the exception.

bryevdv: Unfortunately that won't work with the way we currently run processes:
```
In [3]: try:
   ...:     proc = run(["yes"], stdout=PIPE, stderr=STDOUT, text=True, timeout=1)
   ...: except Exception as e:
   ...:     print(proc.stdout)
   ...: 

TimeoutExpired: Command '['yes']' timed out after 1 seconds

During handling of the above exception, another exception occurred:

NameError                                 Traceback (most recent call last)
Cell In[3], line 4
      2     proc = run(["yes"], stdout=PIPE, stderr=STDOUT, text=True, timeout=1)
      3 except Exception as e:
----> 4     print(proc.stdout)

NameError: name 'proc' is not defined
```
Maybe it's possible to re-work things to use `capture_output=True` instead of pipes but maybe that could be investigated as a follow-on.

manopapad: Yup, an investigation for another day

manopapad (approved):

eddy16112 (approved):

PR #759, opened 2023-06-08
Rev up the version in cmake

m3vaz:


trxcllnt (changes requested):
Why is the entire file a diff? Did the line endings change or something?

m3vaz:
Reverted and re-did the change

PR #760, opened 2023-06-12
Use cuda-version to constrain CTK version pulled from conda

manopapad:


m3vaz:
12.1 libraries are not yet available on conda-forge, you might have some luck looking at the `nvidia`

PR #763, opened 2023-06-15
Mapper unification

magnatelee:
Some of the APIs have been taking a context argument simply because the core mapper can't handle certain operation types. This PR changes the core mapper so it'd be a Legate mapper and not a full-blown Legion mapper on its own. As a result, the APIs taking a context argument for no good reason become context free.


manopapad (changes requested):
Still looking through the python code, but submitting the C++ part of the review to enable cherry-picking.

src/core/mapping/core_mapper.cc
```diff
@@ -139,270 +89,88 @@ CoreMapper::CoreMapper(Legion::Mapping::MapperRuntime* rt,
...
     case LEGATE_CORE_TUNABLE_TOTAL_GPUS: {
-      pack_tunable<int32_t>(machine.total_gpu_count(), output);  // assume symmetry
-      return;
+      return Scalar(int32_t(machine.total_gpu_count()));  // assume symmetry
     }
     case LEGATE_CORE_TUNABLE_TOTAL_OMPS: {
-      pack_tunable<int32_t>(machine.total_omp_count(), output);  // assume symmetry
-      return;
-    }
-    case LEGATE_CORE_TUNABLE_NUM_PIECES: {
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/763/commits/0518ba4d5902d3ae8c08d7193faa7e2fc0e8c045

legate/core/runtime.py
```diff
@@ -1543,6 +1539,193 @@ def create_store(
...
+        launch_domain : Rect, optional
+            Launch domain of the task.
+
+        Returns
+        -------
+        ManualTask
+            A new task
+        """
+
+        from .operation import ManualTask
```
manopapad, line 1570: This import no longer needed. Similar for other imports from `.operation` in functions copied from context.py.

legate/core/runtime.py
```diff
@@ -1543,6 +1539,193 @@ def create_store(
...
+
+        if store.ndim > 1:
+            raise NotImplementedError(
+                "Tree reduction doesn't currently support "
+                "multi-dimensional stores"
+            )
+
+        result = self.create_store(store.type)
+
+        # A single Reduce operation is mapepd to a whole reduction tree
```
manopapad, line 1716: No longer need to eagerly flush the scheduling widow (presumably for the partitions to be created)?

manopapad (approved):
Couple minor comments on the python side

legate/core/runtime.py
magnatelee: Why is that unnecessary? Note that the import for `ManualTask` at the beginning of this file is only for type checking.

legate/core/runtime.py
magnatelee: yeah I don't think that's necessary anymore.

legate/core/runtime.py
magnatelee: fyi, if I remove the function local imports, I see name errors.

manopapad: OK, I thought those imports could be moved to the top level, but that's not possible because it would cause a circular import.

PR #772, opened 2023-06-23
avoid undefined behavior in `Span::end`

ericniebler:
The expression `&data[size_]` is forming a reference to the last+1 element of the range. That memory address may not be dereferencable.

Rather than forming a reference and taking the address, we can obtain the same result with some simple pointer arithmetic.


magnatelee (approved):

magnatelee:
@ericniebler the two CI failures are orthogonal to this PR, so feel free to merge this.

PR #777, opened 2023-06-28
Add a small type cache for array-typed scalar arguments

magnatelee:
This PR adds a small type cache for array-typed scalar arguments so their types are deduplicated. Creating a fresh array type whenever an array-typed scalar argument is passed to a task turns out to be problematic for any attempt to cache task metadata.

### DIFF
--- a/legate/core/runtime.py
+++ b/legate/core/runtime.py
@@ -1131,6 +1131,8 @@ class Runtime:
             ProcessorKind.CPU: self.core_library.LEGATE_CPU_VARIANT,
         }
 
+        self._array_type_cache: dict[tuple[ty.Dtype, int], ty.Dtype] = {}
+
     @property
     def has_cpu_communicator(self) -> bool:
         return self._comm_manager.has_cpu_communicator
@@ -2085,6 +2087,17 @@ class Runtime:
                 return result
 
         return wrapper
+
+    def find_or_create_array_type(
+        self, elem_type: ty.Dtype, n: int
+    ) -> ty.Dtype:
+        key = (elem_type, n)
+        cached = self._array_type_cache.get(key)
+        if cached is not None:
+            return cached
+        arr_type = ty.array_type(elem_type, n)
+        self._array_type_cache[key] = arr_type
+        return arr_type
 
 
 runtime: Runtime = Runtime(core_library)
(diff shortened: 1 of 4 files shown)


magnatelee:
CC @rohany

legate/core/runtime.py
```diff
@@ -2086,6 +2088,17 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    def find_or_create_array_type(
+        self, elem_type: ty.Dtype, n: int
+    ) -> ty.Dtype:
+        key = (elem_type, n)
+        cached = self._array_type_cache.get(key)
+        if cached is not None:
+            return cached
+        arr_type = ty.array_type(elem_type, n)
+        self._array_type_cache[key] = arr_type
+        return arr_type
```
bryevdv, line 2100: This is another common way to express this kind of thing:
```suggestion
        key = (elem_type, n)
        if key not in self._array_type_cache:
            arr_type = ty.array_type(elem_type, n)
            self._array_type_cache[key] = arr_type       
        return self._array_type_cache[key]
```

bryevdv (approved):
LGTM just one suggestion, assuming this will eventually be subsumed in a cpp runtime (otherwise docstring and units tests would be advised)

legate/core/runtime.py
magnatelee: I remember @jjwilke, you, and I talked about this pattern when we were on #457 and the pattern you suggested seems slightly less efficient than my original code because it tries to look up `key` twice from the dict. Is that not the case? I actually like your suggestion and that's what I believe I've been writing, but since our discussion, I've been kind of converted, though I honestly don't care that much about this micro-optimization.

magnatelee:
> assuming this will eventually be subsumed in a cpp runtime (otherwise docstring and units tests would be advised)

Yeah I guess this could be a user facing API, though it's not meant to be consumed by clients directly. And yes, this will certainly be ported over to the cpp runtime.

legate/core/runtime.py
bryevdv: It's definitely one more lookup, but its also less variable assignment and use, which also incur lookups into the namespace, more or less. I'd just be surprised if it didn't wash out (or wasn't acceptable even if it didn't quite) FWIW what I would typically write myself is actually
```
        key = (elem_type, n)
        if key not in self._array_type_cache:
            self._array_type_cache[key] = ty.array_type(elem_type, n)       
        return self._array_type_cache[key]
```

PR #778, opened 2023-06-29
Bump the Legion commit for a fix for multi-node runs

magnatelee:


scripts/generate-conda-envs.py
```diff
@@ -109,6 +109,7 @@ def conda(self) -> Reqs:
             "make",
             "rust",
             "ninja",
+            "pkg-config",
```
manopapad, line 112: I suggest we explicitly add `openssl` here, even though it's pulled by other packages.

install.py
```diff
@@ -382,6 +382,9 @@ def validate_path(path):
     pip_install_cmd = [sys.executable, "-m", "pip", "install"]
     cmd_env = dict(os.environ.items())
 
+    if "OPENSSL_DIR" not in cmd_env and "CONDA_PREFIX" in cmd_env:
+        cmd_env.update({"OPENSSL_DIR": cmd_env["CONDA_PREFIX"]})
```
manopapad, line 386: This is fine as a workaround, but IMHO the best place for this is in `get_legion.cmake`, so it will get picked up even if the user goes directly through cmake instead of using `install.py`. I've asked @trxcllnt what the best way is to locate the OpenSSL headers in cmake, while giving preference to the installation under `$CONDA_PREFIX`.

scripts/generate-conda-envs.py
magnatelee: https://github.com/nv-legate/legate.core/pull/778/commits/a1e81d4aa1beda94255cea4f78c67814dc39bf90

install.py
magnatelee: I agree, and that's why the commit message was saying this is a workaround. But some of the folks are out of the office now and I don't think the proper fix to cmake can  be done anytime soon, as it should also work without conda, so let's make a judgment call to merge this now to unblock the users that are currently held up.

manopapad: Yes absolutely

PR #784, opened 2023-07-04
More error checking for stores

magnatelee:
This PR makes two changes to error checking for `legate::Store`

1) Type validation for reduction accessors (was missing in #745)
2) Exceptions instead of `LEGATE_ABORT`

### DIFF
--- a/src/core/data/store.h
+++ b/src/core/data/store.h
@@ -267,6 +267,7 @@ class FutureWrapper {
 
 /**
  * @ingroup data
+ *
  * @brief A multi-dimensional data container storing task data
  */
 class Store {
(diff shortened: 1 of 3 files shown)


src/core/data/store.h
```diff
@@ -392,6 +420,12 @@ class Store {
    * @brief Returns a write-only accessor to the store for the entire domain. Validates type and
    * dimension by default, disable with VALIDATE_TYPE = false.
```
manopapad, line 421:
```suggestion
   * @brief Returns a write-only accessor to the store for the entire domain.
```

src/core/data/store.h
```diff
@@ -404,6 +438,12 @@ class Store {
    * @brief Returns a read-write accessor to the store for the entire domain. Validates type and
    * dimension by default, disable with VALIDATE_TYPE = false.
```
manopapad, line 439:
```suggestion
   * @brief Returns a read-write accessor to the store for the entire domain.
```

src/core/data/store.h
```diff
@@ -425,9 +465,13 @@ class Store {
    * @tparam EXCLUSIVE Indicates whether reductions can be performed in exclusive mode. If
    * `EXCLUSIVE` is `false`, every reduction via the acecssor is performed atomically.
```
manopapad, line 466:
```suggestion
   * @tparam EXCLUSIVE Indicates whether reductions can be performed in exclusive mode. If
   * `EXCLUSIVE` is `false`, every reduction via the accessor is performed atomically.
```

manopapad (approved):
Some typos in docs, otherwise looks good

PR #785, opened 2023-07-04
Use stable UIDs for common fixed-size array types

magnatelee:
#777 got me thinking that fixed UIDs for fixed-size array types would be a better fix for the original problem, so this PR attempts to fix UIDs for "common" array types, i.e., an array type whose element type is a primitive type and whose size is smaller than 256.


rohany:
I think that this makes sense but removing the cache all-together doesn't sound right. I don't have a use case right now, but I imagine that for non-primitive types the type IDs should still be cached for the same reasons we wanted a cache in the first place.

magnatelee:
@bryevdv I appreciate if you could take a quick look at changes to the Python code!

magnatelee:
> I think that this makes sense but removing the cache all-together doesn't sound right. I don't have a use case right now, but I imagine that for non-primitive types the type IDs should still be cached for the same reasons we wanted a cache in the first place.

@rohany I disagree. if the argument type needs to be cached , the client library can do that by itself instead of relying on the core's cache, which I now think wasn't a great idea in the first place. also note that the whole reason the type cache was needed was that `(some_type,)` has been a syntactic sugar for a fixed-size array type of `some_type`, which I feel is now obsolete thanks to the core type system. I think we'll keep this syntactic sugar for now, but if the type needs to be cached, you could just un-sugar it. Also note that non-primitive types aren't currently supported and will be only after the python binding for `legate::Scalar` is done.

legate/core/operation.py
```diff
@@ -399,13 +397,39 @@ def add_scalar_arg(
...
-        self._scalar_args.append((value, dtype))
+        sanitized: ty.Dtype
+        if isinstance(dtype, ty.Dtype):
+            sanitized = dtype
+            if isinstance(sanitized, ty.FixedArrayDtype) and not isinstance(
+                value, (tuple, Shape)
+            ):
+                raise ValueError(f"{value} is not a valid scalar")
+        elif isinstance(dtype, tuple):
+            assert len(dtype) == 1 and isinstance(dtype[0], ty.Dtype)
```
bryevdv, line 413: Is `add_scalar_arg` a function that users (i.e. devs building tools on `legate.core` in other projects) will use directly? If so, I'd consider `dtype` as a user-supplied value, and therefore not appropriate for an assertion since it is strictly outside out internal control (could raise a `ValueError` instead)

legate/core/operation.py
```diff
@@ -399,13 +397,39 @@ def add_scalar_arg(
...
-        if not isinstance(dtype, (ty.Dtype, tuple)):
-            raise ValueError(f"Unsupported type: {dtype}")
-        self._scalar_args.append((value, dtype))
+        sanitized: ty.Dtype
+        if isinstance(dtype, ty.Dtype):
+            sanitized = dtype
+            if isinstance(sanitized, ty.FixedArrayDtype) and not isinstance(
+                value, (tuple, Shape)
+            ):
+                raise ValueError(f"{value} is not a valid scalar")
```
manopapad, line 411: Should we also check the length of `value` against `sanitized. num_elements()` at this point?

src/core/type/type_info.cc
```diff
@@ -181,8 +181,17 @@ std::unique_ptr<Type> string_type() { return std::make_unique<StringType>(); }
...
                                        uint32_t N) noexcept(false)
 {
-  return std::make_unique<FixedArrayType>(
-    Runtime::get_runtime()->get_type_uid(), std::move(element_type), N);
+  // We use UIDs of the following format for "common" fixed array types
+  //    1B            1B
+  // +--------+-------------------+
+  // | length | element type code |
+  // +--------+-------------------+
+  auto generate_uid = [](const Type& elem_type, uint32_t N) {
```
manopapad, line 189: Is there a need for a lambda here?

manopapad (approved):
Minor suggestions

legate/core/operation.py
```diff
@@ -399,13 +397,39 @@ def add_scalar_arg(
...
+            or (
+                isinstance(sanitized, ty.FixedArrayDtype)
+                and sanitized.element_type.is_primitive
+            )
+        ):
+            raise NotImplementedError(
+                f"Scalar of type {dtype} is not yet supported"
+            )
+
+        self._scalar_args.append((value, sanitized))
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/3c30a9c9421863ef1248d522434a87a1462efc45

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+    @pytest.mark.parametrize("elem_type", _PRIMITIVES)
+    @pytest.mark.parametrize("size", [1, 10, 100, 255])
+    def test_uid(self, elem_type: ty.Dtype, size: int) -> None:
+        arr_type = ty.array_type(elem_type, size)
+        assert arr_type.uid & 0x00FF == elem_type.code
+        assert arr_type.uid >> 8 == size
+
+        for primitive_type in _PRIMITIVES:
+            assert arr_type.uid != primitive_type.uid
+            assert arr_type.uid != ty.string.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/3c30a9c9421863ef1248d522434a87a1462efc45

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+class TestFixedArrayType:
+    @pytest.mark.parametrize("elem_type", _PRIMITIVES)
+    @pytest.mark.parametrize("size", [1, 10, 100, 255])
+    def test_uid(self, elem_type: ty.Dtype, size: int) -> None:
+        arr_type = ty.array_type(elem_type, size)
+        assert arr_type.uid & 0x00FF == elem_type.code
+        assert arr_type.uid >> 8 == size
+
+        for primitive_type in _PRIMITIVES:
+            assert arr_type.uid != primitive_type.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/3c30a9c9421863ef1248d522434a87a1462efc45

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_same_type(self, elem_type: ty.Dtype) -> None:
+        type1 = ty.array_type(elem_type, 1)
+        type2 = ty.array_type(elem_type, 1)
+
+        assert type1.uid == type2.uid
+
+        type1 = ty.array_type(elem_type, 1)
+        type2 = ty.array_type(elem_type, 2)
+
+        assert type1.uid != type2.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/f547e698d060c38c75bae043617362d2f600166b

legate/core/operation.py
```diff
@@ -410,21 +410,25 @@ def add_scalar_arg(
...
-                isinstance(sanitized, ty.FixedArrayDtype)
-                and sanitized.element_type.is_primitive
+        def is_supported_scalar_arg_type(dtype: ty.Dtype) -> bool:
+            return (
+                dtype.is_primitive
+                or dtype == ty.string
+                or (
+                    isinstance(dtype, ty.FixedArrayDtype)
+                    and dtype.element_type.is_primitive
+                )
```
bryevdv, line 428: Suggest pulling this outside to the module level, so that there isn't the cost of defining this function added to every call to `add_scalar_arg`. As a side bonus the formatting will probably also improve with the extra room

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @pytest.mark.parametrize("elem_type", _PRIMITIVES)
+    def test_big(self, elem_type: ty.Dtype) -> None:
+        arr_type = ty.array_type(elem_type, 256)
+
+        assert arr_type.uid >= 0x10000
+
+        for primitive_type in _PRIMITIVES:
+            assert arr_type.uid != primitive_type.uid
+            assert arr_type.uid != ty.string.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/commit/3c30a9c9421863ef1248d522434a87a1462efc45

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+
+        assert type1.uid != type2.uid
+        assert type1.uid >= 0x10000
+        assert type2.uid >= 0x10000
+
+        for primitive_type in _PRIMITIVES:
+            assert type1.uid != primitive_type.uid
+            assert type2.uid != primitive_type.uid
+            assert type1.uid != ty.string.uid
+            assert type2.uid != ty.string.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/commit/3c30a9c9421863ef1248d522434a87a1462efc45

tests/unit/legate/core/test_type_uid.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_isomorphic_non_primitive(self, elem_type: ty.Dtype) -> None:
+        type1 = ty.array_type(ty.array_type(elem_type, 1), 1)
+        type2 = ty.array_type(ty.array_type(elem_type, 1), 1)
+
+        assert type1.uid != type2.uid
+
+        type1 = ty.array_type(ty.struct_type([elem_type]), 1)
+        type2 = ty.array_type(ty.struct_type([elem_type]), 1)
+
+        assert type1.uid != type2.uid
```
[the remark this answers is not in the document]
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/c369e1b0d9ef4d8dfd11c36d137961be4b60d0f6

legate/core/operation.py
magnatelee: yeah I initially had a `TypeError` in that place, but mypy seems to be able to catch the issue, so I was just hoping users would use mypy (plus the docstring already states expect types). but yeah I agree that should be an exception, though I'm a little sad that checks like this are still necessary despite mypy and all type annotations we put in.

legate/core/operation.py
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/119261a765d2ae3585b4053fa5ee12c9ef6c3f28

legate/core/operation.py
magnatelee: https://github.com/nv-legate/legate.core/pull/785/commits/119261a765d2ae3585b4053fa5ee12c9ef6c3f28

legate/core/operation.py
bryevdv: mypy can help make sure our usage of an API (in other library code or examples, etc.) stays correct, but at the end of the day, outside users can pass in anything they please, so explicit error checking of inputs will always be relevant on public API codepaths.

src/core/type/type_info.cc
magnatelee: It's just a stylistic choice. I've been using a lambda whenever the logic in a scope gets complicated enough to warrant a descriptive name.

bryevdv (approved):

src/core/type/type_info.cc
magnatelee: also, lambdas like this will be just compiled away, so no runtime overhead would exist.

PR #788, opened 2023-07-07
GH, docker based (reusable) workflows

sandeepd-nv:


continuous_integration/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+ENV HISTFILE=/home/coder/.cache/._bash_history
+
+ARG REPO_OWNER
+ENV REPO_OWNER=${REPO_OWNER}
+
+ARG REPO_NAME
+ENV REPO_NAME=${REPO_NAME}
+
+ARG GITHUB_TOKEN
+ENV GITHUB_TOKEN=${GITHUB_TOKEN}
```
trxcllnt, line 23: This embeds the GITHUB_TOKEN in the history of the docker image. Please mount secrets into the image instead of define them as build args.

.github/workflows/gh-build.yml
```diff
@@ -67,27 +43,47 @@ jobs:
...
+      - name: Build docker image
+        run: |
+          echo BUILD_TARGET: ${{inputs.build-target}}
+          echo USE_CUDA: ${{env.USE_CUDA}}
+
+          docker build \
+            --build-arg AWS_SESSION_TOKEN=${{env.AWS_SESSION_TOKEN}} \
+            --build-arg AWS_ACCESS_KEY_ID=${{env.AWS_ACCESS_KEY_ID}} \
+            --build-arg AWS_SECRET_ACCESS_KEY=${{env.AWS_SECRET_ACCESS_KEY}} \
+            --build-arg GITHUB_TOKEN=${{env.GITHUB_TOKEN}} \
```
trxcllnt, line 55: This embeds these vars in the history of the docker image. Please mount secrets into the image instead of define them as build args.

continuous_integration/home/coder/.local/bin/activate-conda-env
trxcllnt, line 1: This file isn't necessary if there's a `DEFAULT_CONDA_ENV=legate` envvar.

trxcllnt (changes requested):
1. The PR injects secrets as build args, which implicitly embeds them in the final image history. Please mount them as docker secrets instead.
2. Pulling large image layers from `ghcr.io` is very slow unless on a network internal to GitHub.
3. I've fixed a bunch of things in PR #725, and also have upstream fixes in the 23.08 base images. We should coordinate on merging those first, because they should simplify much of this PR.

continuous_integration/Dockerfile
sandeepd-nv: Moved this to the build stage and since the final image is created from stage0 this will no longer be visible in the image history: https://github.com/nv-legate/legate.core/actions/runs/5553609192/jobs/10142367105

.github/workflows/gh-build.yml
sandeepd-nv: Since we are using a multistage docker build and the final image is created from stage0 (which does not define or use these variables) and since these variables are only used in the "build" stage (which is discarded and does not go into the final image) these will not be visible in the image history: https://github.com/nv-legate/legate.core/actions/runs/5553609192/jobs/10142367105

continuous_integration/home/coder/.local/bin/activate-conda-env
sandeepd-nv: 1. I need to create a new legate environment in run-all-tests.sh and don't want a premature activation of the legate environment (if that is even possible). This is why I need explicit control over environment activation. 

2. I am also not keen on running the test script through entrypoint because the test does not make use of any facility provided by the system of scripts embedded in the base image.

.github/workflows/gh-build.yml
m3vaz: There is also an open PR to update the version of docker in the base runner image (https://github.com/actions/runner/pull/2694)

m3vaz: Marking resolved as this would only show up on the intermediate image.

m3vaz (changes requested):
Spacing and capitalization nits. The naming of the scripts and the functions in the scripts *will* cause confusion further down the line if there are repeats and if the only difference is `-` vs `_`
Consider creating a `conda-management-utils` file to host all your utility functions and use arguments on your functions
edit: also indentation nits, we should be consistent in the two space indent

m3vaz:
@sandeepd-nv what would be the retention period of images on ghcr.io?

continuous_integration/home/coder/.local/bin/activate-conda-env
m3vaz: I would agree with @sandeepd-nv here, if we can have finer grain control over how we're managing the environment, that would be ideal. We can fall back on the silent default if we're not manipulating the environment itself, but if we are interacting with the environment we should try to be explicit.

m3vaz: Marking resolved.

m3vaz (approved):

m3vaz:
LGTM, please submit when able.

PR #791, opened 2023-07-11
Pin host compilers to 11.* during environment generation

m3vaz:


m3vaz:
Focusing on dealing with the compilers pin to start with. Will try to and address the environment resolution issues in a later PR.

PR #797, opened 2023-07-20
Bump up the version of legate.core

marcinz:
- Update version of legate.core in `CMakeLists.txt`
- Bump up the version of Legion
- ~~Bump up the version of Thrust~~

### DIFF
--- a/cmake/versions.json
+++ b/cmake/versions.json
@@ -7,7 +7,7 @@
     },
     "Legion": {
       "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "70f9fbbce07ec4772696c57036d5fc7f84ea264d"
+      "git_tag" : "a2ec81dde867e542b335ea98826475f8d601c2ad"
     }
   }
 }
--- a/cmake/versions.json
+++ b/cmake/versions.json
@@ -7,7 +7,7 @@
     },
     "Legion": {
       "git_url" : "https://gitlab.com/StanfordLegion/legion.git",
-      "git_tag" : "14790b975da2bf559c2035b405286995a41aee69"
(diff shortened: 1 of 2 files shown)


cmake/versions.json
```diff
@@ -1,13 +1,13 @@
 {
   "packages" : {
     "Thrust" : {
-      "version" : "1.17.0.0",
+      "version" : "2.1.0.0",
```
magnatelee, line 4: Please make sure cuNumeric builds correctly with this version of Thrust. Also, I believe this version of Thrust is using the new `nv/target` headers that aren't available in CTK <= 11.5, which means this would effectively bump up the minimum CTK requirement. So I think at the very minimum the team members should be aware of this change. (I don't recall we discussed this in any of our meetings, maybe it was brought up in a devops sync-up?)

marcinz: No, we did not discuss this. Let me undo this change then, and we can discuss it on Tuesday.

magnatelee (approved):

magnatelee:
@marcinz I believe there's a Legion fix that we want to be included: https://gitlab.com/StanfordLegion/legion/-/commit/a72251082b7c21f674f830cfcd5a44d25107eb20 so I suggest we hold off on merging this, as we will want to bump the commit hash again. (cc @manopapad )

marcinz:
OK, I'll wait for this fix to be merged.

manopapad:
The fix has been merged up to legion control_replication top-of-tree, you can bump the legion commit hash to 14790b975da2bf559c2035b405286995a41aee69

marcinz:
@manopapad That commit seems to be failing. I'll try the latest.

marcinz:
I will split the legion update and version update into separate PRs. This way, we can commit version update without waiting for Legion to pass the tests.

PR #801, opened 2023-07-25
GPU test run visualization script

manopapad:


scripts/viz-gpu-test-log.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def write_json(self, out: TextIO) -> None:
+        # First sort tests by start time, then duration.
+        self._tests.sort(key=lambda test: (test.ts, test.dur))
+        for i, test in enumerate(self._tests):
+            test.tid = i
+        out.write("[\n")
+        for test in self._tests:
+            out.write(test.to_json())
+        out.write("]\n")
```
bryevdv, line 140: I would rewrite all this to start from data and use `json.dumps`, and not modify the original data i.e. 
```python
tests = sorted(self._tests, key=lambda test: (test.ts, test.dur))
for i, test in enumerate(tests):
    test.tid = i
json.dumps([test.to_dict() for test in tests])
```
nice to have would be to add and pass on `**kwargs` to `json.dumps` to expose formatting options

manopapad:
Continued in a follow-on PR

PR #802, opened 2023-07-26
Update Thrust to the latest version

marcinz:
This update will have impact on the CTK requirement. There is be a corresponding cunumeric PR to test the impact of the update on cunumeric: nv-legate/cunumeric#1006

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ endif()
 # - Download and initialize RAPIDS CMake helpers -----------------------------
 
 if(NOT EXISTS ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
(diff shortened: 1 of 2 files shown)


manopapad:
I think we can stick to thrust 1.17 for now, and avoid having to bump our minimum CTK version. It appears that thrust is moving towards a model where thrust, cub and libcu++ are distributed together, and don't depend on specific CTK versions.  Once https://github.com/nvidia/cccl and the corresponding conda package https://anaconda.org/conda-forge/cuda-cccl-impl are mature we can switch to those.

marcinz:
Rapids cmake has been updated for the new Thrust: https://github.com/rapidsai/rapids-cmake/pull/399 

They apply quite a few patches and install their own libcudacxx. I will check if that works with the older CTK.

PR #805, opened 2023-07-28
GH build releasables

sandeepd-nv:


.github/workflows/ci-gh-cpu-build-and-test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     with:
       build-target: cpu
       repos-name: ${{ github.event.repository.name }}
-      runs-on: ${{ contains(github.repository, 'nv-legate/legate.core') && 'linux-amd64-cpu4' || 'ubuntu-latest' }}
+      runs-on: ${{ contains(github.repository, 'nv-legate/legate.core') && 'linux-amd64-32cpu' || 'ubuntu-latest' }}
```
marcinz, line 41: Is the name of the runner misspelled? Should it be cpu32?

.github/workflows/gh-build.yml
```diff
@@ -54,29 +61,26 @@ jobs:
 
       - name: Build docker image
         run: |
-          echo BUILD_TARGET: ${{ inputs.build-target }}
-          echo USE_CUDA: ${{ env.USE_CUDA }}
-
           IMAGE_TAG=${{ env.IMAGE_NAME }}:${{ inputs.sha }}
 
+          chmod +x continuous_integration/build-docker-image
```
marcinz, line 66: Should the mode of the file be changed in the repo instead of here?

continuous_integration/home/coder/.local/bin/build-legate
```diff
@@ -0,0 +1,35 @@
...
+    build-legate-cpp;
+    build-legate-wheel;
+    build-legate-conda;
+}
+
+build_legate_release() {
+    . conda-utils
+    activate_conda_env
+    mkdir -p /tmp/env_yaml /tmp/conda-build
+    conda env export > /tmp/env_yaml/pre_build.yml
```
marcinz, line 14: What is this for? Same for the next one.

.github/workflows/ci-gh-cpu-build-and-test.yml
sandeepd-nv: No, the name is correct. CPU 4 is the 4th CPU while a 32 CPU machine has 32 CPUs. The linux-amd64-cpu4 label is incorrect.

.github/workflows/gh-build.yml
sandeepd-nv: Yes.

continuous_integration/home/coder/.local/bin/build-legate
sandeepd-nv: Good catch, we don't need this.

marcinz (approved):

marcinz:
conda mambabuild --output-folder /tmp/conda-build -c nvidia -c conda-forge --no-include-recipe conda/conda-build

cuda-version seems to be mismatched with actual cuda installed. That's really the only issue I see. I am able to build the package from this PR on my machine.

conda/conda-build/build.sh
```diff
@@ -1,5 +1,40 @@
...
+        echo "Symlink /usr/bin/gcc created."
+        return 0
+    else
+        echo "GCC binary not found at $GCC_BINARY. Please make sure it exists."
+        return 1
+    fi
+}
+
+check_and_create_gcc_symlink
+
```
marcinz, line 37: So, according to @trxcllnt, this bug in `sccache` should not be there in the latest version. Perhaps we just need to update the dev container we are running on. Maybe we could go back then to the standard compiler handling in the recipe.

marcinz (changes requested):

marcinz (approved):

PR #809, opened 2023-08-04
Pre-seeding random nuber generators

ipdemes:
We want to seed `np.random` and `random` to avoid control-replication violations when running multi-rank.
We also want to warn users who tries to call `seed` inside of their code (python side)

### DIFF
--- a/legate/core/__init__.py
+++ b/legate/core/__init__.py
@@ -131,8 +131,8 @@ _np.random.seed(1234)
 _random.seed(1234)
 
 
-def _warn_seed(func):
-    def wrapper(*args, **kw):
+def _warn_seed(func: AnyCallable)->AnyCallable:
+    def wrapper(*args:Any, **kw:Any)->Any:
         print(
             "WARNING: Seeding the random number generator inside the legate-based code can lead to undefined behavior and/or errors when the program is executed with multiple ranks."
         )
--- a/legate/core/__init__.py
+++ b/legate/core/__init__.py
@@ -124,11 +124,11 @@ from .types import (
 )
 from .io import CustomSplit, TiledSplit, ingest
 
+import warnings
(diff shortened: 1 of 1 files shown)


bryevdv (approved):
LGTM was it possible to do local testing to confirm this satisfies the need?

legate/core/__init__.py
```diff
@@ -123,3 +123,25 @@
...
+from typing import Any
+from .runtime import AnyCallable
+
+_np.random.seed(1234)
+_random.seed(1234)
+
+
+def _warn_seed(func: AnyCallable) -> AnyCallable:
+    def wrapper(*args: Any, **kw: Any) -> Any:
+        print(
```
manopapad, line 138: Can we use `warnings.warn` instead, so this message is not reported multiple times?

legate/core/__init__.py
```diff
@@ -123,3 +123,25 @@
...
+            "WARNING: Seeding the random number generator inside the legate-based code can lead to undefined behavior and/or errors when the program is executed with multiple ranks."
```
manopapad, line 139:
```suggestion
            "Seeding the random number generator with a non-constant value "
            "inside Legate can lead to undefined behavior and/or errors when "
            "the program is executed with multiple ranks."
```

legate/core/__init__.py
manopapad: I was thinking actually to use the python `warnings` module https://docs.python.org/3/library/warnings.html#warnings.warn, which makes sure to deduplicate messages. AFAICT what you're doing now just introduces formatting, and doesn't take advantage of the rest of the functionality of the `warnings` module.

ipdemes: done

manopapad (approved):

PR #813, opened 2023-08-09
Add support for custom wrappers

bryevdv:
This PR adds two new command line options `--wrapper` and `--wrapper-inner` to allow users to insert their own custom wrapper executables int the legate invocation chain. 

Both wrappers come after `bind.sh` so that `%%LEGATE_GLOBAL_RANK%%` can be expanded in the invocation.

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -148,7 +148,7 @@ profiling.add_argument(
     dest="cprofile",
     action="store_true",
     required=False,
-    help="profile Python execution with the cprofile module, "
+    help="profile Python execution with the cprofile module "
     "[legate-only, not supported with standard Python invocation]",
 )
 
@@ -158,7 +158,7 @@ profiling.add_argument(
     dest="nvprof",
     action="store_true",
     required=False,
-    help="run Legate with nvprof, "
+    help="run Legate with nvprof "
     "[legate-only, not supported with standard Python invocation]",
 )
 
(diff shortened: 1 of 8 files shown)


legate/driver/args.py
```diff
@@ -356,20 +356,48 @@
...
 
+other.add_argument(
+    "--wrapper",
+    dest="wrapper",
+    required=False,
+    action="append",
+    default=[],
+    help="Specify another executable (and any command-line arguments for that "
+    "executable) that will be responsible to indirectly invoke legate, "
+    "including any other wrappers. May contain the special string "
```
manopapad, line 374:
```suggestion
    "executable) to wrap the Legate executable invocation. This wrapper will "
    "come right after the launcher invocation, and will be passed the rest of "
    "the Legate invocation (including any other wrappers) to execute. May "
    "contain the special string "
```

suggested rewording, similar for `--wrapper-inner`

manopapad (approved):
Looks good. Optionally suggest rewording the help text a bit to make it clearer what this is expected to do.

legate/driver/args.py
bryevdv: https://github.com/nv-legate/legate.core/pull/813/commits/dc345d6ed0a428c5e786b84efa836b52a87f92c7

manopapad (approved):

PR #818, opened 2023-08-11
Enable OpenMP in GH CI.

sandeepd-nv:


marcinz (commented):
Is OpenMP actually used in any tests?

sandeepd-nv:
> Is OpenMP actually used in any tests?

Yes, cunumeric CI has OpenMP tests. We need to build legate.core with OpenMP support for that to work.

sandeep-datta:
> I meant, is OpenMP used in any tests as in this PR? I can see that it is enabled, but is it just for build? I don't see where it is used (the corresponding options).

Yes, it's just for build.

m3vaz (approved):
+1 Thank You

Issue #810, opened 2023-08-07
Fix per_node_count_ in sgarding function


ipdemes:
Here is the description from @lightsighter  on this issue:

the bugs that look like this:
```
legion_python: /home/idemeshko/development/legate.core_2309/src/core/runtime/shard.cc:126: virtual Legion::ShardID legate::LegateShardingFunctor::shard(const DomainPoint&, const Domain&, size_t): Assertion `start_node_id_ <= shard_id && shard_id < end_node_id_' failed.
```
look like a legate issue, I see them when running the test_ingest.py test.  In that case the problem is that we're doing an index launch of 15 points (0,0)-(4,2) over two shards, but sharding functor has this preset `per_node_count_ `member variable which i think is saying how many points get assigned to each shard, and that is set to 4, but with 15 points, then some of those points get sharded to a non-extistent shard (hence the assertion), i feel like legate either needs to register a different sharding functor for those cases, or it needs to compute `per_node_count_` based on the size of the launch bounds:
https://github.com/nv-legate/legate.core/blob/branch-23.09/src/core/runtime/shard.cc#L111
the result of calling `linearize` on this line with point (3,0), returns 9 which then divides by 4 to get 2, and that is bigger than your choice of shard IDs which is either 0 or 1:
https://github.com/nv-legate/legate.core/blob/branch-23.09/src/core/runtime/shard.cc#L124
it's unclear to me where `per_node_count_` is coming from since it gets passed in from python when the LegateShardingFunctor was registered
note we weren't seeing this problem before because legion wasn't invoking the sharding functors in some cases where it should have, so this is just a latent bug that's being uncovered by legion calling the sharding functors more often

manopapad:
I believe this is a general issue with our sharding functors, that they don’t support overdecomposition, which the user can cause by doing a manual task launch. I can trigger it by changing cunumeric like so:

```
diff --git a/cunumeric/deferred.py b/cunumeric/deferred.py
index 5cbea74b..8d97849d 100644
--- a/cunumeric/deferred.py
+++ b/cunumeric/deferred.py
@@ -35,7 +35,7 @@ from typing import (

 import legate.core.types as ty
 import numpy as np
-from legate.core import Annotation, Future, ReductionOp, Store
+from legate.core import Annotation, Future, Rect, ReductionOp, Store
 from numpy.core.numeric import (  # type: ignore [attr-defined]
     normalize_axis_tuple,
 )
[...]

PR #819, opened 2023-08-11
Adding support for oversubscription

ipdemes:
This PR is fixing https://github.com/nv-legate/legate.core/issues/810


src/core/mapping/base_mapper.cc
```diff
@@ -217,11 +217,13 @@ void BaseMapper::slice_task(const Legion::Mapping::MapperContext ctx,
   if (task.sharding_space.exists())
     sharding_domain = runtime->get_index_space_domain(ctx, task.sharding_space);
 
-  auto lo = key_functor->project_point(sharding_domain.lo(), sharding_domain);
-  auto hi = key_functor->project_point(sharding_domain.hi(), sharding_domain);
+  auto lo                 = key_functor->project_point(sharding_domain.lo(), sharding_domain);
+  auto hi                 = key_functor->project_point(sharding_domain.hi(), sharding_domain);
+  int32_t shards_per_proc = (sharding_domain.get_volume() + local_range.total_proc_count() - 1) /
+                            local_range.total_proc_count();
```
magnatelee, line 223: I don't think the `key_functor` always preserves the volume of the sharding domain. it'd be safe to calculate the number of tasks using `lo` and `hi`.

magnatelee: `linearize(lo, hi, hi) + 1` would do it

magnatelee: I'd also call it `tasks_per_proc` instead of `shards_per_proc`.

src/core/mapping/machine.cc
```diff
@@ -142,9 +142,9 @@ LocalProcessorRange::LocalProcessorRange(uint32_t offset,
 
 const Processor& LocalProcessorRange::operator[](uint32_t idx) const
 {
-  auto local_idx = (idx % total_proc_count_) - offset_;
+  auto local_idx = (idx - offset_) % procs_.size();
```
magnatelee, line 145: I feel the caller would guarantee that `0 <= idx - offset_ < procs_.size()`, if so we don't need this modulo operator at all. can you confirm if this is really the case?

src/core/mapping/base_mapper.cc
ipdemes: Thank you, changed that.

src/core/mapping/machine.cc
ipdemes: We would actually still need to get modulus. ` idx-offset` can sill be` >=procs_.size()`, the changes in the base mapper just help us to make sure that neighboring indices are mapped to the same proc after we execute the modulus operation

magnatelee: if I understand your code correctly, the indices linearized from neighboring points should all be mapped to the same processor id at line 226 by the division with `tasks_per_proc`. so I still don't understand why a processor index minus the offset can still be greater than the number of local processors.

ipdemes: yes  you are right, we don't need mod operator

I also found a case when calculated idx is smaller than offset_  (so I added the check):
in cholesky example we have 25 index points distributed between 2 shards with 2 procs per shard.
Here the first 12 points are assigned to shard 0, and the remaining 13 are assigned to shard 1. 
When following the  idx calculation in the mapper, the idx for the point 13 will be `1`, when offset for this idx is `2`. So I added:
```
if (offset_ > idx) local_idx = idx;
```

magnatelee: that still doesn't sound right. I think the number of tasks per shard should have been 13 (= (25 + 2 - 1) / 2), then you wouldn't have had that case. keeping the last one getting fewer tasks would likely make the logic cleaner.

ipdemes: You were right that we should be dividing by the `tasks_per_proc` , however   I had to modify the proposed formula to deal with the offset per slice. I think the latest version does the right thing now.

ipdemes: Note that I still need to do `idx=idx%procs_.size()` in the machine.cc since the logic in the mapper assumes the case when all the tasks are distributed linearly across all shards. However, there are cases when we have all tasks assigned to a single shard (in bincount example we get 2 shards, 4 procs per shard, 5 points in sharding domain where all 5 points are assigned to the same shard)

src/core/runtime/shard.cc
```diff
@@ -121,7 +121,18 @@ class LegateShardingFunctor : public Legion::ShardingFunctor {
...
+    uint32_t tasks_per_proc = 1;
+    if (diff > 0) {
+      auto num_shards      = end_node_id_ - start_node_id_;
+      uint32_t tasks_count = (launch_space.get_volume() + num_shards - 1) / num_shards;
+      tasks_per_node       = std::max(per_node_count_, tasks_count);
+      tasks_per_proc       = (tasks_per_node + per_node_count_ - 1) / per_node_count_;
+    }
+    auto proc_id  = (linearize(lo, hi, point) + offset_) / tasks_per_proc + start_node_id_;
+    auto shard_id = proc_id / per_node_count_;
+
```
magnatelee, line 135: if you changed `LegateShardingFunctor` to be instantiated with a processor range (say `start_proc_id_` and `end_proc_id_`), instead of a pair of start and end nod ids, you could have used the same logic as `slice_task ` here; i.e.,
```
  uint32_t task_count = linearize(lo, hi, hi) + 1;
  uint32_t proc_count = end_proc_id_ - start_proc_id_ + 1;
  uint32_t idx = (linearize(lo, hi, p) * proc_count) / task_count + start_proc_id_;
  auto shard_id = global_proc_id / per_node_count_;
```

src/core/mapping/base_mapper.cc
```diff
@@ -217,11 +217,13 @@ void BaseMapper::slice_task(const Legion::Mapping::MapperContext ctx,
...
-  auto hi = key_functor->project_point(sharding_domain.hi(), sharding_domain);
+  auto lo                    = key_functor->project_point(sharding_domain.lo(), sharding_domain);
+  auto hi                    = key_functor->project_point(sharding_domain.hi(), sharding_domain);
+  uint32_t total_tasks_count = linearize(lo, hi, hi) + 1;
+
   for (Domain::DomainPointIterator itr(input.domain); itr; itr++) {
-    auto p   = key_functor->project_point(itr.p, sharding_domain);
-    auto idx = linearize(lo, hi, p);
+    auto p       = key_functor->project_point(itr.p, sharding_domain);
+    uint32_t idx = linearize(lo, hi, p) * local_range.total_proc_count() / total_tasks_count;
```
magnatelee, line 226: You need to add the starting processor id here so `idx` denotes a global processor id. I think we should modify `LocalProcessorRange` so its `offset_` be the starting processor id.

ipdemes: do you think using local processor ID here will result in incorrect logic ? We use local proc ID down this line anyway

magnatelee: `idx` per your calculation is an offset from the starting global processor id. that's why you needed a modulo operator in the `LocalProcessorRange::operator[]`. if you don't add the tarting global processor id, then you code won't handle correctly the case where a processor range start from a non-zero global processor id.

ipdemes: would it be correct to calculate starting processor id as `machine_desc.processor_range().low`?

magnatelee: yes

legate/core/machine.py
```diff
@@ -196,6 +196,15 @@ def _get_range(self, kind: ProcessorKind) -> ProcessorRange:
     def only(self, *kinds: ProcessorKind) -> Machine:
         return Machine([self._get_range(kind) for kind in kinds])
 
+    def slice(self, start_proc_id: int, end_proc_id: int) -> Machine:
```
magnatelee, line 199: why do we need this API? can't you just use an index expression?

tests/integration/scoping/examples/test_scoping.py
```diff
@@ -90,7 +90,19 @@ def test_cpu_only():
...
+
+
+def test_shifted_slices():
+    # this will test processor slicing of the machine
+    m = get_machine()
+    for i in range(len(m)):
+        for j in range(i + 1, len(m)):
+            with m.slice(i, j):
+                task = user_context.create_auto_task(
+                    user_lib.shared_object.EMPTY
```
magnatelee, line 103: this task should really check if the mapping was done correctly

src/core/runtime/shard.cc
```diff
@@ -117,21 +117,24 @@ class LegateShardingFunctor : public Legion::ShardingFunctor {
...
     return shard_id;
   }
 
  private:
   LegateProjectionFunctor* proj_functor_;
-  uint32_t start_node_id_;
-  uint32_t end_node_id_;
+  uint32_t start_proc_id_;
+  uint32_t end_proc_id_;
   uint32_t offset_;
```
magnatelee, line 138: I don't think we need `offset_` anymore.

src/core/mapping/base_mapper.cc
```diff
@@ -1029,8 +1033,13 @@ void BaseMapper::map_copy(const Legion::Mapping::MapperContext ctx,
...
     auto hi           = key_functor->project_point(sharding_domain.hi(), sharding_domain);
     auto p            = key_functor->project_point(copy.index_point, sharding_domain);
-    auto idx          = linearize(lo, hi, p);
-    target_proc       = local_range[idx];
+
+    uint32_t start_proc_id     = machine_desc.processor_range().low;
+    uint32_t total_tasks_count = linearize(lo, hi, hi) + 1;
+    auto idx =
+      linearize(lo, hi, p) * local_range.total_proc_count() / total_tasks_count + start_proc_id;
+    ;
```
magnatelee, line 1041: why?

src/core/mapping/machine.cc
```diff
@@ -263,7 +263,7 @@ LocalProcessorRange Machine::slice(TaskTarget target,
       return LocalProcessorRange();
   }
 
-  return LocalProcessorRange(slice.low - global_range.low,
+  return LocalProcessorRange(slice.low - global_range.low + machine_desc.processor_range().low,
```
magnatelee, line 266: I believe `slice.low - global_range.low + machine_desc.processor_range().low` is the same as `slice.low`.

magnatelee:
@ipdemes you should also update `map_copy` so it has the same slicing logic as `slice_task`.

src/core/runtime/shard.cc
ipdemes: yes, removing it

src/core/mapping/machine.cc
ipdemes: yes, thank you.

tests/integration/scoping/src/library.cc
```diff
@@ -43,6 +43,23 @@ void validate(legate::TaskContext& context)
   }
 }
 
+void map_check(legate::TaskContext& context)
+{
+  int32_t task_count          = context.get_launch_domain().get_volume();
+  int32_t shard_id            = context._get_shard_id();
```
magnatelee, line 49: can we use `legate::Processor::get_executing_processor().address_space()` instead? we should stop exposing Legion details to the API

magnatelee (approved):

PR #821, opened 2023-08-24
Clean up reduction store init using the new future map reduction API

magnatelee:


src/core/utilities/deserializer.cc
```diff
@@ -67,9 +65,9 @@ void TaskDeserializer::_unpack(Store& value)
...
 void TaskDeserializer::_unpack(FutureWrapper& value)
 {
-  auto read_only   = unpack<bool>();
-  auto has_storage = unpack<bool>();
-  auto field_size  = unpack<uint32_t>();
+  auto read_only    = unpack<bool>();
+  auto future_index = unpack<int32_t>();
+  auto field_size   = unpack<uint32_t>();
 
   auto point = unpack<std::vector<int64_t>>();
```
jjwilke, line 72: predates the PR, but there is redundant code here that requires the serialization format to be updated in multiple places.  Some possibilities:

```
void TaskDeserializer::_unpack(Domain& domain){
  auto point = unpack<std::vector<int64_t>>();
  domain.dim = static_cast<int32_t>(point.size());
  for (int32_t idx = 0; idx < domain.dim; ++idx) {
    domain.rect_data[idx]              = 0;
    domain.rect_data[idx + domain.dim] = point[idx] - 1;
  }  
}

static auto _unpack_future_wrapper(TaskDeserializer& ser){
  return std::make_tuple(
    /*read_only=*/unpack<bool>(),
    /*future_index=*/unpack<int32_t>(),
    /*field_size=*/unpack<uint32_t>(),
    /*domain=*/unpack<Domain>());
  }
}


auto [ready_only, future_index, field_size, domain] = _unpack_future_wrapper(*this);
...
```

jjwilke (approved):
LGTM

src/core/utilities/deserializer.cc
magnatelee: Yeah I agree there are some snippets that can be factored out as you suggested. I haven't done them yet as the format is still in flux and there's a great chance that future refactorings may render such deduplication obsolete. Let's address this in a follow-on PR once it becomes necessary.

PR #823, opened 2023-08-25
Supporting unordered detach operation

ipdemes:


legate/core/store.py
```diff
@@ -158,6 +159,8 @@ def attach_external_allocation(
         assert self.physical_region_refs == 0
         # Record the attached memory ranges, and confirm no overlaps with
         # previously encountered ranges.
+        if self.detach_future and not self.detach_future.is_ready():
+            self.detach_future.wait()
```
magnatelee, line 163: I don't think this is the right place to do this check. it should be done when the region field is recycled. and the recycling logic should be tweaked such that it favors free region fields with no pending detach over those with one.

legate/core/runtime.py
```diff
@@ -346,6 +358,31 @@ def allocate_field(self, field_size: Any) -> tuple[Region, int, bool]:
...
+    while not is_ready and count > 0:
+        field_info = free_fields.popleft()
+        if field_info[2] is not None and field_info[2].is_ready():
+            is_ready = True
+        elif field_info[2] is None:
+            is_ready = True
+        else:
+            free_fields.append(field_info)
+        count -= 1
+    # FIXME: should I return None here instead waiting for future?
```
ipdemes, line 377: @magnatelee : what would you suggest doing in the case when all fields in the free_fields container have not been deallocated yet: should we call `wait` on the `detach_future`  or return `None`?

legate/core/store.py
ipdemes: @magnatelee : I believe I have added requested logic. Do you think I should still keep the check here? Can there be the case when RegionField is re-attached without going through the FieldManager ?

magnatelee: This if-statement should be changed to an assertion or an exception, as the store at this line shouldn't have any outstanding detachment. Also, re-attaching is illegal. Look at the check at line 160.

legate/core/runtime.py
```diff
@@ -358,15 +395,15 @@ def __init__(
         # This is a sanitized list of (region,field_id) pairs that is
         # guaranteed to be ordered across all the shards even with
         # control replication
-        self.free_fields: Deque[tuple[Region, int]] = deque()
+        self.free_fields: Deque[
```
magnatelee, line 398: I feel `_try_reuse_field` would become much simpler if we had two deques, one for those free of pending detach ops and one for those with them.

legate/core/runtime.py
magnatelee: Let's wait on the future for now. Though any stores with attachments are expected to be small (otherwise, attached allocations couldn't even have been created in a single memory), we don't want to spew stores in the pathological case where the client code keeps creating a store and attaching an allocation to it.

magnatelee (approved):

magnatelee:
@ipdemes I didn't do a thorough review on this, as we know that this code won't be maintained in the future. Let's just merge it for now and move on.

legate/core/runtime.py
```diff
@@ -346,6 +358,31 @@ def allocate_field(self, field_size: Any) -> tuple[Region, int, bool]:
         return self._region, field_id, revived
 
 
+def _try_reuse_field(
```
manopapad, line 361: I suggest just blocking on the first Future, I'm worried that two shards might test the same future at slightly different points in time, one of them finds it ready and the other doesn't, and they end up using different fields.

```
def _try_reuse_field(
     free_fields: Deque[tuple[Region, int, Union[Future, None]]]
) -> Optional[tuple[Region, int]]:
    if len(free_fields) == 0:
        return None
    field_info = free_fields.popleft()
    if field_info[2] is not None:
        field_info[2].wait()
    return field_info[0], field_info[1]
```

legate/core/runtime.py
```diff
@@ -586,24 +632,26 @@ def detach_external_allocation(
...
-    ) -> None:
+    ) -> Union[None, Future]:
         # If the detachment was previously deferred, then we don't
         # need to remove the allocation from the map again.
         if not previously_deferred:
             self._remove_allocation(alloc)
         if defer:
             # If we need to defer this until later do that now
+            # IRINA FIXME should I pass unordered here?
             self._deferred_detachments.append((alloc, detach))
```
manopapad, line 643: I think you need to record the `RegionField`'s `self.field` in this scenario, so when the detachment actually happens later in `perform_detachments` you can call `Field.add_detach_future` at that point.

I would pass a `Field` argument to this method, change this line to

```
self._deferred_detachments.append((alloc, detach, field))
```

change the type of `AttachmentManager._deferred_detachments` appropriately, and change `perform_detachments` as follows:

```
    def perform_detachments(self) -> None:
        detachments = self._deferred_detachments
        self._deferred_detachments = list()
        for alloc, detach, field in detachments:
            detach_future = self.detach_external_allocation(
                alloc, detach, defer=False, previously_deferred=True
            )
            if detach_future is not None:
                field.add_detach_future(detach_future)
```

legate/core/runtime.py
```diff
@@ -428,22 +469,27 @@ def try_reuse_field(self) -> Optional[tuple[Region, int]]:
 
         # First, if we have a free field then we know everyone has one of those
         if len(self.free_fields) > 0:
-            return self.free_fields.popleft()
+            return _try_reuse_field(self.free_fields)
 
         self._field_match_manager.update_free_fields()
```
manopapad, line 474:
```suggestion
        self._field_match_manager.update_free_fields()

        # If any free fields were discovered on all shards, push their
        # unordered detachments to the task stream now, so we can safely
        # block on them later without fear of deadlock.
        if self.free_fields > 0:
            self._runtime._progress_unordered_operations()
```

legate/core/store.py
```diff
@@ -246,13 +254,17 @@ def detach_external_allocation(
...
             self.attached_alloc, detach, defer
         )
         self.physical_region = None
         self.physical_region_mapped = False
         self.physical_region_refs = 0
         self.attached_alloc = None
+        if unordered:
+            self.detach_future = detach_future
+            if self.detach_future is not None:
+                self.field.add_detach_future(self.detach_future)
```
manopapad, line 267: I think you can probably just get rid of `RegionField.detach_future` at this point, and just set `self.field` directly:

```suggestion
        if detach_future is not None:
            self.field.add_detach_future(detach_future)
```

legate/core/store.py
```diff
@@ -246,13 +254,17 @@ def detach_external_allocation(
         assert self.attached_alloc is not None
         detach = attachment_manager.remove_detachment(self.detach_key)
         detach.unordered = unordered  # type: ignore[union-attr]
-        attachment_manager.detach_external_allocation(
+        detach_future = attachment_manager.detach_external_allocation(
             self.attached_alloc, detach, defer
```
manopapad, line 258: Pass `self.field` here (see further discussion on this on another comment).

```suggestion
            self.attached_alloc, detach, defer, self.field
```

legate/core/runtime.py
ipdemes: makes sense

legate/core/runtime.py
```diff
@@ -632,15 +624,15 @@ def detach_external_allocation(
...
     ) -> Union[None, Future]:
         # If the detachment was previously deferred, then we don't
         # need to remove the allocation from the map again.
         if not previously_deferred:
             self._remove_allocation(alloc)
         if defer:
             # If we need to defer this until later do that now
-            # IRINA FIXME should I pass unordered here?
-            self._deferred_detachments.append((alloc, detach))
+            self._deferred_detachments.append((alloc, detach, dependent_field))
```
manopapad, line 635: Suggest `assert(dependent_field is not None)` here

manopapad (approved):
One minor suggestion, otherwise looks good

Issue #829, opened 2023-09-04
Launcher: detect when externally launched


manopapad:
Currently if Legate is launched externally, the user still has to specify the number of nodes/ranks to the launcher, otherwise the launcher will think this is an execution on 1 node and 1 rank (the defaults), and disable networking, leading to the spawned processes executing independently, instead of performing a joint execution:

```
(a) iblis:~/a/cunumeric> cat b.py
from legate.core import get_machine
print(f"there are {len(get_machine())} processors available")

# wrong: separate execution
(a) iblis:~/a/cunumeric> mpirun -n 2 legate --cpus 3 b.py
[0 - 7f2e3e23b000]    0.000141 {4}{threads}: reservation ('CPU proc 1d00000000000003') cannot be satisfied
[0 - 7f41dae27000]    0.000141 {4}{threads}: reservation ('CPU proc 1d00000000000003') cannot be satisfied
there are 3 processors available
there are 3 processors available

# correct: joint execution
(a) iblis:~/a/cunumeric> mpirun -n 2 legate --ranks-per-node 2 --cpus 3 b.py
*** WARNING (proc 0): Running with multiple processes per host without shared-memory communication support (PSHM).  This can significantly reduce performance.  Please re-configure GASNet using `--enable-pshm` to enable fast intra-host comms.
[0 - 7f1a6e7fb000]    0.000159 {4}{threads}: reservation ('CPU proc 1d00000000000003') cannot be satisfied
[1 - 7fdf9d24e000]    0.000159 {4}{threads}: reservation ('CPU proc 1d00010000000003') cannot be satisfied
there are 6 processors available
```

See https://github.com/nv-legate/cunumeric/issues/1039 for a motivating example.

When `--launcher` is `none`, we could check for the presence of the launcher rank envvars (`OMPI_COMM_WORLD_RANK` etc.), and if found deduce that we are in a parallel launch. I suggest also printing out on the first rank (`detected_rank_id == 0`) that we detected an external launch in this case.

We would need to change the logic for things like setting `-ll:networks none` from `ranks > 1` to `is_in_parallel_launch`, as we now have a case where `ranks` is nominally 1, but we are inside a parallel launch.

bryevdv:
> When --launcher is none, we could check for the presence of the launcher rank envvars (OMPI_COMM_WORLD_RANK etc.), and if found deduce that we are in a parallel launch. I suggest also printing out on the first rank (detected_rank_id == 0) that we detected an external launch in this case.

[...]

PR #832, opened 2023-09-06
auto-detect multi-node based on env vars

bryevdv:
fixes: #829 

This PR add support to modulate the default `nodes` and `ranks_per_node` values in the presence of common relevant environment variables. This is most useful when legate is "externally launched" but the `--help` string is updated with the corresponding information when this occurs:
```
  --nodes NODES         Number of nodes to use [default auto-detected from OMPI] (default: 4)
  --ranks-per-node RANKS_PER_NODE
                        Number of ranks (processes running copies of the program) to launch per node. [default auto-detected from OMPI] (default: 2)
```

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -53,20 +53,36 @@ def detect_multi_node_defaults() -> tuple[dict[str, Any], dict[str, Any]]:
     ranks_per_node_kw = dict(RANKS_PER_NODE.kwargs)
     where = None
 
-    if nodes_env := getenv("OMPI_COMM_WORLD_SIZE"):
+    if ranks_env := getenv("OMPI_COMM_WORLD_SIZE"):
         if ranks_per_node_env := getenv("OMPI_COMM_WORLD_LOCAL_SIZE"):
-            nodes, ranks_per_node = int(nodes_env), int(ranks_per_node_env)
+            ranks, ranks_per_node = int(ranks_env), int(ranks_per_node_env)
+            if ranks % ranks_per_node != 0:
+                raise ValueError(
+                    "Detected incompatible ranks and ranks-per-node from "
+                    "the environment"
+                )
+            nodes = ranks // ranks_per_node
             where = "OMPI"
 
-    elif nodes_env := getenv("MV2_COMM_WORLD_SIZE"):
+    elif ranks_env := getenv("MV2_COMM_WORLD_SIZE"):
         if ranks_per_node_env := getenv("MV2_COMM_WORLD_LOCAL_SIZE"):
-            nodes, ranks_per_node = int(nodes_env), int(ranks_per_node_env)
+            ranks, ranks_per_node = int(ranks_env), int(ranks_per_node_env)
+            if ranks % ranks_per_node != 0:
+                raise ValueError(
+                    "Detected incompatible ranks and ranks-per-node from "
+                    "the environment"
+                )
+            nodes = ranks // ranks_per_node
             where = "MV2"
 
     elif nodes_env := getenv("SLURM_JOB_NUM_NODES"):
         if ranks_env := getenv("SLURM_NTASKS"):
             nodes, ranks = int(nodes_env), int(ranks_env)
-            assert ranks % nodes == 0
+            if ranks % nodes != 0:
+                raise ValueError(
+                    "Detected incompatible nodes and ranks from the "
+                    "environment"
+                )
             ranks_per_node = ranks // nodes
             where = "SLURM"
 
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -17,6 +17,10 @@
 from __future__ import annotations
 
 from argparse import REMAINDER, ArgumentDefaultsHelpFormatter, ArgumentParser
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any
 
 from .. import __version__
 from ..util.args import InfoAction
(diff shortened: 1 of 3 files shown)


bryevdv:
@manopapad note that I do also see:
```
dev310 ❯ OMPI_COMM_WORLD_SIZE=4 OMPI_COMM_WORLD_LOCAL_SIZE=2 OMPI_COMM_WORLD_RANK=1 legate 
WARNING: Disabling control replication for interactive run
Error: Could not determine node-local rank
```
but I am not sure that is a real issue (I think that error is coming from Legion)

legate/driver/args.py
```diff
@@ -42,6 +46,45 @@
...
+
+    elif nodes_env := getenv("MV2_COMM_WORLD_SIZE"):
+        if ranks_per_node_env := getenv("MV2_COMM_WORLD_LOCAL_SIZE"):
+            nodes, ranks_per_node = int(nodes_env), int(ranks_per_node_env)
+            where = "MV2"
+
+    elif nodes_env := getenv("SLURM_JOB_NUM_NODES"):
+        if ranks_env := getenv("SLURM_NTASKS"):
+            nodes, ranks = int(nodes_env), int(ranks_env)
+            assert ranks % nodes == 0
```
bryevdv, line 69: `ValueError` might be better here &mdash; generally asserts should not be applied to validate user-supplied values, which is how I'd describe these.

manopapad: Agreed, best change that to `ValueError`.

legate/driver/args.py
```diff
@@ -42,6 +46,45 @@
...
+            ranks_per_node = ranks // nodes
+            where = "SLURM"
+
+    else:
+        nodes = defaults.LEGATE_NODES
+        ranks_per_node = defaults.LEGATE_RANKS_PER_NODE
+
+    nodes_kw["default"] = nodes
+    ranks_per_node_kw["default"] = ranks_per_node
```
manopapad, line 78: Since the defaults are already specified in `util.shared_args.NODES` and `util.shared_args.RANKS_PER_NODE`, you could just leave them alone if `where` is None: 
```suggestion
    if where:
        nodes_kw["default"] = nodes
        ranks_per_node_kw["default"] = ranks_per_node
```

legate/driver/args.py
```diff
@@ -42,6 +46,45 @@
...
 
 
+def detect_multi_node_defaults() -> tuple[dict[str, Any], dict[str, Any]]:
+    from os import getenv
+
+    nodes_kw = dict(NODES.kwargs)
+    ranks_per_node_kw = dict(RANKS_PER_NODE.kwargs)
+    where = None
+
+    if nodes_env := getenv("OMPI_COMM_WORLD_SIZE"):
```
manopapad, line 56: I realized as I was reading your testcase that I actually made a mistake in my original ask; `OMPI_COMM_WORLD_SIZE` and `MV2_COMM_WORLD_SIZE` don't refer to number of nodes, but total number of ranks; therefore this blob and the next should do something like:

```
    if ranks_env := getenv("OMPI_COMM_WORLD_SIZE"):
         if ranks_per_node_env := getenv("OMPI_COMM_WORLD_LOCAL_SIZE"):
            ranks, ranks_per_node = int(ranks_env), int(ranks_per_node_env)
            if ranks % ranks_per_node != 0:
                raise ValueError(
                    "Detected incompatible ranks and ranks-per-node from the environment"
                )
            nodes = ranks // ranks_per_node
            where = "OMPI"
```

manopapad (changes requested):

manopapad:
> but I am not sure that is a real issue (e.g. are other env vars necessary for "real" usage?)

That error message is coming from `bind.sh` I believe. Try also specifying `OMPI_COMM_WORLD_RANK` and `OMPI_COMM_WORLD_LOCAL_RANK`.

legate/driver/args.py
bryevdv: This occurred to me but I would prefer to keep an explicit catch-all "else" clause

legate/driver/args.py
bryevdv: https://github.com/nv-legate/legate.core/pull/832/commits/2975cc7ab7304e8f068a2707096ced9c12e4486e

legate/driver/args.py
bryevdv: https://github.com/nv-legate/legate.core/pull/832/commits/2975cc7ab7304e8f068a2707096ced9c12e4486e

legate/driver/args.py
bryevdv: https://github.com/nv-legate/legate.core/pull/832/commits/2975cc7ab7304e8f068a2707096ced9c12e4486e

legate/util/shared_args.py
```diff
@@ -57,8 +57,7 @@
         default=defaults.LEGATE_RANKS_PER_NODE,
         dest="ranks_per_node",
         help="Number of ranks (processes running copies of the program) to "
-        "launch per node. The default (1 rank per node) will typically result "
-        "in the best performance.",
+        "launch per node.",
```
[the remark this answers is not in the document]
bryevdv: https://github.com/nv-legate/legate.core/pull/832/commits/2975cc7ab7304e8f068a2707096ced9c12e4486e

bryevdv:
:+1: adding `OMPI_COMM_WORLD_LOCAL_RANK` fixed the `bind.sh` issue

manopapad (approved):

PR #834, opened 2023-09-13
Respect the type of passed storage in create_store

manopapad:


legate/core/runtime.py
```diff
@@ -1504,7 +1511,9 @@ def create_store(
             shape = Shape(shape)
 
         kind = (
-            Future
+            type(data)
+            if data is not None
+            else Future
             if optimize_scalar and shape is not None and shape.volume() == 1
             else RegionField
```
bryevdv, line 1518: TBH I'd split this up some way, chained ternaries are a bit awkward for people and for `black`

legate/core/runtime.py
```diff
@@ -1492,6 +1492,13 @@ def create_store(
     ) -> Store:
         from .store import RegionField, Storage, Store
 
+        if (
+            data is not None
+            and not isinstance(data, RegionField)
+            and not isinstance(data, Future)
```
bryevdv, line 1498:
```python
if data and not isinstance(data, (RegionField, Future))
``` 
would avoid the  `black` vertical explosion I think

bryevdv (approved):

PR #835, opened 2023-09-18
Enabling shard-local future creation for io

ipdemes:
This PR fixes false-positive control-replication violation error in legate.core.
The IO function `make_partition`  will have each shard making its own set of local futures and then creating a custom future map from that (totally legal), but the control replication checks even on level 1 will likely flag that as illegal if they make different numbers of future calls by default.
So, instead of using `legion_future_from_untyped_pointer` we want to use recently introduced  `legion_future_from_untyped_pointer_detailed` function where we can set `shard_local` to `True` for IO.
This will tell the runtime that this is a sharded future creation and to avoid the control replication checks.

I also add `shard_local`  variable to the `set_value`  method of legate.core for the future (in case other libraries built on top of legion need it)


magnatelee (approved):

legate/core/_legion/future.py
```diff
@@ -94,10 +96,17 @@ def set_value(
...
+            provenance = ""
         if self.handle is not None:
             raise RuntimeError("Future must be unset to set its value")
-        self.handle = legion.legion_future_from_untyped_pointer(
-            runtime, ffi.from_buffer(data), size
+        self.handle = legion.legion_future_from_untyped_pointer_detailed(
+            runtime,
+            ffi.from_buffer(data),
+            size,
+            False,
```
manopapad, line 107: Can we instead do

```
None if provenance is None else provenance.encode()
```

in this and other newly introduced calls, to avoid passing a "fake" provenance (instead by passing NULL we explicitly tell Legion there's no provenance)

 I don't know if it even makes a difference...

PR #842, opened 2023-09-27
Avoid gc infinite loop at runtime destruction time

manopapad:
Just wait on the futures instead of spin-looping


ipdemes:
@lightsighter what happens if we destroy future we get from unordered detachment without waiting on it?

ipdemes (approved):
This fixes wait on gc

PR #846, opened 2023-09-29
v23.09.00

marcinz:
v23.07.00


marcinz:
This must be merged with a merge commit.

manopapad:
It doesn't seem possible to allow merge commits only on specific branches, so I just disabled the requirement for pull requests on the `main` branch. You should be able to just merge `branch-23.09` into `main` and push the update without going through a PR.

PR #848, opened 2023-10-03
Satisfy new flake8 type check rules

bryevdv:
This PR resolves this class of new `flake8` errors:
> legate/core/_legion/future.py:52:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`

All `legate.core` and `cunumeric` tests were checked locally to pass. Still it's unclear if some of the instances really do intend strict type equality. If so those can be changed back with a `# noqa` added.


manopapad (approved):

manopapad:
All the `same_handle` checks are morally equivalent to `__eq__` checks, so I think it makes sense to compare type equality in those cases. If `#noqa` works to appease `flake8` in those cases, then I think that's preferable.

manopapad (approved):

PR #849, opened 2023-10-03
Expand ranges when reading thread_siblings_list

vzhurba01:
Handle both ranges and individual values

### DIFF
--- a/legate/util/system.py
+++ b/legate/util/system.py
@@ -154,8 +154,7 @@ def extract_values(line: str) -> tuple[int, ...]:
     return tuple(
         sorted(
             chain.from_iterable(
-                expand_range(r)
-                for r in line.strip().split(",")
+                expand_range(r) for r in line.strip().split(",")
             )
         )
     )
--- a/legate/util/system.py
+++ b/legate/util/system.py
@@ -99,16 +99,7 @@ class System:
                 line = open(
                     f"/sys/devices/system/cpu/cpu{i}/topology/thread_siblings_list"  # noqa E501
                 ).read()
-                sibling_sets.add(
-                    tuple(
-                        sorted(
-                            chain.from_iterable(
-                                expand_range(r)
-                                for r in line.strip().split(",")
-                            )
-                        )
-                    )
-                )
+                sibling_sets.add(extract_values(line.strip()))
             return tuple(
                 CPUInfo(siblings) for siblings in sorted(sibling_sets)
             )
@@ -150,8 +141,21 @@ class System:
 
 
 def expand_range(value: str) -> tuple[int, ...]:
+    if value == "":
+        return tuple()
     if "-" not in value:
         return tuple((int(value),))
     start, stop = value.split("-")
 
     return tuple(range(int(start), int(stop) + 1))
+
+
+def extract_values(line: str) -> tuple[int, ...]:
+    return tuple(
+        sorted(
+            chain.from_iterable(
+                expand_range(r)
+                for r in line.strip().split(",")
+            )
+        )
+    )
--- a/legate/util/system.py
+++ b/legate/util/system.py
@@ -19,6 +19,7 @@ import os
 import platform
 import sys
 from functools import cached_property
+from itertools import chain
 
 from .fs import get_legate_paths, get_legion_paths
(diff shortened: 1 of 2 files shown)


legate/util/system.py
```diff
@@ -86,6 +86,13 @@ def LIB_PATH(self) -> str:
     def cpus(self) -> tuple[CPUInfo, ...]:
         """A list of CPUs on the system."""
 
+        def expand_range(value: str) -> tuple[int, ...]:
+            if "-" not in value:
+                return tuple((int(value),))
+            start, stop = value.split("-")
+
+            return tuple(x for x in range(int(start), int(stop) + 1))
```
bryevdv, line 94: can just call `tuple` on the `range` directly, no need for a generator
```suggestion
            return tuple(range(int(start), int(stop) + 1))
```

legate/util/system.py
```diff
@@ -98,9 +105,13 @@ def cpus(self) -> tuple[CPUInfo, ...]:
...
-                sibling_sets.add(
-                    tuple(sorted(int(x) for x in line.strip().split(",")))
-                )
+
+                flattened = []
+                for x in (
+                    expand_range(value) for value in line.strip().split(",")
+                ):
+                    flattened += [y for y in x]
+                sibling_sets.add(tuple(sorted(flattened)))
```
bryevdv, line 114: `itertools.chain` could simplify here:
```python
In [10]: line = "1-3,4,5,8-12,15,20"

In [11]: tuple(sorted(chain.from_iterable(expand_range(r) for r in line.strip().split(","))))
Out[11]: (1, 2, 3, 4, 5, 8, 9, 10, 11, 12, 15, 20)
```
In fact with `chain`  you could have `expand_range` just return the `range` directly, and not even call `tuple` to materialize it

legate/util/system.py
bryevdv: Also `expand_range` doesn't really need to be a closure I don't think. It could go in `legate.util` and then it would be possible to add plain pytest unit tests for it.

vzhurba01: Done. Added test_expand_range.

legate/util/system.py
vzhurba01: That's amazing, done!

I'll keep the tuple to so that the type checking for pre-commit stays happy.

bryevdv (approved):
Looks great!

tests/unit/legate/util/test_system.py
```diff
@@ -105,3 +106,29 @@ def test_gpus_osx(self) -> None:
...
+        new_line = (1, 3)
+        assert convert(old_line) == new_line
+
+        old_line = "1-3"
+        new_line = (1, 2, 3)
+        assert convert(old_line) == new_line
+
+        old_line = "1-3,4,5,8-12,15,20"
+        new_line = (1, 2, 3, 4, 5, 8, 9, 10, 11, 12, 15, 20)
+        assert convert(old_line) == new_line
```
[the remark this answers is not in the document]
vzhurba01: Thank you so much for the detailed explanation! Done.

tests/unit/legate/util/test_system.py
```diff
@@ -105,3 +105,77 @@ def test_gpus_osx(self) -> None:
...
+    @pytest.mark.parametrize(
+        "val", ("0,1-2", "1-2,0", "0,1-2,3,4-5,6", "5-6,4,1-3,0")
+    )
+    def test_mixed(self, val: str) -> None:
+        expected = {
+            "0,1-2": (0, 1, 2),
+            "1-2,0": (0, 1, 2),
+            "0,1-2,3,4-5,6": (0, 1, 2, 3, 4, 5, 6),
+            "5-6,4,1-3,0": (0, 1, 2, 3, 4, 5, 6),
+        }
```
bryevdv, line 180: BTW just FYI it's common to move values + expected values together at the same time through parameters. The danger here is that the keys in the `expected` dict become out of sync with the `val` parameter values since they are duplicated. I think this is fine in this small case, but if you are interested there are more examples, etc on the pytest site

vzhurba01: Created PR https://github.com/nv-legate/legate.core/pull/851 because I really like the non-duplicating version.

PR #850, opened 2023-10-04
Add missing 12.0 CUDA libraries to env generation script

manopapad:
These libraries used to be included automatically with one of the other CUDA packages, but not anymore, so we have to include them manually.


m3vaz (approved):
PR should remove the manual install in the CI files to make sure it's flowing through everywhere. Otherwise, LGTM

PR #851, opened 2023-10-04
Use testdata to remove duplicate test dictionary

vzhurba01:


tests/unit/legate/util/test_system.py
```diff
@@ -133,49 +133,54 @@ def test_errors(self) -> None:
...
+    @pytest.mark.parametrize("val,expected", testdata_range_ordered)
+    def test_range_ordered(self, val: str, expected: tuple[int, ...]) -> None:
+        assert m.extract_values(val) == expected
+
+    testdata_mixed = {
+        ("0,1-2", (0, 1, 2)),
+        ("1-2,0", (0, 1, 2)),
+        ("0,1-2,3,4-5,6", (0, 1, 2, 3, 4, 5, 6)),
+        ("5-6,4,1-3,0", (0, 1, 2, 3, 4, 5, 6)),
+    }
```
bryevdv, line 182: Only comment: some are lists of tuples and some are sets of tuples? FWIW I think you could make the `testdata_foo` values all be dicts, and then pass e.g. `testdata_foo.items()` to the `parametrize` call

bryevdv (approved):
One comment to consider, otherwise LGTM

tests/unit/legate/util/test_system.py
vzhurba01: Good catch. Fixed.

PR #861, opened 2023-10-16
Use conda for all dependencies.

trivialfis:
For consistency and performance.

I use `mamba` for creating environments, which is much faster than the resolver from pip.

### DIFF
--- a/scripts/generate-conda-envs.py
+++ b/scripts/generate-conda-envs.py
@@ -488,7 +488,7 @@ if __name__ == "__main__":
             use=config.use,
             python=config.python,
             conda_sections=conda_sections,
-            pip=PIP_TEMPLATE.format(pip_sections=pip_sections),
+            pip=PIP_TEMPLATE.format(pip_sections=pip_sections) if pip_sections else "",
         )
         with open(f"{filename}.yaml", "w") as f:
             f.write(out)
--- a/scripts/generate-conda-envs.py
+++ b/scripts/generate-conda-envs.py
@@ -193,11 +193,12 @@ class TestsConfig(SectionConfig):
             "pytest",
             "types-docutils",
             "pynvml",
+            "tifffile",
         )
 
     @property
     def pip(self) -> Reqs:
-        return ("tifffile",)
+        return ()
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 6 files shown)


scripts/generate-conda-envs.py
```diff
@@ -485,7 +488,9 @@ def section_selected(section):
             use=config.use,
             python=config.python,
             conda_sections=conda_sections,
-            pip=PIP_TEMPLATE.format(pip_sections=pip_sections),
+            pip=PIP_TEMPLATE.format(pip_sections=pip_sections)
+            if pip_sections
+            else "",
```
bryevdv, line 493: I think it would be better to handle this in the template, to keep all the cases here consistent

bryevdv:
> better suited for pip sourcing

@manopapad No special concerns from me.  If they were put in `pip` section originally then that's because there was some issue obtaining then via `conda` at the time. If the issues are now resolved then switching to using `conda` seems fine. (Would just acknowledge that if anything ever becomes difficult to obtain via conda it might need to get moved to `pip`  in general).

Just one comment regarding moving some logic to the jinja template

scripts/generate-conda-envs.py
trivialfis: Thank you for the suggestion, reverted the change.

trivialfis:
I have tested with `mamba env update -f`, but your suggestion is better, I will omit the pip section.

trivialfis:
Looking into the code, I think the generator doesn't depend on jinja2, actually, it doesn't have any dependency since it's the script used to list dependencies in the first place. I think we might want to stay with an empty section or use conditions in the Python format function.

bryevdv:
Oh right we had used jinja originally but then removed it, and I misled myself with the `_TEMPLATE` name. In that case yes your original code seems appropriate.

manopapad:
So this is good to merge then?

bryevdv:
No I think the going back to 
```
pip=PIP_TEMPLATE.format(pip_sections=pip_sections)
            if pip_sections
            else "",
```
is probably the best idea.

trivialfis:
> No I think the going back to the original

Done.

bryevdv (approved):
LGTM @manopapad

Issue #862, opened 2023-10-16
New warnings on non-editable build


manopapad:
@bryevdv does anything look concerning?

```
  -- Installing: /home/mpapadakis/noneditable/legate.core/_skbuild/linux-x86_64-3.10/cmake-install/lib/liblegion_canonical_python.so
  '/home/mpapadakis/noneditable/env/bin/python3' '-m' 'pip' 'install' '--global-option=--cmake-build-dir' '--global-option=/home/mpapadakis/noneditable/legate.core/_skbuild/linux-x86_64-3.10/cmake-build/_deps/legion-build/runtime' '--global-option=--prefix' '--global-option=/home/mpapadakis/noneditable/legate.core/_skbuild/linux-x86_64-3.10/cmake-install' '--ignore-installed' '--root' '/' '--prefix' '/home/mpapadakis/noneditable/env' '.'
  DEPRECATION: --build-option and --global-option are deprecated. pip 24.0 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859
  WARNING: Implying --no-binary=:all: due to the presence of --build-option / --global-option.
```

```
  /home/mpapadakis/noneditable/env/lib/python3.10/site-packages/setuptools/command/build_py.py:204: _Warning: Package 'legate.core._lib' is absent from the `packages` configuration.
  !!

          ********************************************************************************
          ############################
          # Package would be ignored #
          ############################
          Python recognizes 'legate.core._lib' as an importable package[^1],
          but it is absent from setuptools' `packages` configuration.

          This leads to an ambiguous overall configuration. If you want to distribute this
          package, please make sure that 'legate.core._lib' is explicitly added
          to the `packages` configuration field.

          Alternatively, you can also rely on setuptools' discovery methods
          (for example by using `find_namespace_packages(...)`/`find_namespace:`
          instead of `find_packages(...)`/`find:`).

          You can read more about "package discovery" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

          If you don't want 'legate.core._lib' to be distributed and are
          already explicitly excluding 'legate.core._lib' via
          `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
          you can try to use `exclude_package_data`, or `include-package-data=False` in
          combination with a more fine grained `package-data` configuration.

          You can read more about "package data files" on setuptools documentation page:

[...]

PR #865, opened 2023-10-18
make _lib a proper module

bryevdv:
This clears the `pip install` warning about "Package would be ignored" described in #862

All `cunumeric` tests pass when built against an wheel install with this change. Since we do intend there to be `legate.core._lib` package, this seems like the correct action, or at the very least, a harmless resolution.


manopapad:
@Jacobfaib any concerns with this?

Jacobfaib (approved):

PR #866, opened 2023-10-18
Drop compilers and OpenMPI from conda env generator

manopapad:
@sandeepd-nv @mag1cp1n If we don't pull the compilers from conda anymore, I'm going to need some help on how to adjust the CI scripts (if you think it's necessary).


manopapad:
I adjusted the approach according to @m3vaz's feedback. Apparently the `--compilers` and `--openmpi` options are still used in CI, so we want to keep them, but we can turn them off by default.

bryevdv (approved):

scripts/generate-conda-envs.py
```diff
@@ -382,42 +364,39 @@ def __call__(self, parser, namespace, values, option_string):
...
         choices=OS_NAMES,
-        default=None,
-        help="OS to generate for (default: all OSes)",
+        default=("osx" if sys.platform == "darwin" else "linux"),
+        help="OS to generate for",
     )
     parser.add_argument(
         "--compilers",
         action=BooleanFlag,
-        dest="compilers",
```
m3vaz, line 404: Can we retain the `dest` arg for the parse?

scripts/generate-conda-envs.py
```diff
@@ -382,42 +364,39 @@ def __call__(self, parser, namespace, values, option_string):
...
-        default=None,
-        help="Whether to include conda compilers or not (default: both)",
+        default=False,
+        help="Whether to include conda compilers or not",
     )
     parser.add_argument(
         "--openmpi",
         action=BooleanFlag,
-        dest="openmpi",
```
m3vaz, line 411: Can we retain the `dest` arg for the parse?

scripts/generate-conda-envs.py
```diff
@@ -382,42 +364,39 @@ def __call__(self, parser, namespace, values, option_string):
...
-        default=None,
-        help="Whether to include openmpi or not (default: both)",
+        default=False,
+        help="Whether to include openmpi or not",
     )
     parser.add_argument(
         "--ucx",
         action=BooleanFlag,
-        dest="ucx",
```
m3vaz, line 418: Can we retain the `dest` arg for the parse?

m3vaz (approved):
marking as +1 since the removal of the CTK is just doing some housekeeping
would like to keep the explicit `dest` as well but non-blocking issue

scripts/generate-conda-envs.py
```diff
@@ -276,11 +276,6 @@ def filename(self) -> str:
 
 CTK_VERSIONS = (
     "none",
-    "10.2",
```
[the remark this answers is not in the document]
m3vaz: this is housekeeping, marking as resolved.

PR #876, opened 2023-10-26
Mac CI

sandeepd-nv:


.github/workflows/gh-build-and-test.yml
```diff
@@ -0,0 +1,71 @@
...
+            test-scope: unit
+            runner: linux-amd64-2gpu
+            enabled: ${{ inputs.platform == 'linux' }}
+
+          - name: mypy
+            test-scope: mypy
+            runner: linux-amd64-gpu-v100-latest-1
+            enabled: ${{ inputs.platform == 'linux' }}
+
+          # - name: mypy (OS X)
```
m3vaz, line 55: Do we have someone who's following up with the things mypy is flagging?

conda/conda-build/meta.yaml
```diff
@@ -114,8 +115,10 @@ requirements:
     - rust
     - ninja
     - cmake {{ cmake_version }}
-    - {{ compiler('c') }} =11.2
-    - {{ compiler('cxx') }} =11.2
+    - {{ compiler('c') }} =11.2   # [linux]
+    - {{ compiler('cxx') }} =11.2 # [linux]
+    - {{ compiler('c') }}   # [osx]
```
m3vaz, line 120: Can you align the selectors so `# [osx]` is under `# [linux]`

continuous_integration/scripts/build-legate
```diff
@@ -0,0 +1,75 @@
...
+    # Navigate to the parent of the parent of SCRIPT_DIR, then get the full path
+    REPO_DIR="$(cd "$SCRIPT_DIR/../.." && pwd)"
+
+    export REPO_DIR
+
+    export PATH="${PATH}:${REPO_DIR}/continuous_integration/scripts"
+    export ARTIFACTS_DIR="${REPO_DIR}/.artifacts"
+}
+
+# export CONDA_BUILD_THREADS=1
```
m3vaz, line 18: debugging?

continuous_integration/scripts/build-legate
```diff
@@ -0,0 +1,75 @@
...
+}
+
+copy_ci_artifacts() {
+    echo Copying CI artifacts
+
+    cp -r /tmp/out          "$ARTIFACTS_DIR"
+    cp -r /tmp/conda-build  "$ARTIFACTS_DIR"
+}
+
+# NOTE: This script must be run from the project's root folder
```
m3vaz, line 47: Not necessarily something that needs to be done as part of the PR, but if the script is in a known place, we can use the location of the script to get back to the root folder.

continuous_integration/scripts/build-legate-cpp
```diff
@@ -6,7 +6,11 @@ build_legate_cpp() {
     # Build + package legate.core C++ libs
     local cmake_args=(${CMAKE_ARGS:-});
     cmake_args+=(-DBUILD_SHARED_LIBS=ON);
-    cmake_args+=(-DBUILD_MARCH=${BUILD_MARCH:-haswell});
+
+    if [[ ! "$(uname)" == "Darwin" ]]; then
```
m3vaz, line 10: `A != B` instead of `A == B`?

continuous_integration/scripts/setup-utils
```diff
@@ -0,0 +1,268 @@
...
+        git config --global --add user.name "anon"
+    else
+        echo "Note: git user.name is already set to $current_name"
+    fi
+
+    # Fix "fatal: detected dubious ownership in repository at '/tmp/legate.core'"
+    # during local builds.
+    git config --global --add safe.directory "$REPO_DIR"
+}
+
```
m3vaz, line 187: Can you add a seperator line here, since everything above is OS-specific helper functions, everything below is a generalized function?

m3vaz (changes requested):
Left a few comments, but nothing major. In principle +1

conda/conda-build/meta.yaml
sandeepd-nv: Done.

continuous_integration/scripts/build-legate
sandeepd-nv: Removed.

continuous_integration/scripts/build-legate
```diff
@@ -0,0 +1,75 @@
...
+# export JOBS=1
+
+build_legate_ci() {
+    build-legate-cpp;
+    build-legate-wheel;
+    build-legate-conda;
+}
+
+build_legate_release() {
+    # export NVCC_PREPEND_FLAGS='--verbose'
```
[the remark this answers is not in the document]
sandeepd-nv: Removed.

continuous_integration/scripts/build-legate
sandeepd-nv: > if the script is in a known place, we can use the location of the script to get back to the root folder.
Already doing this, just forgot to remove the comment. The comment has been removed now.

continuous_integration/scripts/build-legate-cpp
sandeepd-nv: Done.

continuous_integration/scripts/build-legate-cpp
```diff
@@ -6,7 +6,11 @@ build_legate_cpp() {
...
+        cmake_args+=(-DBUILD_MARCH=${BUILD_MARCH:-x86-64});
```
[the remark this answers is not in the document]
sandeepd-nv: Removed the potentially incorrect default value from this line. Now BUILD_MARCH will be set in build-legate.

continuous_integration/scripts/conda-utils
```diff
@@ -0,0 +1,99 @@
...
+
+    return 0;
+}
+
+get_yaml_and_make_conda_env() {
+    set -e;
+
+    local yaml_file="";
+
+    # if ! find_yaml_file; then
```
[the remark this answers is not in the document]
sandeepd-nv: Removed.

continuous_integration/scripts/setup-utils
sandeepd-nv: Done.

.github/workflows/gh-build-and-test.yml
sandeepd-nv: I have asked @manopapad to help with this.

.github/workflows/gh-build.yml
```diff
@@ -54,57 +60,34 @@ jobs:
...
+            -e AWS_REGION \
+            -e AWS_SESSION_TOKEN \
+            -e AWS_ACCESS_KEY_ID \
+            -e AWS_SECRET_ACCESS_KEY \
+            -e GITHUB_TOKEN \
+            -e USE_CUDA \
+            -v "$(pwd):$(pwd)" \
+            -v "$ARTIFACTS_DIR:$(pwd)/.artifacts" \
+            --rm "condaforge/miniforge3:latest" \
+            /bin/bash -c "cd $(pwd) && continuous_integration/scripts/build-legate ${{ inputs.build-type}}"
```
marcinz, line 78: Do we need to run docker directly? Isn't there a way to use the `container` feature of the github runner? It seems that we should run the same `build-legate` script, but either run it in a container or not, right?

.github/workflows/gh-test.yml
```diff
@@ -18,24 +18,84 @@ on:
...
+        type: boolean
+
+env:
+  ARTIFACT_NAME: "${{ inputs.repos-name }}-${{ inputs.build-target }}-${{ inputs.sha }}"
+  ARTIFACTS_DIR: "${{ github.workspace }}/.artifacts"
 
 jobs:
-  test:
-    if: github.repository_owner == 'nv-legate'
+  test-in-container:
```
marcinz, line 33: I guess the same comment here. It seems that we are repeating pretty much the same steps, with the major difference being running in a container or not. Wouldn't we want the steps in a separate reusable workflow, and then we would just invoke the workflow from a container or without a container?

.github/workflows/gh-build.yml
sandeepd-nv: AFAIK the container feature cannot be conditionally switched off (say using if). So we will need to ...
1. Convert gh-build.yml in to a local action.
2. Create two different reusable workflows, one with container and one without and call the above local action from these workflows. 

The reusable part part come out to only 3 steps: checkout, build and upload. IMO the reusability is minimal (due to the limitations of the Github workflow markup language) and not worth the effort. Especially since the github container directive does pretty much  the same thing.

.github/workflows/gh-test.yml
sandeepd-nv: See above comment.

.github/workflows/gh-build.yml
```diff
@@ -36,15 +38,19 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+
     steps:
       - name: Checkout ${{ inputs.repos-name }} (= this repo)
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          persist-credentials: false
```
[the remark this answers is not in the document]
sandeepd-nv: Not sure what this does. Picked this up from other nv-legate repositories.

m3vaz (approved):

.github/workflows/gh-build.yml
marcinz: I think it may make sense to do that in the long run. Basically, creating one workflow that does the work so to say, and then have different call sites for it. Right now, we have dispatch on `use-container` anyway, so I think we actually don't add anything if we do the dispatch in a different place. Yes, things would have to move around a bit, but the benefit would be a factored out build workflow that can be called from different contexts (e.g., different driver versions, different setups, etc.).

.github/workflows/gh-build-and-test.yml
sandeepd-nv: It's fixed now.

.github/workflows/ci-gh.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "branch-*"
+
+jobs:
+  build-and-test:
+    strategy:
+      fail-fast: false
+      matrix:
```
marcinz, line 17: I think we can move container into this matrix, and then use it in the top level in gh-build-and-test. I think that should work, as everything else either runs in a container or not. But we could do that in a separate PR.

continuous_integration/scripts/build-legate-wheel
```diff
@@ -21,15 +21,17 @@ build_legate_wheel() {
 
     local cmake_args=(${CMAKE_ARGS:-});
     cmake_args+=("-DFIND_LEGATE_CORE_CPP=ON");
-    cmake_args+=("-Dlegate_core_ROOT=$HOME/legate/build");
+    cmake_args+=("-Dlegate_core_ROOT=$REPO_DIR/legate/build");
```
sandeepd-nv, line 24: This should have been $REPO_DIR/build.

PR #878, opened 2023-10-26
Remove extraneous PYTHONPATH entry

manopapad:
It doesn't seem like this is necessary. Removing it because on non-editable installations it ends up adding `<venv-root>/lib/pythonX.Y/site-packages/legate` to the `PYTHONPATH`, which causes sub-modules under legate to potentially clash with top-level modules of the same name.


jjwilke (approved):
LGTM

PR #885, opened 2023-11-15
adding a workaround  for NumPy's conversion from unit64 to float

ipdemes:
adding a workaround  for NumPy's expected behavior when adding an integer to np.uint64, resulting in float64
see https://github.com/numpy/numpy/issues/5745

### DIFF
--- a/legate/core/store.py
+++ b/legate/core/store.py
@@ -1233,14 +1233,7 @@ class Store:
             If ``dim`` is not a valid dimension name or ``index`` is
             out of bounds
         """
-        # This is a workaround for NumPy's expected behavior when adding
-        # an integer to np.uint64, resulting in float64.
-        if type(index) is not int:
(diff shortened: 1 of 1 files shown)


legate/core/store.py
```diff
@@ -1233,6 +1233,14 @@ def project(self, dim: int, index: int) -> Store:
...
             out of bounds
         """
+        # This is a workaround for NumPy's expected behavior when adding
+        # an integer to np.uint64, resulting in float64.
+        if type(index) is not int:
+            import numpy as np
+
+            assert np.issubdtype(type(index), np.integer)
+            index = int(index)
+
```
magnatelee, line 1243: the check itself is as expensive as the casting, so we better cast it to `int` all the time
```suggestion
        index = int(index)
```

magnatelee (approved):

PR #892, opened 2023-11-23
Restrict to cuTensor 1.X (2.X has a different API)

manopapad:


m3vaz (approved):

m3vaz:
This may hit a runtime dependency later on since the meta.yaml doesn't constrain the cutensor still, but othwerise merging to unblock CI.

PR #899, opened 2023-11-30
Print driver output only on head nodes

bryevdv:
This PR prevents informational messages, e.g. profiling output locations, from being printed on every node in a multi-node scenario.

### DIFF
--- a/legate/driver/driver.py
+++ b/legate/driver/driver.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from datetime import datetime
+from io import StringIO
 from shlex import quote
 from subprocess import run
 from textwrap import indent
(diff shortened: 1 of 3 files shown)


legate/driver/driver.py
```diff
@@ -226,35 +224,36 @@ def print_verbose(
...
     Returns
     -------
-        None
+        str
 
     """
+    out = StringIO()
 
-    print(f"\n{rule('Legion Python Configuration')}")
+    out.write(f"\n{rule('Legion Python Configuration')}")
```
manopapad, line 232: `out.write` doesn't add the newline automatically, as `print` does.

manopapad (changes requested):

legate/driver/driver.py
bryevdv: doh https://github.com/nv-legate/legate.core/pull/899/commits/9b97a36ef085fe0cd6c087f46caad89a979d8a4f

manopapad (approved):

PR #908, opened 2023-12-06
Pass cmake executable to scikit-build

### LINKED ISSUE #837: install.py ignores --with-cmake for scikit-build
install.py is not passing cmake_exe to scikit-build, which results in that part of the build process using the wrong cmake. 

```
(/sdf/home/s/seshu/cunumeric-cpu) [seshu@sdfiana002 legate.core]$ ./install.py --legion-src-dir /sdf/home/s/seshu/cunumeric-cpu/legion --openmp --no-cuda --network ucx -j8 --verbose --clean --with-cmake /sdf/home/s/seshu/cunumeric-cpu/bin/cmake

Verbose build is on                                                                                                    
networks: ['ucx']                                                                                                      
cuda: False                                                                                                            arch: all-major                                                                                                        
openmp: True                                                                                                           
march: haswell                                                                                                         hdf: False                                                                                                             
llvm: False                                                                                                            
spy: False                                                                                                             
build_docs: False                                                                                                      
conduit: None                                                                                                          
gasnet_system: None                                                                                                    
nccl_dir: None                                                                                                         
cmake_exe: /sdf/home/s/seshu/cunumeric-cpu/bin/cmake  
...
```

I can see it still calls /usr/bin/cmake3:

```
 Not searching for unused variables given on the command line.
  CMake Error at CMakeLists.txt:17 (cmake_minimum_required):
    CMake 3.22.1 or higher is required.  You are running version 3.20.2
                           
                                                                                                                      
  -- Configuring incomplete, errors occurred!
    File "/sdf/home/s/seshu/cunumeric-cpu/lib/python3.9/site-packages/skbuild/setuptools_wrap.py", line 666, in setup        env = cmkr.configure(
    File "/sdf/home/s/seshu/cunumeric-cpu/lib/python3.9/site-packages/skbuild/cmaker.py", line 357, in configure
      raise SKBuildError(msg)                                                                                              
  An error occurred while configuring with CMake.
    Command:                                                                                                                 /usr/bin/cmake3 /sdf/home/s/seshu/cunumeric-cpu/legate.core...
```

manopapad:
Fixes #837

### DIFF
--- a/install.py
+++ b/install.py
@@ -315,6 +315,13 @@ def install(
     if legion_dir is not None and legion_src_dir is not None:
         sys.exit("Cannot specify both --legion-dir and --legion-src-dir")
 
+    join = os.path.join
+    exists = os.path.exists
+    dirname = os.path.dirname
+    realpath = os.path.realpath
+
+    cmake_exe = realpath(cmake_exe)
+
     print(f"Verbose build is {'on' if verbose else 'off'}")
     if verbose:
         print(f"networks: {networks}")
@@ -352,11 +359,6 @@ def install(
         print(f"legion_url: {legion_url}")
         print(f"legion_branch: {legion_branch}")
         print(f"unknown: {unknown}")
-
-    join = os.path.join
-    exists = os.path.exists
-    dirname = os.path.dirname
-    realpath = os.path.realpath
 
     legate_core_dir = dirname(realpath(__file__))
 
@@ -469,6 +471,8 @@ def install(
 
     if not build_isolation:
         pip_install_cmd += ["--no-build-isolation"]
+
+    pip_install_cmd += [f"--global-option=--cmake-executable={cmake_exe}"]
 
     if editable:
         cmd_env.update({"SETUPTOOLS_ENABLE_FEATURES": "legacy-editable"})


install.py
```diff
@@ -315,6 +315,13 @@ def install(
     if legion_dir is not None and legion_src_dir is not None:
         sys.exit("Cannot specify both --legion-dir and --legion-src-dir")
 
+    join = os.path.join
+    exists = os.path.exists
+    dirname = os.path.dirname
+    realpath = os.path.realpath
```
bryevdv, line 321: What's wrong with 
```python
from  os.path import dirname, exists, join, realpath
```
at the top? Then several places can be shortened

bryevdv:
> give up and ask scikit-build to let us control this through an environment variable

At this point I've resorted to env vars for passing config to build backends in pretty much every situation it's come up. I'm surprised skbuild doesn't already offer the option, is definitely worth an ask.

manopapad:
Asked at https://github.com/scikit-build/scikit-build/issues/1045

vyasr:
Unfortunately I don't know of a good way to control this in scikit-build right now either. Based on [the code for its CMake inference](https://github.com/scikit-build/scikit-build/blob/9787cc2582311f731b3d0411b1175bd852db2ad6/skbuild/constants.py#L17) it should your PATH settings, so perhaps you could force the desired CMake to the front of the PATH? It's just calling `shutil.which` under the hood.

CharlelieLrt:
Just to precise, the desired cmake was already to the front of my PATH, but I think skbuild is picking up in priority cmake3 (which is NOT the desired cmake and is NOT to the front of the PATH).

vyasr:
Oh that is super unfortunate, I missed that it was searching for cmake3. Does it work for you if you alias cmake3 to the desired cmake? (Obviously a workaround, but trying to find a reasonable one for you while we follow up on the scikit-build front)

CharlelieLrt:
Yes, I just symlinked cmake3 to the desired cmake and it worked.

vyasr:
OK looked into this a bit further and chatted with scikit-build devs. scikit-build currently takes the first CMake executable that it finds, then errors if the CMake version is too old. That behavior is what you're observing above. That behavior is fixed in the newer scikit-build-core builder ([which I've just completed migrating all of RAPIDS to](https://github.com/rapidsai/build-planning/issues/2)), and I would recommend that legate look into doing that at some point in the future to get the latest and greatest. If using scikit-build-core, while you still wouldn't be able to specify what CMake to use, as long as one of the CMake executables on the path was new enough you would bypass this issue.

Separately, for the case of actually passing in a specific executable I've opened https://github.com/scikit-build/scikit-build-core/issues/576. The most likely way that [the scikit-build issue](https://github.com/scikit-build/scikit-build/issues/1045) will be resolved is by scikit-build-core implementing this feature and then scikit-build classic's internals being refactored to use scikit-build-core. That's always been the long-term goal for unifying the projects.

vyasr:
Just FYI, scikit-build-core now supports specifying `CMAKE_EXECUTABLE` as of 0.8.0.

PR #910, opened 2023-12-09
Use labels from the nvidia channel to restrict CTK versions

manopapad:
Package versions on the nvidia channel don't always correspond to CTK version (e.g. cusparse 12.2 can be part of CTK 12.3). The only robust way to ensure that all selected nvidia conda channel packages are coming from a compatible CTK collection is to restrict based on label.


bryevdv (approved):
What a mess 😮

m3vaz (approved):
> (e.g. cusparse 12.2 can be part of CTK 12.3). 

This sounds *very* wrong.

+1 for the workaround.

Issue #904, opened 2023-12-04
Build from source on Apple Silicon: march=native not supported


CharlelieLrt:
I am trying to build legate locally from source on a M1 Macbook Pro (Apple Silicon), following the instructions given in the [BUILD.md](https://github.com/nv-legate/legate.core/blob/HEAD/BUILD.md#dependency-listing)
OS: MacOS 12.6.7
Dependencies are pulled with conda with the following environment config (generated with `./generate-conda-envs.py --python 3.10 --os osx`):

```yaml
name: legate-test
channels:
  - conda-forge
  - nvidia
dependencies:

  - python=3.10,!=3.9.7  # avoid https://bugs.python.org/issue45121

  # build
  - cmake>=3.24,!=3.25.0
  - cython
  - git
  - make
  - ninja
  - numba
  - openssl
  - pkg-config
  - rust
  - scikit-build>=0.13.1
  - setuptools>=60
  - zlib

  # runtime
  - cffi
  - llvm-openmp
  - numpy>=1.22
  - libblas=*=*openblas*
  - openblas=*=*openmp*
  - openblas<=0.3.21
  - opt_einsum
  - scipy
  - typing_extensions

  # tests
  - clang-tools>=8
  - clang>=8
  - colorama
  - coverage
  - mock
  - mypy>=0.961
  - pre-commit
  - pytest-cov
  - pytest-lazy-fixture
  - pytest-mock
  - pytest
  - types-docutils
  - pynvml
  - tifffile

  # docs
  - pandoc
  - doxygen
  - ipython
  - jinja2
  - markdown<3.4.0
  - pydata-sphinx-theme>=0.13
  - myst-parser
  - nbsphinx
  - sphinx-copybutton
  - sphinx>=4.4.0
```

Legate is built with `./install.py --max-dim 5 --openmp --hdf5 --build-tests --build-examples`, which fails with the following error:

```console
 Not searching for unused variables given on the command line.
 -- The C compiler identification is AppleClang 14.0.0.14000029
 -- The CXX compiler identification is AppleClang 14.0.0.14000029
 -- Detecting C compiler ABI info
 -- Detecting C compiler ABI info - done
 -- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
 -- Detecting C compile features
 -- Detecting C compile features - done
 -- Detecting CXX compiler ABI info
 -- Detecting CXX compiler ABI info - done
[...]

PR #911, opened 2023-12-09
Fix compilation error with std::min on MacOS

manopapad:
See https://github.com/nv-legate/legate.core/issues/904#issuecomment-1839813308

### DIFF
--- a/examples/io/src/read_file.cc
+++ b/examples/io/src/read_file.cc
@@ -48,7 +48,8 @@ struct read_fn {
     // Compute the absolute offsets to the section that this reader task
     // is supposed to read from the file
     int64_t my_lo = my_id * size / num_readers;
-    int64_t my_hi = std::min((my_id + 1) * size / num_readers, size);
+    int64_t my_hi =
+      std::min(static_cast<int64_t>((my_id + 1) * size / num_readers), static_cast<int64_t>(size));
 
     // Then compute the extent for the output and create the output buffer to populate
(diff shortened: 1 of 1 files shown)


examples/io/src/read_file.cc
```diff
@@ -48,7 +48,8 @@ struct read_fn {
     // Compute the absolute offsets to the section that this reader task
     // is supposed to read from the file
     int64_t my_lo = my_id * size / num_readers;
-    int64_t my_hi = std::min((my_id + 1) * size / num_readers, size);
+    int64_t my_hi =
+      std::min(static_cast<int64_t>((my_id + 1) * size / num_readers), static_cast<int64_t>(size));
```
Jacobfaib, line 52: I think it is safe to cast just `size` to `int64_t` here. i.e.
```c++
auto ssize = static_cast<int64_t>(size);
int64_t my_hi =
      std::min((my_id + 1) * ssize / num_readers, ssize);
```

manopapad: @CharlelieLrt could you please check if this change works for you also? I can't check myself :-(

CharlelieLrt: Yes, it works.

manopapad: fe0205d

Jacobfaib (approved):

manopapad:
CI failures are unrelated, merging

PR #912, opened 2023-12-09
Remove some irrelevant Legion build options

manopapad:


Jacobfaib (approved):
Internal needs the same change as well?

manopapad:
Yes, I will propagate the changes today

PR #914, opened 2023-12-11
Add cuda-cudart-static to generated conda environments

marcinz:
Adding this package should resolve the build issues that have cropped up recently.

Issues that came up in the process:

- Fixed the runner that does not exist anymore
- Update cuda version to a format introduced in another PR
- Update the documentation
- Change to using labels for the nvidia channel
- Fix a test issue that was crashing the tests on OS X
- Do not use GPU runners for CPU tests
- Provide an option for not building the profiler and turn it off in the CI builds (provides a good speedup)
- Switch to using cuda 12.2.2
- Require that cuda is installed (__cuda requirement) in the meta.yaml, but do not request any specific version
- Add openmpi to requirements if we build with UCX

### DIFF
--- a/cmake/thirdparty/get_legion.cmake
+++ b/cmake/thirdparty/get_legion.cmake
@@ -121,7 +121,7 @@ function(find_or_configure_legion)
       set(Legion_BACKTRACE_USE_LIBDW OFF)
     endif()
 
-    set(Legion_BUILD_RUST_PROFILER OFF CACHE STRING)
+    set(Legion_BUILD_RUST_PROFILER OFF CACHE BOOL "Whether to build the Legion profiler" FORCE)
 
     rapids_cpm_find(Legion ${version} ${FIND_PKG_ARGS}
         CPM_ARGS
--- a/cmake/thirdparty/get_legion.cmake
+++ b/cmake/thirdparty/get_legion.cmake
@@ -121,6 +121,8 @@ function(find_or_configure_legion)
       set(Legion_BACKTRACE_USE_LIBDW OFF)
     endif()
 
+    set(Legion_BUILD_RUST_PROFILER OFF CACHE STRING)
+
     rapids_cpm_find(Legion ${version} ${FIND_PKG_ARGS}
         CPM_ARGS
           ${legion_cpm_git_args}
@@ -132,7 +134,6 @@ function(find_or_configure_legion)
                                  "Legion_BUILD_BINDINGS ON"
                                  "Legion_REDOP_HALF ON"
                                  "Legion_REDOP_COMPLEX ON"
-                                 "Legion_BUILD_RUST_PROFILER ON"
                                  "Legion_UCX_DYNAMIC_LOAD ON"
     )
   endif()
--- a/conda/conda-build/meta.yaml
+++ b/conda/conda-build/meta.yaml
@@ -39,7 +39,7 @@
 {% else %}
     {% set version = environ.get('GIT_DESCRIBE_TAG', default_env_var).lstrip('v') %}
 {% endif %}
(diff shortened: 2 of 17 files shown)


m3vaz (changes requested):
This would also need to be updated in meta.yaml so that the metadata is tagged on the output conda package

marcinz:
> This would also need to be updated in meta.yaml so that the metadata is tagged on the output conda package

Will do, but, for now, CI does not seem to be picking the jobs up.

marcinz:
@manopapad Do we need to run `mypy` if it runs as a part of pre-commit? Currently, we are running it for every build, but that does not seem to add any value if I am not wrong.

conda/conda-build/meta.yaml
```diff
@@ -177,13 +178,14 @@ requirements:
 {% endif %}
 {% if ucx_configured_bool %}
     - ucx >=1.14
+    - openmpi
```
marcinz, line 181: We would have probably caught this if we were testing our conda build. In the current build process, we have openmpi from the env generation script during the build, but it was not present in the package requirements.

conda/conda-build/meta.yaml
```diff
@@ -177,13 +178,14 @@ requirements:
...
 {% endif %}
 
   run_constrained:
     - __glibc >=2.17  # [linux]
     - python != 3.9.7
 {% if gpu_enabled_bool %}
-    - __cuda >={{ cuda_version }}
+    - __cuda
```
marcinz, line 188: Testing for a particular version of cuda does not seem to make sense as what we really need is a compatible driver, and there is no way to test for a driver. So, for now, we assume that it is enough to have any version of cuda, and if the driver is incompatible, we will get errors later.

continuous_integration/scripts/build-legate-wheel
```diff
@@ -21,7 +21,7 @@ build_legate_wheel() {
 
     local cmake_args=(${CMAKE_ARGS:-});
     cmake_args+=("-DFIND_LEGATE_CORE_CPP=ON");
-    cmake_args+=("-Dlegate_core_ROOT=$REPO_DIR/legate/build");
+    cmake_args+=("-Dlegate_core_ROOT=$REPO_DIR/build");
```
marcinz, line 24: This typo was causing the legate cpp from the first build not being found and the second build was actually building everything from scratch. Because the second build has only the extra cmake arguments required on the top of the cpp build, it was building a wrong configuration causing illegal instruction issues and entirely skipping ucx tests.

cmake/thirdparty/get_legion.cmake
```diff
@@ -121,6 +121,8 @@ function(find_or_configure_legion)
       set(Legion_BACKTRACE_USE_LIBDW OFF)
     endif()
 
+    set(Legion_BUILD_RUST_PROFILER OFF CACHE BOOL "Whether to build the Legion profiler" FORCE)
```
marcinz, line 124: We should probably unhardcode this in internal as well.

m3vaz (approved):

bryevdv (approved):
LGTM (test changes)

manopapad (approved):

manopapad:
> Do we need to run mypy if it runs as a part of pre-commit?

Sorry, I missed this. I don't think it's necessary to run mypy again if it's getting run as part of pre-commit (assuming we have something like pre-commit-ci running pre-commit also on the server side; we can't guarantee that developers will have run it locally before committing).

marcinz:
Yes, on this repo we have pre-commit.ci.

marcinz:
Actually, I will just leave mypy jobs as they are for now since we are not getting too much development going on here atm.

PR #917, opened 2023-12-19
improve multi-node detection

bryevdv:
This PR makes the detection of multi-node configurations (especially for SLURM) more robust.

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -113,16 +113,6 @@ def _get_slurm_config() -> tuple[int, int] | None:
 
     # use SLURM_TASKS_PER_NODE if it is given
     if tasks_per_node_env is not None:
-        if ntasks_env is not None:
-            raise ValueError(
-                "Ambiguous congfiguration, both SLURM_TASKS_PER_NODE and "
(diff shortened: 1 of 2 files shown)


legate/driver/args.py
```diff
@@ -46,50 +47,153 @@
...
+        if ntasks_env is not None:
+            raise ValueError(
+                "Ambiguous congfiguration, both SLURM_TASKS_PER_NODE and "
+                "SLURM_NTASKS are set."
+            )
+        if nprocs_env is not None:
+            raise ValueError(
+                "Ambiguous congfiguration, both SLURM_TASKS_PER_NODE and "
+                "SLURM_NPROCS are set."
+            )
```
manopapad, line 125: I think it's quite possible that all are defined at the same time, so I think we need to remove this check.

manopapad (approved):
Need to remove a check, otherwise looks good

legate/driver/args.py
bryevdv: https://github.com/nv-legate/legate.core/pull/917/commits/20bfa5505fb4c3295acddee06465d2b66661e56d

PR #918, opened 2023-12-20
SLURM_TASKS_PER_NODE is not always an integer

manopapad:
We're seeing it reported e.g. as "1(x2)"


bryevdv (approved):
Is there a spec for this value so that we could implement a proper solution?

PR #919, opened 2023-12-20
handle SLURM_TASKS_PER_NODE more thoroughly

bryevdv:
This PR allows for correct handling of cases like SLURM_TASKS_PER_NODE="3(x2)"`

### DIFF
--- a/legate/driver/args.py
+++ b/legate/driver/args.py
@@ -16,6 +16,7 @@
 #
 from __future__ import annotations
 
+import re
 from argparse import REMAINDER, ArgumentDefaultsHelpFormatter, ArgumentParser
 from os import getenv
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


legate/driver/args.py
```diff
@@ -99,12 +100,39 @@ def _get_mv2_config() -> tuple[int, int] | None:
...
+_SLURM_CONFIG_ERROR = (
+    "Expected SLURM_TASKS_PER_NODE to be a single integer ranks per node, or "
+    "of the form 'A(xB)' where A is an integer ranks per node, and B is an "
+    "integer number of nodes, got SLURM_TASKS_PER_NODE={value}"
+)
+
+
 def _get_slurm_config() -> tuple[int, int] | None:
     if not (nodes_env := getenv("SLURM_JOB_NUM_NODES")):
         return None
```
bryevdv, line 112: Should we always demand this? the new `A(xB)` case does not rely on `SLURM_JOB_NUM_NODES`

manopapad (approved):

legate/driver/args.py
bryevdv: half of the `SLURM_TASKS_PER_NODE` case does need it tho (single integer case). It seems like its always provided in any case, so I'll leave this as it is for now.

PR #929, opened 2024-01-12
Update Base Mapper to Work with the Modified Legion Mapping Interface for Replicating Tasks

lightsighter:
This pull request updates the `BaseMapper` class so that it works with new interface for replicating tasks in Legion and removes the old replication function which no longer exists in the mapping interface. This request should only be merged after the `shardrefine` branch is merged into the `control_replication` branch of Legion or it will likely break things.


magnatelee:
Changes in this PR have been migrated to the internal repo.

PR #932, opened 2024-01-26
Remove unnecessary flags in Legion CMake build

elliottslaughter:
The flag `--root` here is a workaround for buggy behavior in `pip install` where pip would ignore the `--prefix` setting when `--global-option` was supplied. As it turns out, `--global-option` is deprecated anyway so we removed in Legion. As a result, this is no longer required; the prefix is already set in the upstream project and the need for `--root` goes away with the removal of `--global-option`.

See this Legion MR for details:

https://gitlab.com/StanfordLegion/legion/-/merge_requests/1072

I'm not sure which branch to send this patch to so if I'm set the wrong one, let me know and I'll update it.


manopapad (approved):
Tested in on local machine and it seems to work, waiting on CI

manopapad:
Editable and non-editable installs seem to work still on my local machine. Merging from main branch, to see if that fixes CI...

PR #941, opened 2024-04-24
Update README.md

bryevdv:
Omit extraneous info about forcing CPU-only installs from the README

No link to install on this PR since there is not an install user guide section but if this repo is going away eventually then does not seem very important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ Only linux-64 packages are available at the moment.
 The default package contains GPU support, and is compatible with CUDA >= 11.8
 (driver >= 520), and Volta or later GPU architectures. There are also CPU-only
 packages available, which will be automatically selected when installing on a
-machine without GPUs available. [BUILD.md](BUILD.md) for instructions on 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,17 +232,11 @@ conda install -c nvidia -c conda-forge -c legate legate-core
...
-
-```
-conda ... legate-core=*=*_cpu
-```
-
-See [BUILD.md](BUILD.md) for instructions on building Legate Core from source.
+The default package contains GPU support, and is compatible with CUDA >= 11.8
+(driver >= 520), and Volta or later GPU architectures. There are also CPU-only
+packages available, which will be automatically selected when installing on a
+machine without GPUs available. [BUILD.md](BUILD.md) for instructions on
```
manopapad, line 238:
```suggestion
machine without GPUs available. See [BUILD.md](BUILD.md) for instructions on 
```

manopapad (approved):

PR #942, opened 2024-04-25
Simple addition enabling the passing of scalar arguments to Reduce

JosephGuman:
Ad hoc change for cunumeric np.unique(return_index=True) functionality. I made this separate from the more comprehensive  equivalent function in Task, but it might be worthwhile to move these functions up into some higher level shared ancestor. @magnatelee @rohany 

Signed-off-by: Joseph Guman <joeytg@stanford.edu>

### DIFF
--- a/legate/core/context.py
+++ b/legate/core/context.py
@@ -18,6 +18,8 @@ from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union
 
 import numpy as np
 
+import legate.core.types as ty
+
 from . import Future, legion
 from ._legion.util import Logger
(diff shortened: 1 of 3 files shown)


rohany:
I'll take a pass through this, but @manopapad can you review or should we wait for wonchan to come back from vacation?

legate/core/context.py
```diff
@@ -18,6 +18,8 @@
 
 import numpy as np
 
+import legate.core.types as ty
```
rohany, line 21: I think if this is used only in type annotations it needs to go below in the guarded `if TYPE_CHECKING` imports.

PR #974, opened 2025-03-26
Fix typing of legate.timing.time

elliottslaughter:
`legate.timing.time` function takes an *optional* argument and the current type definition makes it appear to be mandatory. This adds the optional argument to the type hints so that it can get type checked in user code properly.


src/python/legate/timing/_lib/timing.pyi
```diff
@@ -20,4 +20,4 @@ class PyTime(Unconstructable):
     def __div__(self, rhs: int | PyTime) -> float: ...
     def __rdiv__(self, lhs: int | PyTime) -> float: ...
 
-def time(units: str) -> PyTime: ...
+def time(units: str = ...) -> PyTime: ...
```
ipdemes, line 23: Should we do something like this instead:?
```suggestion
def time(units: str | None = None) -> PyTime: ...
```

elliottslaughter:
The CI issues appear to be in C++, so I cannot see how they are related to my change?

ipdemes (approved):

ipdemes:
@elliottslaughter : you need to use pre-commit on modified file to get CI to pass. For this please do following in the legate repo:
```
pre-commit install
pre-commit run --files /path/to/modified/file
```

elliottslaughter:
@ipdemes I ran all the pre-commit hooks locally and they pass for me. It is possible something else is going wrong?

```
yamllint.............................................(no files to check)Skipped
check json...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
debug statements (python)............................(no files to check)Skipped
fix end of files.........................................................Passed
pretty format json...................................(no files to check)Skipped
trim trailing whitespace.................................................Passed
angle-includes.......................................(no files to check)Skipped
copyright................................................................Passed
legate-assert........................................(no files to check)Skipped
c-begone.............................................(no files to check)Skipped
traced-throw.........................................(no files to check)Skipped
codespell................................................................Passed
clang-format.........................................(no files to check)Skipped
ruff-format..............................................................Passed
ruff.....................................................................Passed
mypy.....................................................................Passed
cython-lint..........................................(no files to check)Skipped
double-quote Cython strings..........................(no files to check)Skipped
Check minimum Python version.........................(no files to check)Skipped
Test shell scripts with shellcheck...................(no files to check)Skipped
Lint GitHub Actions workflow files...................(no files to check)Skipped
legate-defined.......................................(no files to check)Skipped
no-default...........................................(no files to check)Skipped
cmake-format.........................................(no files to check)Skipped
- hook id: cmake-format
cmake-lint...........................................(no files to check)Skipped
- hook id: cmake-lint
dej2lint.............................................(no files to check)Skipped
check_cmake_format...................................(no files to check)Skipped
```

manopapad:
The CI is probably not properly set up on the public repos yet. Pre-commit in particular seems to be missing shellcheck @m3vaz. Clang-tidy seems to be flagging issues that are not flagged in the internal repo, wdyt is causing the divergence @Jacobfaib?

Jacobfaib:
> Clang-tidy seems to be flagging issues that are not flagged in the internal repo, wdyt is causing the divergence @Jacobfaib?

The "issue" is that `clang-tidy` was updated to a new version with more strict checks. These were fixed on the internal side weeks ago, have we not been syncing the public repo against it?

manopapad:
I duplicated this to the internal mirror to run CI, and merged it there.

Issue #975, opened 2025-03-26
Legate nested imports not considered exported by Python type checkers


elliottslaughter:
When you write a Legate program like:

```python
from legate.timing import time
_ = time()
```

A Python type checker like [basedpyright](https://docs.basedpyright.com/latest/) will report a warning that looks like:

```
warning: "time" is not exported from module "legate.timing"                                                                                                      
    Import from "legate.timing._lib.timing" instead (reportPrivateImportUsage)
```

This occurs because the Legate `timing` module does not re-export `time`:

https://github.com/nv-legate/legate/blob/1478fa09e7e3cb6b177bef0aae74cd04c3a43417/src/python/legate/timing/__init__.py#L6

My understanding based on the [pyright documentation](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface) is that there are two ways to fix this:

Option 1: re-export by name:

```python
from ._lib.timing import time as time
```

Option 2: explicit export list:

```python
from ._lib.timing import time
__all__ = ["time"]
```

Note that this is a pervasive issue, and it applies to more than just the `timing` module. Almost every Legate module suffers from the same issue, and some modules are rather larger than others. So I think it's worth planning how to address this, because at the moment it prevents users from cleanly type checking their Legate client code.

manopapad:
@bryevdv wdyt?

bryevdv:
I'd suggest constructing the `__all__` value. Then that can be put under test as a roadblock for unintentional changes to the intended/expected API.

manopapad:
Makes sense, let's go with `__all__`. @bryevdv please put this in your roadmap.

manopapad:
A fix was attempted but it didn't actually resolve @elliottslaughter 's issues.

elliottslaughter:
Maybe I was testing too old of a version. This appears to be fixed now.

PR #982, opened 2025-04-12
Export time from legate.timing

elliottslaughter:
This is a band-aid around #975 until we get a real fix. For now this addresses the main sore point in my code that I cannot use `legate.timing.time` without type checkers complaining that the function has not been exported.

### DIFF
--- a/src/python/legate/timing/__init__.py
+++ b/src/python/legate/timing/__init__.py
@@ -5,4 +5,4 @@ from __future__ import annotations
 
 from ._lib.timing import time
 
-__all__ = ['time']
+__all__ = ["time"]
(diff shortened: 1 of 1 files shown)


elliottslaughter:
I am not trying to duplicate that effort, as long as it will land in a reasonable time frame.

However, to be honest, you might want to reconsider your approach if you want this to be an open source project with an active community. Right now the process is fairly opaque, doesn't work smoothly (e.g., every PR I've sent so far has failed for unrelated reasons), and I cannot see if I am doing duplicate work or not.

bryevdv:
> However, to be honest, you might want to reconsider your approach if you want this to be an open source project with an active community. Right now the process is fairly opaque, doesn't work smoothly (e.g., every PR I've sent so far has failed for unrelated reasons), and I cannot see if I am doing duplicate work or not.

cc @manopapad

bryevdv:
The PR for this has been merged in the other repo

elliottslaughter:
Feel free to close this out when you're ready.

